PR #1, opened 2020-04-09
Fix markdown

e-kwsm:


jasukhar:
Hello @e-kwsm ! Thank you for the pull request! Probably the best option will be to replace html table code completely with markdown code. if you are willing to make this change please update your pull request, otherwise we will do this on our side.

e-kwsm:
> Probably the best option will be to replace html table code completely with markdown code.

Markdown does not support merging rows/columns (`rowspan`/`colspan`), so I leave the tables that uses the attributes.

jasukhar:
I agree that changes are not straightforward and requires a lot of changes. After closer look, we want to refactor readme later to remove html and  tables when they are not really needed. Thank you for pointing us to the issue.

rscohn2:
If you call it README.rst, you can use restructured text instead of markdown, which has better support for [tables](https://docutils.sourceforge.io/docs/user/rst/quickref.html#tables). MKL specification is written in restructured text.

PR #4, opened 2020-04-24
[build] Fix #3 remove CMAKE_BUILD_TYPE overriding, set to "Release" if None

### LINKED ISSUE #3: Overriding CMAKE_BUILD_TYPE
The following code:

https://github.com/oneapi-src/oneMKL/blob/7f40c123571995d2e89a910e7844145ec91cd937/CMakeLists.txt#L26-L32

overrides the default `CMAKE_BUILD_TYPE` passed by the command line, and expect users to use `BUILD_DEBUG` flag. Is not really the way CMake users expect to build the programs, some may say is not idiomatic. In practice, it causes some conflicts with Jenkins scripts and user expectations about building the project. 

Is there any particular reason for this? otherwise, `CMAKE_BUILD_TYPE` should be enough to cover for the use case...

mkrainiuk:


CMakeLists.txt
```diff
@@ -21,14 +21,15 @@ cmake_minimum_required (VERSION 3.13)
...
-## Build type
-option(BUILD_DEBUG "" OFF)
+# Define build type
+set(DEFAULT_BUILD_TYPE "Release")
 
-if(BUILD_DEBUG)
-    set(CMAKE_BUILD_TYPE "Debug")
+if("${CMAKE_BUILD_TYPE}" STREQUAL "")
+message(STATUS "CMAKE_BUILD_TYPE: None, set to ${DEFAULT_BUILD_TYPE} by default")
+set(CMAKE_BUILD_TYPE ${DEFAULT_BUILD_TYPE} CACHE STRING
```
vmalia, line 29: Is there a reason for not indenting statements inside if-else?

mkrainiuk: fixed

jasukhar:
intelci:run

PR #5, opened 2020-04-28
Adding BATCH and Extensions BLAS to oneMKL

mmeterel:
This PR is adding BATCH and Extensions APIs and functional tests for BLAS to oneMKL.


mmeterel:
> Looks good to me.
> We will prepare separate PR for splitting template declarations from instantiations for Compile-time API, as @lhuot suggested last time.

@mkrainiuk I was actually adding the changes you have mentioned into my USM related PR. Do you want to wait for that one (WW19) if it is not urgent?

mkrainiuk:
> > Looks good to me.
> > We will prepare separate PR for splitting template declarations from instantiations for Compile-time API, as @lhuot suggested last time.
> 
> @mkrainiuk I was actually adding the changes you have mentioned into my USM related PR. Do you want to wait for that one (WW19) if it is not urgent?

No, it's not, if you already have it. But just in case I added script to generate a file for these templates based on blas.hpp in scripts/generate_ct_templates.py

PR #24, opened 2020-06-08
Enable USM for cuBLAS Level 1/2/3

mmeterel:
# Description

This PR is enabling Unified Shared Memory (USM) for cuBLAS backend for BLAS Level 1/2/3 APIs.

### DIFF
--- a/include/onemkl/blas/detail/cublas/blas_ct.hpp
+++ b/include/onemkl/blas/detail/cublas/blas_ct.hpp
@@ -1970,6 +1970,7 @@ cl::sycl::event syr2<library::cublas, backend::nvidiagpu>(
     auto done =
         onemkl::cublas::syr2(queue, upper_lower, n, alpha, x, incx, y, incy, a, lda, dependencies);
     syr2_postcondition(queue, upper_lower, n, alpha, x, incx, y, incy, a, lda, dependencies);
+    done.wait();
     return done;
 }
 
@@ -1982,6 +1983,7 @@ cl::sycl::event syr2<library::cublas, backend::nvidiagpu>(
     auto done =
         onemkl::cublas::syr2(queue, upper_lower, n, alpha, x, incx, y, incy, a, lda, dependencies);
     syr2_postcondition(queue, upper_lower, n, alpha, x, incx, y, incy, a, lda, dependencies);
+    done.wait();
     return done;
 }
 
@@ -1992,6 +1994,7 @@ cl::sycl::event scal<library::cublas, backend::nvidiagpu>(
     scal_precondition(queue, n, alpha, x, incx, dependencies);
     auto done = onemkl::cublas::scal(queue, n, alpha, x, incx, dependencies);
     scal_postcondition(queue, n, alpha, x, incx, dependencies);
+    done.wait();
     return done;
 }
 
@@ -2002,6 +2005,7 @@ cl::sycl::event scal<library::cublas, backend::nvidiagpu>(
     scal_precondition(queue, n, alpha, x, incx, dependencies);
     auto done = onemkl::cublas::scal(queue, n, alpha, x, incx, dependencies);
     scal_postcondition(queue, n, alpha, x, incx, dependencies);
+    done.wait();
     return done;
 }
 
@@ -2012,6 +2016,7 @@ cl::sycl::event scal<library::cublas, backend::nvidiagpu>(
     scal_precondition(queue, n, alpha, x, incx, dependencies);
     auto done = onemkl::cublas::scal(queue, n, alpha, x, incx, dependencies);
     scal_postcondition(queue, n, alpha, x, incx, dependencies);
+    done.wait();
     return done;
 }
 
@@ -2022,6 +2027,7 @@ cl::sycl::event scal<library::cublas, backend::nvidiagpu>(
     scal_precondition(queue, n, alpha, x, incx, dependencies);
     auto done = onemkl::cublas::scal(queue, n, alpha, x, incx, dependencies);
     scal_postcondition(queue, n, alpha, x, incx, dependencies);
+    done.wait();
     return done;
 }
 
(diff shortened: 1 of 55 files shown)


include/onemkl/blas/detail/cublas/blas_ct.hpp
```diff
@@ -4018,6 +4183,7 @@ cl::sycl::event symv<library::cublas, backend::nvidiagpu>(
     auto done = onemkl::cublas::symv(queue, upper_lower, n, alpha, a, lda, x, incx, beta, y, incy,
                                      dependencies);
     symv_postcondition(queue, upper_lower, n, alpha, a, lda, x, incx, beta, y, incy, dependencies);
+    done.wait();
```
lhuot, line 4186: Could you please explain why we need to wait here?
And if synchronization is required why do we return the event?

src/blas/backends/cublas/cublas_level3.cpp
```diff
@@ -387,7 +387,26 @@ inline cl::sycl::event gemm(Func func, cl::sycl::queue &queue, transpose transa,
...
+            auto a_     = reinterpret_cast<const cuDataType *>(a);
+            auto b_     = reinterpret_cast<const cuDataType *>(b);
+            auto c_     = reinterpret_cast<cuDataType *>(c);
+            cublasStatus_t err;
+            CUBLAS_ERROR_FUNC(func, err, handle, get_cublas_operation(transa),
+                              get_cublas_operation(transb), m, n, k, (cuDataType *)&alpha, a_, lda,
+                              b_, ldb, (cuDataType *)&beta, c_, ldc);
+        });
+    });
+    return done;
```
lhuot, line 409: Would it be easier to read if we separate buffer and USM APIs in different files?

tests/unit_tests/blas/level1/asum_usm.cpp
```diff
@@ -86,7 +86,9 @@ int test(const device& dev, int64_t N, int64_t incx) {
 #else
         TEST_RUN_CT(main_queue, onemkl::blas::asum,
                     (main_queue, N, x.data(), incx, result_p, dependencies));
+    #ifndef ENABLE_CUBLAS_BACKEND
         main_queue.wait();
```
lhuot, line 90: Why do we wait on the queue for the USM tests and not on the returned event?

src/blas/backends/cublas/cublas_level3.cpp
mmeterel: Thanks for feedback. In most of the backends and header files (except mklgpu), Buffer and USM APIs are implemented in same file. Personally, I think, this is more concise and consistent with rest of code.

include/onemkl/blas/detail/cublas/blas_ct.hpp
mmeterel: This is due to this issue (https://github.com/intel/llvm/issues/1569). 
However, I just found out that, the issue is fixed in recent builds, so I will test and update the code. (remove the wait().

tests/unit_tests/blas/level1/asum_usm.cpp
mmeterel: This is due to this issue (intel/llvm#1569).
However, I just found out that, the issue is fixed in recent builds, so I will test and update the code. (remove #ifndef ENABLE_CUBLAS_BACKEND).

include/onemkl/blas/detail/cublas/blas_ct.hpp
mmeterel: Removed the done.wait() command.

tests/unit_tests/blas/level1/asum_usm.cpp
mmeterel: Line removed.

src/blas/backends/cublas/cublas_level3.cpp
mmeterel: I guess, I am more concerned about the consistency across the repo. If we agree on splitting the files, I would prefer doing it for all .cpp/.hpp files for all backends.

lhuot: yes the suggestion was a "generic" one to apply everywhere applicable

tests/unit_tests/blas/level1/dot.cpp
```diff
@@ -103,7 +103,17 @@ int test(const device& dev, int N, int incx, int incy) {
...
-        good                 = check_equal(result_accessor[0], result_ref, N, std::cout);
+#ifndef ENABLE_CUBLAS_BACKEND
+        good = check_equal(result_accessor[0], result_ref, N, std::cout);
+#else
+        if (std::is_same<fp, fp_res>::value) {
+            good = check_equal(result_accessor[0], result_ref, N, std::cout);
+        }
+        else {
+            // Increase error margin for cuBLAS since computation is done in float instead of double.
+            good = check_equal(result_accessor[0], result_ref, N * 10e5, std::cout);
```
mkrainiuk, line 114: Is it expected error for users in this case? Or should we add this information as known limitation?

tests/unit_tests/blas/level1/rotg_usm.cpp
```diff
@@ -73,6 +73,8 @@ int test(const device &dev) {
     c     = rand_scalar<fp_scalar>();
     a_ref = a;
     b_ref = b;
+    s_ref = s;
+    c_ref = c;
```
mkrainiuk, line 77: Do you use these new variables?

tests/unit_tests/blas/level1/dot.cpp
mmeterel: Good point. This is a known limitation since cuBLAS does not support this API. Before this PR, this test was not enabled, hence we did not see this limitation. Other option would be to throw not_supported exception. What do you think?

tests/unit_tests/blas/level1/rotg_usm.cpp
mmeterel: Yes, they are used. They were not initialized before. I am fixing that here.

tests/unit_tests/blas/level1/dot.cpp
mkrainiuk: I think if cuBLAS doesn't support this functionality we can throw not_supported exception, so that it will be transparent for users. I guess they also can work it around on their side, right?

mmeterel: Yes, the user can call dot with double precision (slower but accurate)

PR #26, opened 2020-06-22
Enable Batch and Blas-like APIs for cuBLAS

mmeterel:
This PR is enabling Batch and Blas-like APIs for cuBLAS backend for the supported functions.

### DIFF
--- a/src/blas/backends/cublas/cublas_batch.cpp
+++ b/src/blas/backends/cublas/cublas_batch.cpp
@@ -170,18 +170,18 @@ inline cl::sycl::event gemm_batch(Func func, cl::sycl::queue &queue, transpose *
         cgh.interop_task([=](cl::sycl::interop_handler ih) {
             auto sc        = CublasScopedContextHandler(queue);
             auto handle    = sc.get_handle(queue);
-            int64_t stride = 0;
+            int64_t offset = 0;
             cublasStatus_t err;
             for (int64_t i = 0; i < group_count; i++) {
                 auto **a_ = reinterpret_cast<const cuDataType **>(a);
                 auto **b_ = reinterpret_cast<const cuDataType **>(b);
                 auto **c_ = reinterpret_cast<cuDataType **>(c);
-                stride    = (i == 0) ? 0 : stride + group_size[i - 1];
                 CUBLAS_ERROR_FUNC(func, err, handle, get_cublas_operation(transa[i]),
                                   get_cublas_operation(transb[i]), (int)m[i], (int)n[i], (int)k[i],
-                                  (cuDataType *)&alpha[i], a_ + stride, (int)lda[i], b_ + stride,
-                                  (int)ldb[i], (cuDataType *)&beta[i], c_ + stride, (int)ldc[i],
+                                  (cuDataType *)&alpha[i], a_ + offset, (int)lda[i], b_ + offset,
+                                  (int)ldb[i], (cuDataType *)&beta[i], c_ + offset, (int)ldc[i],
                                   (int)group_size[i]);
+                offset += group_size[i];
             }
         });
     });
--- a/src/blas/backends/cublas/cublas_batch.cpp
+++ b/src/blas/backends/cublas/cublas_batch.cpp
@@ -16,7 +16,9 @@
 *  limitations under the License.
 *
 **************************************************************************/
+#include <CL/sycl/detail/pi.hpp>
 #include "cublas_helper.hpp"
+#include "cublas_scope_handle.hpp"
 #include "include/exceptions_helper.hpp"
 #include "onemkl/blas/detail/cublas/onemkl_blas_cublas.hpp"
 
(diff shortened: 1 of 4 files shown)


src/blas/backends/cublas/cublas_extensions.cpp
```diff
@@ -73,44 +68,50 @@ void gemm_ext(cl::sycl::queue &queue, transpose transa, transpose transb, offset
...
-    throw backend_unsupported_exception();
-}
-
-void gemm_ext(cl::sycl::queue &queue, transpose transa, transpose transb, std::int64_t m,
-              std::int64_t n, std::int64_t k, double alpha, cl::sycl::buffer<double, 1> &a,
-              std::int64_t lda, cl::sycl::buffer<double, 1> &b, std::int64_t ldb, double beta,
-              cl::sycl::buffer<double, 1> &c, std::int64_t ldc) {
-    throw backend_unsupported_exception();
+template <typename Func, typename T_AB, typename T_C, typename DATATYPE_AB, typename DATATYPE_C>
+inline void gemm_ext(Func func, DATATYPE_AB DT_AB, DATATYPE_C DT_C, cl::sycl::queue &queue,
```
lhuot, line 72: To anticipate future functionality, could you use different template type for A and B matrices?

tests/unit_tests/blas/batch/gemm_batch_usm.cpp
```diff
@@ -139,39 +134,53 @@ int test(const device &dev, int64_t group_count) {
...
+    c_array_data.resize(total_size_c);
+    c_ref_array_data.resize(total_size_c);
+
     idx = 0;
     for (i = 0; i < group_count; i++) {
         for (j = 0; j < group_size[i]; j++) {
-            a_array[idx] = (fp *)onemkl::malloc_shared(64, sizeof(fp) * total_size_a[i], dev, cxt);
-            b_array[idx] = (fp *)onemkl::malloc_shared(64, sizeof(fp) * total_size_b[i], dev, cxt);
-            c_array[idx] = (fp *)onemkl::malloc_shared(64, sizeof(fp) * total_size_c[i], dev, cxt);
-            c_ref_array[idx] =
```
lhuot, line 167: Is there a reason why changing from independent pointers as allowed by the API to contiguous matrices?

tests/unit_tests/blas/batch/gemm_batch_usm.cpp
```diff
@@ -139,39 +134,53 @@ int test(const device &dev, int64_t group_count) {
...
-            copy_matrix(c_array[idx], onemkl::transpose::nontrans, m[i], n[i], ldc[i],
-                        c_ref_array[idx]);
+            vector<fp, decltype(uafp)> a_array_temp(uafp), b_array_temp(uafp), c_array_temp(uafp),
+                c_ref_array_temp(uafp);
+            rand_matrix(a_array_temp, transa[i], m[i], k[i], lda[i]);
+            rand_matrix(b_array_temp, transb[i], k[i], n[i], ldb[i]);
+            rand_matrix(c_array_temp, onemkl::transpose::nontrans, m[i], n[i], ldc[i]);
+            copy_matrix(c_array_temp, onemkl::transpose::nontrans, m[i], n[i], ldc[i],
+                        c_ref_array_temp);
+            std::copy(a_array_temp.begin(), a_array_temp.end(), a_array_data.begin() + stride_a);
```
lhuot, line 172: Those stride_? variable name are misleading, I would rather name them offset_?

tests/unit_tests/blas/batch/gemm_batch_usm.cpp
mmeterel: cuBLAS group API is expecting contiguous matrices. This change is done to support all backends.

src/blas/backends/cublas/cublas_extensions.cpp
mmeterel: Makes sense. Will update.

tests/unit_tests/blas/batch/gemm_batch_usm.cpp
mmeterel: Sounds good.

src/blas/backends/cublas/cublas_extensions.cpp
mmeterel: Template parameters are updated.

tests/unit_tests/blas/batch/gemm_batch_usm.cpp
mmeterel: Variable names are updated.

tests/unit_tests/blas/batch/gemm_batch_usm.cpp
lhuot: Do you have a pointer to the cuBLAS documentation that mentions this requirement? I could not find it.

mmeterel: I take it back, you are right. I got confused. cuBLAS is also accepting array of pointers. 
Regarding contiguous matrices: Using malloc_shared for pointer arrays caused "CUDA_ERROR_ILLEGAL_ADDRESS" error in CUDA PI. As a workaround, I switched allocating array pointers and matrices using usm_allocator<T, usm::alloc::shared>. With that, I couldn't find a way to allocate matrices in a loop and still keep them alive beyond scope. Instead, I allocated a large array and copied the temporary matrices into this large array. This large array ended up being contiguous. In backends though, each array is accessed through pointers, so, they don't really care (or use) if the large array is contiguous or not.

src/blas/backends/cublas/cublas_batch.cpp
```diff
@@ -96,78 +105,105 @@ void trsm_batch(cl::sycl::queue &queue, side left_right, uplo upper_lower, trans
...
+        cgh.interop_task([=](cl::sycl::interop_handler ih) {
+            auto sc        = CublasScopedContextHandler(queue);
+            auto handle    = sc.get_handle(queue);
+            int64_t stride = 0;
+            cublasStatus_t err;
+            for (int64_t i = 0; i < group_count; i++) {
+                auto **a_ = reinterpret_cast<const cuDataType **>(a);
+                auto **b_ = reinterpret_cast<const cuDataType **>(b);
+                auto **c_ = reinterpret_cast<cuDataType **>(c);
+                stride    = (i == 0) ? 0 : stride + group_size[i - 1];
```
lhuot, line 179: offset for the variable name would be more appropriate maybe.
Also you can move it after the cublas call and remove the if statement using offset += group_size[i]

mmeterel: Sounds good.

mmeterel: Updated

tests/unit_tests/blas/batch/gemm_batch_usm.cpp
mmeterel: Updated memory allocation for matrices using vector of vectors.

mmeterel: @lhuot Does this look good?

mmeterel: After some study, I realized that, vector of vectors will also provide contiguous data array. So, I did more debug on why malloc_shared does not work in this case. The root-cause is using non-aligned memory. After fixing it, malloc_shared works fine and the non-contiguous data arrays are provided to the library calls.

tests/unit_tests/blas/batch/gemm_batch_usm.cpp
```diff
@@ -116,40 +122,32 @@ int test(const device &dev, int64_t group_count) {
...
-    for (i = 0; i < group_count; i++) {
-        for (j = 0; j < group_size[i]; j++) {
-            a_array.push_back(&a_array_data[idx][0]);
-            b_array.push_back(&b_array_data[idx][0]);
-            c_array.push_back(&c_array_data[idx][0]);
-            c_ref_array.push_back(&c_ref_array_data[idx][0]);
+            a_array[idx]     = (fp *)onemkl::malloc_shared(64, sizeof(fp) * size_a[i], dev, cxt);
+            b_array[idx]     = (fp *)onemkl::malloc_shared(64, sizeof(fp) * size_b[i], dev, cxt);
+            c_array[idx]     = (fp *)onemkl::malloc_shared(64, sizeof(fp) * size_c[i], dev, cxt);
+            c_ref_array[idx] = (fp *)onemkl::malloc_shared(64, sizeof(fp) * size_c[i], dev, cxt);
```
lhuot, line 145: size_{a,b,c} are used only here, I would suggest to not use array to store those sizes and move the size_{a,b,c} from the first loop of the test to the outer loop here.

tests/unit_tests/blas/batch/gemm_batch_usm.cpp
lhuot: Thanks!

tests/unit_tests/blas/batch/gemm_batch_usm.cpp
mmeterel: Refactored size_? parameters

PR #29, opened 2020-07-16
Make README and conan recipe independent of oneAPI version

vmalia:
# Description

Currently, the README and conanfile.py must be updated for every new release of oneAPI. This PR proposes to remove this dependency and always use the latest oneAPI version.

PR #32, opened 2020-07-22
[Conan]Update Conan, CMake and Ninja;Fix Conan builds

vmalia:
# Description

*Update Ninja to 1.10.0 to support FORTRAN builds (Fixes Netlib LAPACK builds via Conan)
*Update CMake to 3.15 to support Ninja 1.10.0
*Update Conan to 1.24 to support OpenSSL package(new dependency for CMake package)
*Disable Conan compiler check to avoid failure when using DPC++ compiler

### DIFF
--- a/conan/profiles/inteldpcpp_lnx
+++ b/conan/profiles/inteldpcpp_lnx
@@ -30,8 +30,8 @@ compiler.libcxx=libstdc++11
 build_type=Release
 
 [build_requires]
-ninja/1.9.0
-cmake/[>=3.13]
+cmake/[>=3.15]
+ninja/1.10.0
 
 [env]
 CONAN_CMAKE_GENERATOR="Ninja"
--- a/conanfile.py
+++ b/conanfile.py
@@ -135,7 +135,7 @@ THIRD-PARTY-PROGRAMS file and in the README.md file included with the Software P
             # Paramaters
             # Conan does not officially support the DPC++ compiler, hence disable the compiler_id check
             "CONAN_DISABLE_CHECK_COMPILER" : True,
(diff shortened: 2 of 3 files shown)


conan/profiles/inteldpcpp_lnx
```diff
@@ -30,8 +30,8 @@ compiler.libcxx=libstdc++11
 build_type=Release
 
 [build_requires]
-ninja/1.9.0
-cmake/[>=3.13]
+cmake/[>=3.15]
+ninja/1.10.0
```
mkrainiuk, line 34: We need to update supported versions in README as well

conanfile.py
```diff
@@ -133,7 +133,9 @@ def build(self):
             "BUILD_DOC"                : self.options.build_doc,
 
             # Paramaters
-            "MKL_ROOT"                 : f"/opt/intel/inteloneapi/mkl/{self.oneapi_version}",
+            # Conan does not officially support the DPC++ compiler, hence disable the compiler_id check
+            "CONAN_DISABLE_CHECK_COMPILER" : True,
+            "MKL_ROOT"                 : f"/opt/intel/inteloneapi/mkl/latest",
```
mkrainiuk, line 138: I see the same changes for MKL_ROOT in PR #29. Do you plan to close it and extend this one with other changes or keep them separated?

vmalia: I will keep the PRs separate. Reverted that change in this PR.

conan/profiles/inteldpcpp_lnx
vmalia: Done

PR #37, opened 2020-08-04
Fix namespaces

vmalia:
# Description

Current rules in Clang-format result in too many changes when modifying the architecture/design of the project. Some rules result in incorrect code and interpretation by the compiler. Some rules are disabled/modified.

Some namespace changes were missed in the previous MR. Adding them here.


jasukhar:
Can we separate these 2 different commits? The standard practice is to have a commit per 1 problem - otherwise it's hard to maintain and debug in the future

PR #41, opened 2020-08-11
Fix parser for helper scripts to support USM API as input

mkrainiuk:
# Description

Generators for Compile-time API can't work with USM API, because func_parser, that they all use, doesn't understand ``dependencies`` parameter with default value ``dependencies = {}``
The fix will make func_parser support this parameter correctly.

This is a helper scrip so it won't affect build and tests.

### DIFF
--- a/scripts/func_parser.py
+++ b/scripts/func_parser.py
@@ -61,7 +61,7 @@ def parse_item(item):
                      for idx, x in enumerate(par_list)]
 
             """Extract names to call_list"""
-            call_list = [x.rsplit(' ', 1)[1].strip(' *').strip('\[\]').strip('&') \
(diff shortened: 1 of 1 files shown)


mmeterel:
Hi @vmalia Did you try other scripts as well? I wonder if they already work and only this one is broken.

mkrainiuk:
@mmeterel , do you see that some script still doesn't work correctly after this fix? 
Previously all scripts didn't work, because they all use this func_parser to get the list of functions, and this parser skips USM API because of the issue that this fix is for.

mmeterel:
@mkrainiuk I did not test the scripts myself. I will probably use them when I add future backends.

PR #55, opened 2020-11-10
Extend RNG domain APIs and enable oneMKL beta10

aelizaro:
# Description

- RNG: Add mrg32k3a engine and default engine
- RNG: Add lognormal, bernoulli, poisson distributions
- RNG: change entry points for mklgpu backend
- Add /MD option for windows in cmake

# Checklist

## All Submissions

[lnx_rng_test_log.txt](https://github.com/oneapi-src/oneMKL/files/5517627/lnx_rng_test_log.txt)

### DIFF
--- a/deps/googletest/cmake/internal_utils.cmake
+++ b/deps/googletest/cmake/internal_utils.cmake
@@ -114,7 +114,7 @@ macro(config_compiler_and_linker)
     # explicitly.
     set(cxx_no_rtti_flags "-qnortti -DGTEST_HAS_RTTI=0")
   elseif (CMAKE_CXX_COMPILER_ID STREQUAL "HP")
-    set(cxx_base_flags "-AA -md")
+    set(cxx_base_flags "-AA -mt")
     set(cxx_exception_flags "-DGTEST_HAS_EXCEPTIONS=1")
     set(cxx_no_exception_flags "+noeh -DGTEST_HAS_EXCEPTIONS=0")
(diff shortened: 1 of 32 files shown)


deps/googletest/cmake/internal_utils.cmake
```diff
@@ -133,7 +133,11 @@ macro(config_compiler_and_linker)
   set(cxx_exception "${cxx_base_flags} ${cxx_exception_flags}")
   set(cxx_no_exception
     "${CMAKE_CXX_FLAGS} ${cxx_base_flags} ${cxx_no_exception_flags}")
+  if(NOT WIN32)
   set(cxx_default "${cxx_exception}")
+  else()
+  set(cxx_default "${cxx_exception} /MD")
```
mkrainiuk, line 139: As far as I know DPC++ compiler defines /MD by default, do we really need to provide it explicitly here?

aelizaro: I haven't found any mentions of /MD or /MT in ninja.build of test part, but without that explicit /MD I see the error that gtest-all is built as a static one

mkrainiuk: Oh, I see, so we use the same set of options for oneMKL build and for gtests, that use not DPC++ compiler. We will need to improve it in the future, but for now I agree, this is the only one solution.

mkrainiuk (approved):

mmeterel (approved):

PR #63, opened 2020-12-09
Fix to solve ONEMKL::SYCL::SYCL target not found bug

AlexReinemann:
This contains a fix for an issue that caused the SYCL environment variables to not be initialized correctly

All tests have passed locally.
Tests have not passed in test environment yet

### DIFF
--- a/cmake/FindCompiler.cmake
+++ b/cmake/FindCompiler.cmake
@@ -25,10 +25,6 @@ include(FindPackageHandleStandardArgs)
 check_cxx_compiler_flag("-fsycl" is_dpcpp)
 
 if(is_dpcpp)
-  find_package_handle_standard_args(SYCL DEFAULT_MSG)
-endif()
-
-if(SYCL_FOUND AND is_dpcpp)
   # Workaround for internal compiler error during linking if -fsycl is used
   get_filename_component(SYCL_BINARY_DIR ${CMAKE_CXX_COMPILER} DIRECTORY)
   find_library(SYCL_LIBRARY NAMES sycl PATHS "${SYCL_BINARY_DIR}/../lib")
--- a/cmake/FindTBB.cmake
+++ b/cmake/FindTBB.cmake
@@ -40,6 +40,7 @@ find_library(TBB_LIBRARIES NAMES tbb
                  "../tbb/lib/intel64/gcc4.4" "../tbb/lib/intel64/gcc4.8"
                  "../../tbb/latest/lib/intel64/gcc4.8"
                  "../tbb/lib/intel64/vc14"
+                 "../tbb/lib/intel64/vc_mt"
                  )
 
 #Workaround for ref problem
--- a/cmake/FindTBB.cmake
+++ b/cmake/FindTBB.cmake
@@ -31,7 +31,7 @@ find_path(TBB_LIB_DIR ${TBB_LIBNAME}
       PATH_SUFFIXES "lib" "lib/intel64/gcc4.4" "lib/intel64/gcc4.8"
                "../tbb/lib/intel64/gcc4.4" "../tbb/lib/intel64/gcc4.8"
                "../../tbb/latest/lib/intel64/gcc4.8"
-               "../tbb/lib/intel64/vc_mt" "lib/intel64/vc_mt" "../tbb/lib/intel64/vc14"
+               "../tbb/lib/intel64/vc14"
 )
 
 find_library(TBB_LIBRARIES NAMES tbb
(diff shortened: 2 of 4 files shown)


vmalia (approved):

cmake/FindCompiler.cmake
```diff
@@ -25,10 +25,6 @@ include(FindPackageHandleStandardArgs)
 check_cxx_compiler_flag("-fsycl" is_dpcpp)
 
 if(is_dpcpp)
-  find_package_handle_standard_args(SYCL DEFAULT_MSG)
```
mkrainiuk, line 28: Do you know what exactly this part does and why we don't need it anymore?

mkrainiuk (commented):
I see 8 commits instead of one, is it still some problem with GitHub?

vmalia:
> I see 8 commits instead of one, is it still some problem with GitHub?

Yes, there are 8 commits. What is the problem exactly?

mkrainiuk:
> > I see 8 commits instead of one, is it still some problem with GitHub?
> 
> Yes, there are 8 commits. What is the problem exactly?

Most of them are already in the main repo. From the changes I see only two commits are valid to be committed.

cmake/FindCompiler.cmake
vmalia: No, we couldn't exactly find the purpose of this code. However, @AlexReinemann tested this change with DPC++ LLVM compiler and oneAPI DPC++ compiler and it works as expected.

As far as I understand, checking for "-fsycl" should suffice as an indicator that the compiler is supported. Do you think we need anything beyond this check?

cmake/FindTBB.cmake
```diff
@@ -31,15 +31,15 @@ find_path(TBB_LIB_DIR ${TBB_LIBNAME}
       PATH_SUFFIXES "lib" "lib/intel64/gcc4.4" "lib/intel64/gcc4.8"
                "../tbb/lib/intel64/gcc4.4" "../tbb/lib/intel64/gcc4.8"
                "../../tbb/latest/lib/intel64/gcc4.8"
-               "../tbb/lib/intel64/vc_mt" "lib/intel64/vc_mt"
```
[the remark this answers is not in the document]
AlexReinemann: vc_mt has been removed

cmake/FindCompiler.cmake
AlexReinemann: I believe this set the variable SYCL and was not working correctly.  since the if statement below was changed to not use the variable SYCL, then this call is no longer needed either.

AlexReinemann:
> > > I see 8 commits instead of one, is it still some problem with GitHub?
> > 
> > 
> > Yes, there are 8 commits. What is the problem exactly?
> 
> Most of them are already in the main repo. From the changes I see only two commits are valid to be committed.

The diff only shows 2 files changed. this is likely a weird thing with github keeping a count of commits since the PR was created rather than accounting for those commits already being merged in.

cmake/FindTBB.cmake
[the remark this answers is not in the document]
mkrainiuk: Thanks! 
"lib/intel64/vc14" was lost

cmake/FindTBB.cmake
```diff
@@ -31,15 +31,15 @@ find_path(TBB_LIB_DIR ${TBB_LIBNAME}
...
+               "../tbb/lib/intel64/vc14"
 )
 
 find_library(TBB_LIBRARIES NAMES tbb
         HINTS $ENV{TBBROOT} $ENV{MKLROOT} ${MKL_ROOT} ${TBB_ROOT}
         PATH_SUFFIXES "lib" "lib/intel64/gcc4.4" "lib/intel64/gcc4.8"
                  "../tbb/lib/intel64/gcc4.4" "../tbb/lib/intel64/gcc4.8"
                  "../../tbb/latest/lib/intel64/gcc4.8"
-                 "../tbb/lib/intel64/vc_mt"
```
mkrainiuk, line 42: Could you please add "lib/intel64/vc_mt" here as well for TBBROOT search?

cmake/FindTBB.cmake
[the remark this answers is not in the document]
AlexReinemann: no it wasnt, it was pushed below what is seen in this snippet.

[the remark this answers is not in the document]
mkrainiuk: Hmm, I opened sources from your branch and I see only "../tbb/lib/intel64/vc14" in FindTBB, there is no "lib/intel64/vc14"

cmake/FindTBB.cmake
mkrainiuk: Oh, sorry, I mean "lib/intel64/vc14" and remove "../tbb/lib/intel64/vc_mt"

mkrainiuk (approved):

mmeterel:
Do the tests pass locally on windows as well?

Issue #49, opened 2020-10-03
Couldn't load selected backend


vrpascuzzi:
I've compiled intel/llvm ([a7ad8b8](https://github.com/intel/llvm/commit/a7ad8b8caab43ffe3540dc94692859c0ad3130e8)) for CUDA:
```
python3 ${LLVM_SRC}/buildbot/configure.py --shared-libs --cmake-gen "Unix Makefiles" --cuda -o .
```
and using it built oneapi-src/oneMKL (f805087) against oneAPI/mkl/beta08 (beta09 wasn't supported in this revision):
```
export CXX=`which clang++`
source <path_to>/mkl/2021.1-beta08/env/vars.sh
cmake -DBUILD_FUNCTIONAL_TESTS=OFF  ${ONEMKL_SRC}
```
I'm using CUDA 10.2 > 10.0 so should be compatible. To compile the test program (below):
```
$ clang++ -fsycl -std=c++17 -fsycl-targets=nvptx64-nvidia-cuda-sycldevice \
  -Wno-unknown-cuda-version -DSYCL_TARGET_CUDA test_mkl.cc -lonemkl
```
the test compiles cleanly, but when running I get:
```
 $ SYCL_DEVICE_FILTER=cuda:* ./a.out 
Running on: GeForce RTX 2080 SUPER
terminate called after throwing an instance of 'oneapi::mkl::backend_not_found'
  what():  oneMKL: Couldn't load selected backend
Aborted (core dumped)
```
When adding `-lonemkl_rng_mklgpu` to the compile line:
```
$ clang++ -fsycl -std=c++17 -fsycl-targets=nvptx64-nvidia-cuda-sycldevice -Wno-unknown-cuda-version -DSYCL_TARGET_CU
DA test_mkl.cc -lonemkl -lonemkl_rng_mklgpu
//bld4/opt/intel/inteloneapi/mkl/2021.1-beta08/lib/intel64/libmkl_sycl.so: undefined reference to `cl::sycl::level0::make_platform(unsign
ed long)'
//bld4/opt/intel/inteloneapi/mkl/2021.1-beta08/lib/intel64/libmkl_sycl.so: undefined reference to `cl::sycl::context::context(cl::sycl::d
evice const&, std::function<void (cl::sycl::exception_list)>, bool)'
[...]

PR #75, opened 2021-02-13
Add curand support

vrpascuzzi:
# Description

This PR adds a cuRAND backend to oneMKL.

Addresses #49

# Checklist

## All Submissions

[ctest_curand.LOG](https://github.com/oneapi-src/oneMKL/files/5976718/ctest_curand.LOG)

## New features

It is desirable to be able to use oneMKL RNG on CUDA devices.


vrpascuzzi:
Also will squash/fix commits after review but before merge.

src/rng/backends/curand/mrg32k3a.cpp
```diff
@@ -0,0 +1,956 @@
...
+                auto acc = r.get_access<cl::sycl::access::mode::read_write>(cgh);
+                cgh.codeplay_host_task([=](cl::sycl::interop_handle ih) {
+                    auto r_ptr =
+                        reinterpret_cast<float*>(ih.get_native_mem<cl::sycl::backend::cuda>(acc));
+                    curandStatus_t status;
+                    CURAND_CALL(curandGenerateUniform, status, engine_, r_ptr, n);
+                });
+            })
+            .wait_and_throw();
+        range_transform_fp<float>(queue_, distr.a(), distr.b(), n, r);
```
aelizaro, line 125: For oneMKL RNG we have an additional boundary check in the accurate method: does it makes sense to add range_transfor_fp_accurate, which would perform such check?
described in the spec [modes of random number generation](https://spec.oneapi.com/versions/latest/elements/oneMKL/source/domains/rng/distributions.html)

src/rng/backends/curand/mrg32k3a.cpp
```diff
@@ -0,0 +1,956 @@
...
+        const oneapi::mkl::rng::gaussian<double, oneapi::mkl::rng::gaussian_method::icdf>& distr,
+        std::int64_t n, cl::sycl::buffer<double, 1>& r) override {
+        queue_
+            .submit([&](cl::sycl::handler& cgh) {
+                auto acc = r.get_access<cl::sycl::access::mode::read_write>(cgh);
+                cgh.codeplay_host_task([=](cl::sycl::interop_handle ih) {
+                    auto r_ptr =
+                        reinterpret_cast<double*>(ih.get_native_mem<cl::sycl::backend::cuda>(acc));
+                    curandStatus_t status;
+                    CURAND_CALL(curandGenerateNormalDouble, status, engine_, r_ptr, n, distr.mean(),
```
aelizaro, line 206: I am not sure if[ icdf method ](https://spec.oneapi.com/versions/latest/elements/oneMKL/source/domains/rng/distributions-template-parameter-mkl-rng-method-values.html#onemkl-rng-distributions-template-parameter-mkl-rng-method-values)should call the same curandGenerateNormalDouble as it should use different mathematical methods for gaussian transform. (Unfortunately, I haven't found in cuRAND documentation, what method is used). I would suggest adding implementation only in the method, which is used as by_default(box_muller2).

tests/unit_tests/rng/service/skip_ahead.cpp
```diff
@@ -25,44 +25,44 @@ extern std::vector<cl::sycl::device*> devices;
 
 namespace {
 
-class Philox4x32x10SkipAheadTests : public ::testing::TestWithParam<cl::sycl::device*> {};
+// class Philox4x32x10SkipAheadTests : public ::testing::TestWithParam<cl::sycl::device*> {};
```
aelizaro, line 28: Why are these tests commented?
Does curandSetGeneratorOffset() work in a different way than skip_ahead?
SkipAheadEx should be fine, as the test would be skipped in case of unimplemented

aelizaro (commented):
Thank you very much, @vrpascuzzi, for such a huge work!
I have added a few minor RNG-related things to discuss - could you, please, take a look? I would be glad to help with that.

tests/unit_tests/rng/service/skip_ahead.cpp
vrpascuzzi: Thanks for spotting this! These tests were commented earlier for testing. Fixed.

src/rng/backends/curand/mrg32k3a.cpp
vrpascuzzi: Indeed, I noticed this as well; the inverse cumulative distribution function method is only used for quasirandom number generators in cuRAND, e.g.:
https://docs.nvidia.com/cuda/curand/group__HOST.html#group__HOST_1g56ff2b3cf7e28849f73a1e22022bcbfd

I was a bit surprised the tests passed, but since they did, I kept them. I will unimplement functions using ICDF methods.

src/rng/backends/curand/mrg32k3a.cpp
vrpascuzzi: Thanks for the link! However, I don't really understand the difference between "accurate" and "fast"; it is not clear from the description:

"_...For example, random numbers obtained from the generator of continuous distribution that is uniform on interval [a,b] belong to this interval irrespective of what a and b values may be. Fast mode provides high performance generation and also guarantees that generated random numbers belong to the definitional domain except for some specific values of distribution parameters._"

The ad hoc `range_transform_X` functions implemented adjust the generated pseudorandom numbers are within the specified range. Since cuRAND doesn't have this functionality built-in, and the unit tests specify a custom range for the numbers, the transformations are necessary here.

The unclear piece is "_...except for some specific values of distribution parameters_". What criteria are these?

tests/unit_tests/rng/service/skip_ahead.cpp
vrpascuzzi: Attaching the `ctest` log for closure.
[ctest_curand.LOG](https://github.com/oneapi-src/oneMKL/files/5989107/ctest_curand.LOG)

Note that since cuRAND doesn't accept initializer lists for seeds, `skip_ahead` tests do not run.

src/rng/backends/curand/mrg32k3a.cpp
aelizaro: The test is passed, as it checks only statistical properties, which are ok for both methods. But it should be fine also if it throws an unimplemented exception - the test would be just skipped.

src/rng/backends/curand/mrg32k3a.cpp
aelizaro: For an accurate method, there are no criteria for any specific values of distribution parameters, we just have an additional check, like for uniform:
```
if(res < a)
   res = a;
 if(res > b)
    res = b;
```
for any distribution's parameters

aelizaro (approved):
Looks good to me from the RNG side, thank you @vrpascuzzi! @mkrainiuk, @vmalia could you, please, also take a look?

include/oneapi/mkl/detail/backends_table.hpp
```diff
@@ -58,6 +58,7 @@ static std::map<domain, std::map<device, std::vector<const char*>>> libraries =
           } },
         { device::nvidiagpu,
           {
+/// TODO: ENABLE_CUDA_BACKEND?
```
mkrainiuk, line 61: I think it's better to keep this fine granularity since CUDA libraries are separated by domains, unfortunately we can't do it for  oneMKL, since it has all domains in one library.

mkrainiuk (approved):
Changes look good to me. Please reorganize commits a little bit before pushing them.

include/oneapi/mkl/detail/backends_table.hpp
vmalia: Do we need to remove TODO comments before merge?

tests/unit_tests/include/test_helper.hpp
```diff
@@ -67,11 +67,18 @@
 #define TEST_RUN_INTELGPU_SELECT(q, func, ...)
 #endif
 
+/// TODO: Combine into a single CUDA test?
```
vmalia, line 70: Should we remove TODO comments before merge?

include/oneapi/mkl/detail/backends_table.hpp
vrpascuzzi: Agreed. This comment is a relic.

tests/unit_tests/include/test_helper.hpp
vrpascuzzi: For sure. I will fix up and ping when ready.

vmalia (approved):

mkrainiuk:
I squashed commits to one during merge, but all commits that were made are listed in the commit descriptions. Do you think it's better to split them by some groups of changes?

vrpascuzzi:
@mkrainiuk,
Just some temporary commit messages I'd like to clean up; not all are useful. How do you suggest I update this? I can amend the message but not sure how you'd want to propagate to upstream.

Sorry!

mkrainiuk:
If it's only about message you can share how it should look like and we will update comment in upstream branch.

vrpascuzzi:
OK, let's use:
```
This commit adds a cuRAND backend to oneMKL.

The implementation follows closely to that of the existing x86 and 
Intel GPU, with the following limitations:

(1) the cuRAND Host API does not support copy-construction for 
    generators;
(2) cuRAND does not support initializer lists for seeding
    generators;
(3) cuRAND uses inverse cumulative distribution function (ICDF)
    methods only for quasi-random number generators; and 
(4) cuRAND does not support custom ranges on the output sequence
    of random numbers.

Functionality requiring any of 1-3 above are not implemented here
but could be added at a later time. Additional kernels were added to
support (4).
```
If anything more should be added, just let me know. Also probably worth noting is that I've run some benchmarks and for large workloads, SYCL interoperability shows nearly native performance -- great news for SYCL! There is an observable constant overhead (for reasons known) when the GPU is not being heavily utilized.

Thanks, @mkrainiuk!

mkrainiuk:
I've updated commit message. Thank you @vrpascuzzi for enabling new backend!

vrpascuzzi:
Looks great -- thanks!

PR #78, opened 2021-02-18
build: fix Netlib build, propagate export flag correctly on Windows, remove Fortran symbols

mkrainiuk:
# Description

Netlib backend build on Windows has two problems:
1. Static NETLIB backend failed at config step because NETLIB_COPT flag that marked as required is empty.

NETLIB_COPT flag uses to propagate -Donemkl_EXPORT flag for dynamic oneMKL libraries on Windows, but actually this flag should be defined in oneMKL build part, not in Intel MKL or NETLIB third-party libraries config files.
```
-- Found NETLIB: C:/cache/lapack/3.7.1/win/lib/cblas.dll.lib  
-- Found NETLIB: C:/cache/lapack/3.7.1/win/lib/blas.dll.lib  
CMake Error at C:/cache/cmake/3.16.5/win/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find NETLIB (missing: NETLIB_COPT)
```
This PR fixes this problem by introducing oneMKL build common COPT and propagating it to all backends.

```
-- Found NETLIB: C:/cache/lapack/3.7.1/win/lib/cblas.dll.lib  
-- Found NETLIB: C:/cache/lapack/3.7.1/win/lib/blas.dll.lib  
-- Found NETLIB: C:/cache/lapack/3.7.1/win/include  
-- The C compiler identification is Clang 12.0.0 with MSVC-like command-line
```

2. Problems with Fortran Netlib BLAS dependencies resolution in Netlib backend.

Netlib backend was implemented based on CBLAS. We call directly 4 Fortran functions in BLAS level1 that are not implemented in CBLAS. But using Fortran symbols from Netlib BLAS library causes some problems:

- BLAS library built with GNU Fortran compiler causes VS linker resolution problems. VS linker could add Fortran symbols as dependency on CBLAS library, so test will fail to resolve it at runtime.
```
  Error running test executable.

    Path: 'C:/temp/oneMKL/build/bin/test_main_blas_ct.exe'
    Result: Exit code 0xc0000139
```
- BLAS library built with Intel Fortran compiler has different symbol naming (ifort: `UPPERCASE` vs gfortran: `lowercase_`)

This PR adds ref C implementation of 4 Fortran functions missed from CBLAS, it will help to avoid described problems with Fortran symbols. 

# Checklist

## All Submissions

[llvm_netlib_static_win.txt](https://github.com/oneapi-src/oneMKL/files/6037904/llvm_netlib_static_win.txt)
[llvm_netlib_dynamic_win.txt](https://github.com/oneapi-src/oneMKL/files/6037906/llvm_netlib_dynamic_win.txt)
All related tests were passed in CI, but overall CI is reported as failed because of some infrastructure issues.


vmalia (approved):
Looks like each CI has some different failures that can be linked to some kind of infrastructure issue. Collectively, from both CIs, it looks like all tests pass under all configurations.

mmeterel (approved):
Thanks the PR Maria! Looks good.

Issue #67, opened 2021-01-04
oneMKL runtime API tests fail with seg fault on GPU with Level0 driver


mkrainiuk:
# Summary
oneMKL runtime API tests fail with seg fault on GPU when DPC++ compiler uses Level0 as a backend (default behavior)

# Version
The problem appeared with Intel oneMKL and DPC++ compiler update to 2021.1.

# Environment
The issue can be reproduced with:
* HW: Intel GPU (Gen9)
* OS: Ubuntu 18.04
* Intel oneMKL version:  2021.1
* Intel DPC++ Compiler version: 2021.1

# Steps to reproduce
On the machine with Intel GPU and Level0 driver run oneMKL tests:
```
$> mkdir build && cd build
$> cmake .. && cmake --build . -j 24 && ctest
...
        Start   1: BLAS/RT/GemmTestSuite/GemmTests.HalfHalfFloatPrecision/Column_Major_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz
  1/664 Test   #1: BLAS/RT/GemmTestSuite/GemmTests.HalfHalfFloatPrecision/Column_Major_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz ........................   Passed    0.44 sec
        Start   2: BLAS/RT/GemmTestSuite/GemmTests.HalfHalfFloatPrecision/Row_Major_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz
  2/664 Test   #2: BLAS/RT/GemmTestSuite/GemmTests.HalfHalfFloatPrecision/Row_Major_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz ...........................   Passed    0.43 sec
        Start   3: BLAS/RT/GemmTestSuite/GemmTests.HalfHalfFloatPrecision/Column_Major_Intel_R__Graphics_Gen9__0x193b_
  3/664 Test   #3: BLAS/RT/GemmTestSuite/GemmTests.HalfHalfFloatPrecision/Column_Major_Intel_R__Graphics_Gen9__0x193b_ ..................................***Exception: SegFault  2.72 sec
        Start   4: BLAS/RT/GemmTestSuite/GemmTests.HalfHalfFloatPrecision/Row_Major_Intel_R__Graphics_Gen9__0x193b_
[...]

PR #88, opened 2021-04-08
[LAPACK] Enable LAPACK domain

ericlars:
# Integration of LAPACK mklcpu and mklgpu backends.

- Enables runtime and compile-time APIs
- Adds unit tests

# Checklist

## All Submissions

host and cpu [log.txt](https://github.com/oneapi-src/oneMKL/files/6275192/log.txt)
gpu [log_gpu.txt](https://github.com/oneapi-src/oneMKL/files/6386765/log_gpu.txt)
windows [log_lapack_win.txt](https://github.com/oneapi-src/oneMKL/files/6470936/log_lapack_win.txt)

## New interfaces

it was accepted? # (RFC)
- [x] What version of oneAPI spec the interface is targeted? 1.0 Rev3
- [x] Complete [New features](pull_request_template.md#new-features) checklist

## New features


README.md
```diff
@@ -491,8 +511,9 @@ Then:
 # Inside <path to onemkl>
 mkdir build && cd build
 export CXX=<path_to_dpcpp_compiler>/bin/dpcpp;
-cmake .. [-DMKL_ROOT=<mkl_install_prefix>] \               # required only if environment variable MKLROOT is not set
-         [-DREF_BLAS_ROOT=<reference_blas_install_prefix>] # required only for testing
+cmake .. [-DMKL_ROOT=<mkl_install_prefix>] \                    # required only if environment variable MKLROOT is not set
+         [-DREF_BLAS_ROOT=<reference_blas_install_prefix>] \    # required only for testing
+         [-DREF_LAPACK_ROOT=<reference_lapack_install_prefix>]  # required only for testing
```
mkrainiuk, line 516: Do we expect that REF_BLAS_ROOT and REF_LAPACK_ROOT will be different? Do you think we can combine them together somehow?

cmake/FindLAPACKE.cmake
```diff
@@ -0,0 +1,43 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#===============================================================================
+
+include_guard()
+
+find_library(LAPACKE64_file NAMES lapacke64.dll.lib lapacke64.lib lapacke64 HINTS ${REF_LAPACK_ROOT} PATH_SUFFIXES lib lib64)
+find_package_handle_standard_args(LAPACKE64 REQUIRED_VARS LAPACKE64_file)
+find_library(LAPACK64_file NAMES lapack64.dll.lib lapack64.lib lapack64 HINTS ${REF_LAPACK_ROOT} PATH_SUFFIXES lib lib64)
+find_package_handle_standard_args(LAPACK64 REQUIRED_VARS LAPACK64_file)
+find_library(CBLAS64_file NAMES cblas64.dll.lib cblas64.lib cblas64 HINTS ${REF_LAPACK_ROOT} PATH_SUFFIXES lib lib64)
```
mkrainiuk, line 26: So we have FindNetlib, FindCBLAS, and FindLAPACKE. I also wonder if we could combine them somehow as one FindNetlib.

src/lapack/backends/mkl_common/mkl_lapack.cxx
```diff
@@ -0,0 +1,2852 @@
...
+}
+template <>
+ONEMKL_EXPORT std::int64_t ungqr_batch_scratchpad_size<std::complex<float>>(
+    sycl::queue &queue, std::int64_t *m, std::int64_t *n, std::int64_t *k, std::int64_t *lda,
+    std::int64_t group_count, std::int64_t *group_sizes) {
+    return ::oneapi::mkl::lapack::ungqr_batch_scratchpad_size<std::complex<float>>(
+        queue, m, n, k, lda, group_count, group_sizes);
+}
+template <>
+ONEMKL_EXPORT std::int64_t ungqr_batch_scratchpad_size<std::complex<double>>(
```
mkrainiuk: Looks like very good reuse of code! But I guess it will cost us two device checks: one in these interfaces, and one more inside Intel product. Do you think it doesn't make sense to use C API in case of cpu to skip additional device check for mklcpu?

mkrainiuk (commented):
Changes look great to me. I have several minor questions.
I don't see `RT` tests in log for GPU, do you know why they weren't included?

ericlars:
> Changes look great to me. I have several minor questions.
> I don't see `RT` tests in log for GPU, do you know why they weren't included?

RT gpu tests were skipped because of seg faults due to  [issue 67](https://github.com/oneapi-src/oneMKL/issues/67).

README.md
ericlars: LAPACK looks for ILP binaries whereas BLAS uses LP binaries. LAPACK does not have LP support.

Also, current approach is to use CBLAS and LAPACKE for testing, I think both domains would benefit from blas++ and lapack++ interfaces. An update like that could be a good time to unify the reference root.

src/lapack/backends/mkl_common/mkl_lapack.cxx
ericlars: This could work for the computational routines, but probably not the scratchpad routines. LAPACK introduced scratchpad_size() functions with oneMKL and did not create corresponding C APIs (where the call to the routine with lwork=-1 returns the scratchpad size).

mkrainiuk:
> > Changes look great to me. I have several minor questions.
> > I don't see `RT` tests in log for GPU, do you know why they weren't included?
> 
> RT gpu tests were skipped because of seg faults due to [issue 67](https://github.com/oneapi-src/oneMKL/issues/67).

Do you think we can still test them with OpenCL backend as the problem is L0 specific?

README.md
mkrainiuk: Agree, I guess we can add it to TODO list of future improvements.

ericlars:
> > > Changes look great to me. I have several minor questions.
> > > I don't see `RT` tests in log for GPU, do you know why they weren't included?
> > 
> > 
> > RT gpu tests were skipped because of seg faults due to [issue 67](https://github.com/oneapi-src/oneMKL/issues/67).
> 
> Do you think we can still test them with OpenCL backend as the problem is L0 specific?

I can. Currently the main_test.cpp shared by all domains skips gpu tests with OpenCL backend, is there reason for this?

mkrainiuk:
> > > > Changes look great to me. I have several minor questions.
> > > > I don't see `RT` tests in log for GPU, do you know why they weren't included?
> > > 
> > > 
> > > RT gpu tests were skipped because of seg faults due to [issue 67](https://github.com/oneapi-src/oneMKL/issues/67).
> > 
> > 
> > Do you think we can still test them with OpenCL backend as the problem is L0 specific?
> 
> I can. Currently the main_test.cpp shared by all domains skips gpu tests with OpenCL backend, is there reason for this?

Oh right. Yes, there is a reason for it. When we collecting device list it includes all drivers available on the machine including both L0 and OpenCL, since L0 is default backend for DPC++ we need to skip OpenCL device since it's not compatible, but I think we could improve this logic a little bit and check first what backend DPC++ uses, and after it skip not compatible device. I will think how to improve it. For now let's skip the testing.

mkrainiuk (approved):

mmeterel:
Hi @ericlars Thanks for the PR and all the great work. Did you get a chance to build and test on Windows?

cmake/FindMKL.cmake
```diff
@@ -82,13 +86,13 @@ if (ENABLE_MKLCPU_BACKEND OR ENABLE_MKLGPU_BACKEND)
   if(ENABLE_MKLCPU_THREAD_TBB)
     list(APPEND MKL_LINK_C ${TBB_LINK})
   endif()
-  if(ENABLE_MKLGPU_BACKEND)
+  if(USE_DPCPP_API)
```
mmeterel, line 89: What is the purpose of using USE_DPCPP_API workaround?

ericlars: LAPACK's cpu backend links to libmkl_sycl because of the scratchpad functions not found in the c/fortran API. This workaround was suggested by @mkrainiuk to avoid impacting other domains.

mmeterel:
@ericlars I realized that there is no documentation added for the LAPACK component (unless I missed it). Is it possible to add it in the PR as well (or in a separate PR?)? We typically add the documentation changes along with backend changes. The documentation in open source is very similar to spec documentation with necessary modifications.

ericlars:
> @ericlars I realized that there is no documentation added for the LAPACK component (unless I missed it). Is it possible to add it in the PR as well (or in a separate PR?)? We typically add the documentation changes along with backend changes. The documentation in open source is very similar to spec documentation with necessary modifications.

ericlars:
> @ericlars I realized that there is no documentation added for the LAPACK component (unless I missed it). Is it possible to add it in the PR as well (or in a separate PR?)? We typically add the documentation changes along with backend changes. The documentation in open source is very similar to spec documentation with necessary modifications.

Thanks for the catch, the only document I updated was the top level readme. Because there's no LAPACK specific deviations from the spec I think a reference and possibly a link would suffice. Do you agree?

mmeterel:
> > @ericlars I realized that there is no documentation added for the LAPACK component (unless I missed it). Is it possible to add it in the PR as well (or in a separate PR?)? We typically add the documentation changes along with backend changes. The documentation in open source is very similar to spec documentation with necessary modifications.
> 
> Thanks for the catch, the only document I updated was the top level readme. Because there's no LAPACK specific deviations from the spec I think a reference and possibly a link would suffice. Do you agree?

I believe, we keep the documentation in the open source separate than the spec, because, they might be updated at different times.

numEricL:
> > > @ericlars I realized that there is no documentation added for the LAPACK component (unless I missed it). Is it possible to add it in the PR as well (or in a separate PR?)? We typically add the documentation changes along with backend changes. The documentation in open source is very similar to spec documentation with necessary modifications.
> > 
> > 
> > Thanks for the catch, the only document I updated was the top level readme. Because there's no LAPACK specific deviations from the spec I think a reference and possibly a link would suffice. Do you agree?
> 
> I believe, we keep the documentation in the open source separate than the spec, because, they might be updated at different times.

Ok, let's add this in a separate PR.

ericlars:
> > > @ericlars I realized that there is no documentation added for the LAPACK component (unless I missed it). Is it possible to add it in the PR as well (or in a separate PR?)? We typically add the documentation changes along with backend changes. The documentation in open source is very similar to spec documentation with necessary modifications.
> > 
> > 
> > Thanks for the catch, the only document I updated was the top level readme. Because there's no LAPACK specific deviations from the spec I think a reference and possibly a link would suffice. Do you agree?
> 
> I believe, we keep the documentation in the open source separate than the spec, because, they might be updated at different times.

Ok, let's add this in a separate PR.

asrednit:
intelci: run

mmeterel (approved):

PR #90, opened 2021-05-11
[RNG] Fix mklcpu backend kernel names and clang-format to curand

### LINKED ISSUE #89: Rng mklcpu backend uses ambiguous kernel names
# Summary
In the rng mklcpu backend in `philox4x32x10.cpp` the same kernel names are generated for both the USM and buffer API. This does not cause an error when using dpc++, but according to the SYCL standard, this is illegal and causes problems with some SYCL implementations, tested with hipSYCL. 

# Observed behavior
For example, the kernels at line 328  and 68 in `philox4x32x10.cpp` appear to have the same name. This is a result of the combination of the type of `distr`  and  `philox4x32x10_impl` being used as kernel names. Since kernels are declared for both USM and buffer interfaces with the same type of `distr` this results in ambiguous kernel names.  

# Expected behavior
All kernels should have a unique name.

aelizaro:
# Description

- [x] Fix mklcpu backend kernel names ([issue 89](https://github.com/oneapi-src/oneMKL/issues/89))
- [x] Apply clang-format to curand backend
- [x] Change c++14 to c++17 cmake to avoid llvm failures 

## All Submissions

[log_ct.txt](https://github.com/oneapi-src/oneMKL/files/6461199/log_ct.txt)
[log_rt.txt](https://github.com/oneapi-src/oneMKL/files/6461202/log_rt.txt)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,7 @@ option(BUILD_SHARED_LIBS "Build dynamic libraries" ON)
 
 # Override default CXX compile/link lines for Windows after project
 if(WIN32)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-function -w")
   foreach (flag_var
            CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ option(BUILD_SHARED_LIBS "Build dynamic libraries" ON)
 
 # Override default CXX compile/link lines for Windows after project
 if(WIN32)
-  set(CMAKE_CXX_STANDARD 14)
```
mkrainiuk, line 51: Could you please check if we can remove hardcoded C++ standard version here completely? As I know we don't need it on Linux, so I wonder if we can do the same for Windows.

aelizaro: sure, removed, tests are still passing

mkrainiuk: Thank you!

mkrainiuk (approved):

NadyaTen (approved):

PR #96, opened 2021-06-01
[cublas] Add synchronization to the Iamin USM function

sbalint98:
# Description

This PR adds a missing synchronization for the `iamin` cublas USM function. 

I couldn't reproduce any issues regarding this with dpc++, however, this causes the iamin tests to fail when oneMKL is compelled by hipSYCL. Note that the same synchronization is present for the iamax USM function, which makes me believe that this has been left out by accident.

- [x] Do all unit tests pass locally? Attach a log. ~~[test.log](https://github.com/oneapi-src/oneMKL/files/6576178/test.log)~~  [test.log](https://github.com/oneapi-src/oneMKL/files/6587168/test.log)


sbalint98:
I've added the same synchronization for `sdsdot` as well.

mmeterel (approved):
@sbalint98 Thanks for the PR. The changes look good to me. I did not see issues in my local tests.

mkrainiuk (approved):
Looks good to me, thank you!

Issue #99, opened 2021-06-09
Enable using oneMKL with hipSYCL


sbalint98:
# Summary
With a set of Pull requests, we would like to upstream changes to enable the use of oneMKL BLAS with hipSYCL. The changes and added features encompass the following:
* PR[#101] in case not compiling with dpc++ use of add_sycl_to_target cmake integration -
* PR[#102] Add the option to disable the functions using half data-types
* PR[#103] Use an additional layer of abstraction when invoking host-tasks for the cuBLAS backend
* Add a hipSYCL specific cuBLAS context handler, and host-task invocation
* ~~PR[#100]~~ PR[#104]PR[#105] Replace non-standard types and member function calls, eg `half` -> `cl::sycl::half` and `.get_cl_code()` -> `.what()` -

# Problem statement
Our intention is to make oneMKL more available for the general SYCL community. These changes will allow easier integration with other SYCL implementation and enable adding rocm libraries to oneMKL. 

In case the PRs from #100, #101, #102, #104, #105 are merged, adding the actual hipSYCL support could look like this: https://github.com/sbalint98/oneMKL/tree/ustream-hipsycl-specific-changes . 

when 100-103 is merged into the current develop e8e3dabf9fbda0556b8075c76b657336f88440f0, all tests pass locally: [int_test_oneapi.log](https://github.com/oneapi-src/oneMKL/files/6624843/int_test_oneapi.log)

mmeterel:
@sbalint98 
Thanks for all the PRs. I have a general question for my better understanding. 
Could you please provide more information on what capabilities these PRs are adding? I guess I am not too familiar with hipSYCL, so it would be great if you can give some background. Do you have any future plans after these PRs? For example, will you be adding ROCm backend in the future? Are these PRs enabling that path?

sbalint98:
In general, the currently open PRs make small changes to make adding hipSYCL more elegant. After these (100-103) a last PR can be opend, that will add the actual hipSYCL specific details.

In order to enable adding hipSYCL the following changes are necessary, appart from the ones in #100:
#101 calls the `add_sycl_to_target` cmake command on all relevant targets when compiling for hipSYCL. This is going to be needed by other SYCL implementations as well. 

#102 Is required since hipSYCL does not support `cl::sycl::half` yet. It adds a CMake variable that can be toggled to disable all BLAS functions that operate on `cl::sycl::half` datatype. 

[...]

PR #101, opened 2021-06-09
Add optionaly the add_sycl_to_target cmake integration to blas and rng domains

sbalint98:
# Description
This PR adds support for using the `add_sycl_to_target` cmake integration, which is widely used by other SYCL implementations. This is necessary for enabling hipSYCL to work with oneMKL see #99 

# Checklist


CMakeLists.txt
```diff
@@ -128,7 +130,21 @@ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 if(WIN32)
   add_library(ONEMKL::SYCL::SYCL INTERFACE IMPORTED)
 else()
-  find_package(Compiler REQUIRED)
+  # Find necessary packages
+  string( TOLOWER "${ONEMKL_SYCL_IMPLEMENTATION}" ONEMKL_SYCL_IMPLEMENTATION)
+  if (ONEMKL_SYCL_IMPLEMENTATION STREQUAL "hipsycl")
```
mkrainiuk: Do you think there is a way to detect compiler automatically? At this moment `CMAKE_CXX_COMPILER` is already defined so we could try to check what type of sycl implementation it is in case if user didn't provide `ONEMKL_SYCL_IMPLEMENTATION`.

mkrainiuk (commented):
Thank you for the PR! I have one comment about potential improvement for better user experience.

CMakeLists.txt
sbalint98: I am not sure how this could be done in a straightforward way. In the case of hipSYCL, the compiler used for the compilation is not necessary will be the compiler set in `CMAKE_CXX_COMPILER`. The `add_sycl_to_target` cmake command, will invoke the `syclcc-launcher` that will, in turn, invoke the compiler configured for hipSYCL during installation (eg clang). Therefore I don't see how this could be done, easily. Pinging @illuhad to fact check this.

CMakeLists.txt
```diff
@@ -68,6 +68,8 @@ endif()
 option(ENABLE_CUBLAS_BACKEND "" OFF)
 option(ENABLE_CURAND_BACKEND "" OFF)
 option(ENABLE_NETLIB_BACKEND "" OFF)
+option(DISABLE_HALF_RUTINES "" OFF)
```
mmeterel: i dont think we need this in this PR.

CMakeLists.txt
illuhad: Sorry for the delay - it's indeed not easily possible to detect a SYCL implementation from the compiler in general. SYCL implementations have quite a lot of freedom in how they are designed, and e.g. in certain split compilation scenarios the `CMAKE_CXX_COMPILER` only plays a minor role.

The fundamental issue is that there's no global cmake infrastructure that works throughout the entire SYCL ecosystem, like a `find_package(SYCL)`. Or to be more specific, that DPC++ has not yet adopted `add_sycl_to_target()` which is used by ComputeCpp, hipSYCL and triSYCL due to the flexibility of this approach.

See here and the associated MR for a discussion of the SYCL ecosystem cmake infrastructure, and the difficulties of the assumption that SYCL compilation only requires `CMAKE_CXX_COMPILER` + special compiler flags:
https://gitlab.kitware.com/cmake/cmake/-/issues/21711

hipSYCL indeed does not necessarily use `CMAKE_CXX_COMPILER`. This actually has advantages for multi-language projects, where e.g. some part is OpenMP offload, another part SYCL, another CUDA etc.

What could work is trial-and-error, e.g. `find_package(hipSYCL)`, if that fails, assume DPC++ or vice versa.

mkrainiuk: Hi @illuhad and @sbalint98 . Sorry for the delay! Let me provide one possible suggestion how we can potentially add auto detection for compiler via `execute_process`. I rely on `CMAKE_CXX_COMPILER` to detect vendor, but if hipSYCL doesn't provide this information I guess this option wouldn't work. But if you know any other ways how to get information about vendor from hipSYCL compiler from command line please share, we can try them as well.

CMakeLists.txt
```diff
@@ -128,7 +130,21 @@ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
...
+    add_library(ONEMKL::SYCL::SYCL INTERFACE IMPORTED)
+  elseif(ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
+    message(STATUS "Looking for dpc++")
+    set(USE_ADD_SYCL_TO_TARGET_INTEGRATION false)
+    set(ENABLE_HALF_ROUTINES ON)
+    find_package(Compiler REQUIRED)
+  else()
+    message(FATAL_ERROR "SYCL implementation ${ONEMKL_SYCL_IMPLEMENTATION} is not known")
+  endif()
 endif()
```
mkrainiuk:
```suggestion
  # Find correct compiler packages
  execute_process(COMMAND ${CMAKE_CXX_COMPILER} --version ONEMKL_SYCL_IMPLEMENTATION)
  string(TOLOWER ${ONEMKL_SYCL_IMPLEMENTATION} ONEMKL_SYCL_IMPLEMENTATION)
  if(ONEMKL_SYCL_IMPLEMENTATION MATCHES ".*intel.*")
    message(STATUS "Detected SYCL compiler: Intel")
    set(USE_ADD_SYCL_TO_TARGET_INTEGRATION false)
    set(ENABLE_HALF_ROUTINES ON)
    find_package(Compiler REQUIRED)
  elseif(ONEMKL_SYCL_IMPLEMENTATION MATCHES ".*hipsycl.*")
    message(STATUS "Detected SYCL compiler: hipSYCL")
    find_package(hipSYCL CONFIG REQUIRED)
    set(USE_ADD_SYCL_TO_TARGET_INTEGRATION true)
    add_library(ONEMKL::SYCL::SYCL INTERFACE IMPORTED)
  else()
    message(FATAL_ERROR "SYCL compiler implementation ${ONEMKL_SYCL_IMPLEMENTATION} is not known")
  endif()
endif()
```

CMakeLists.txt
illuhad: Thanks for the suggestion, @mkrainiuk. I'm not sure how well this can work. hipSYCL supports multiple compilation flows, in particular it can also act as a mere library for third-party compilers when targeting CPU exclusively. In this case, it can happen that the underlying compiler changes depending on which targets are selected. For example:

```
$ syclcc --hipsycl-targets="omp" --version
c++ (GCC) 11.1.0
Copyright (C) 2021 Free Software Foundation, Inc.
Dies ist freie Software; die Kopierbedingungen stehen in den Quellen. Es
gibt KEINE Garantie; auch nicht für MARKTGÄNGIGKEIT oder FÜR SPEZIELLE ZWECKE.
$ syclcc --hipsycl-targets="hip:gfx908" --version
clang version 11.1.0
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

cmake tends to not like it when the compiler suddenly changes depending on compiler arguments (e.g. this might also cause certain compiler specific flags to suddenly be not available anymore), which is why it is generally *not* considered good practice to set `CMAKE_CXX_COMPILER` to hipSYCL's `syclcc` even though it might work for some use cases. The better approach is to just use `add_sycl_to_target()` without changing `CMAKE_CXX_COMPILER`. As such, relying on `CMAKE_CXX_COMPILER` for implementation detection is probably not an ideal strategy.

The better approach in the long run in my opinion would probably be for the SYCL implementation ecosystem to come together and have a unified `FindSYCL.cmake` which checks for available SYCL implementations in the filesystem and then provides a unified interface to compile targets with SYCL. But we are still some time away from this, and while this is the case a cmake configuration flag to select the implementation as done here is probably the most straight-forward solution.

mkrainiuk: Thanks @illuhad for detailed explanation! So it's more DPC++ compiler problem, because user has to source DPC++ compiler and define it as cmake compiler, one more variable looks like an overkill. What do you think about having autodetection mechanism for DPC++ compiler and if it's not DPC++ compiler use hipSYCL check? The proposed option is also nice to have to force compiler without check, but I'd prefer to make it optional for better user experience.

illuhad: @mkrainiuk yes, I think it could work to first try to automatically detect DPC++ from `CMAKE_CXX_COMPILER`, and if that fails, try `find_package(hipSYCL)`.

I think we should still add an optional option to force a specific compiler because I expect that the line between DPC++ and mainline clang will get blurry in the future as more components from DPC++ get upstreamed. In the future, there will probably be 
* people using regular clang, who want to run the SYCL path for DPC++
* and people who happen to use clang (with SYCL support) as `CMAKE_CXX_COMPILER` (e.g. because it's the default C++ compiler), but still wish to use hipSYCL.

mkrainiuk: Sure, I'm fine with new option for advanced users, but we need some default detection for easier way to start using this project.
One more question out of curiosity, how hipSYCL plans to support several compilation runs for host and kernels without special compiler? As far as I know standard compiler can't do host/kernels compilation from one source.

illuhad: > Sure, I'm fine with new option for advanced users, but we need some default detection for easier way to start using this project.

Yes, I think we have alignment here.

> One more question out of curiosity, how hipSYCL plans to support several compilation runs for host and kernels without special compiler? As far as I know standard compiler can't do host/kernels compilation from one source.

Currently, as soon as non-CPU devices are targeted as well, hipSYCL forces compilation with clang and hipSYCL's custom clang plugin to perform kernel outlining and device compilation etc. Currently using third-party compilers is only possible when targeting CPU exclusively, although we may also enable this in the future using nvc++ or OpenMP targets.

sbalint98: I rebased to develop and updated with the required logic

src/blas/backends/cublas/CMakeLists.txt
```diff
@@ -20,17 +20,16 @@
...
 set(LIB_OBJ ${LIB_NAME}_obj)
 find_package(cuBLAS REQUIRED)
-
+set(SOURCES cublas_level1.cpp 
+                cublas_level2.cpp 
+                cublas_level3.cpp 
+                cublas_batch.cpp 
+                cublas_extensions.cpp 
+                cublas_scope_handle.cpp 
+                $<$<BOOL:${BUILD_SHARED_LIBS}>: mkl_blas_cublas_wrappers.cpp>)
```
mkrainiuk: Please fix the file name to `cublas_wrappers.cpp`

src/rng/backends/curand/CMakeLists.txt
```diff
@@ -60,13 +60,14 @@ set(LIB_NAME onemkl_rng_curand)
 set(LIB_OBJ ${LIB_NAME}_obj)
 find_package(cuRAND REQUIRED)
 
-add_library(${LIB_NAME})
-add_library(${LIB_OBJ} OBJECT
-  philox4x32x10.cpp
+set(SOURCES philox4x32x10.cpp
   mrg32k3a.cpp
-  $<$<BOOL:${BUILD_SHARED_LIBS}>: mkl_rng_curand_wrappers.cpp>
+  $<$<BOOL:${BUILD_SHARED_LIBS}>: mkl_rng_curand_wrappers.cpp>)
```
mkrainiuk, line 65: Extra `)`

sbalint98: thanks for catching these, I forgot to build the cublas and curand backend apparently while testing.

mkrainiuk (approved):

cmake/FindCompiler.cmake
```diff
@@ -33,13 +33,13 @@ if(is_dpcpp)
   if(UNIX)
     set(UNIX_INTERFACE_COMPILE_OPTIONS -fsycl)
     set(UNIX_INTERFACE_LINK_OPTIONS -fsycl)
-    if(ENABLE_CURAND_BACKEND)
+    if(ENABLE_CURAND_BACKEND OR ENABLE_CUBLAS_BACKEND)
```
mmeterel: Why do we need this addition?

tests/unit_tests/blas/batch/CMakeLists.txt
```diff
@@ -32,6 +32,11 @@ if(BUILD_SHARED_LIBS)
       PUBLIC ${CBLAS_INCLUDE}
   )
   target_link_libraries(blas_batch_rt PUBLIC ONEMKL::SYCL::SYCL)
```
mmeterel: I think, you need to remove this line.

tests/unit_tests/blas/batch/CMakeLists.txt
```diff
@@ -45,3 +50,8 @@ target_include_directories(blas_batch_ct
     PUBLIC ${CBLAS_INCLUDE}
 )
 target_link_libraries(blas_batch_ct PUBLIC ONEMKL::SYCL::SYCL)
```
mmeterel: I think, you need to remove this line.

tests/unit_tests/blas/extensions/CMakeLists.txt
```diff
@@ -32,6 +32,11 @@ if(BUILD_SHARED_LIBS)
       PUBLIC ${CBLAS_INCLUDE}
   )
   target_link_libraries(blas_extensions_rt PUBLIC ONEMKL::SYCL::SYCL)
```
mmeterel: I think, you need to remove this line.

tests/unit_tests/blas/extensions/CMakeLists.txt
```diff
@@ -45,3 +50,8 @@ target_include_directories(blas_extensions_ct
     PUBLIC ${CBLAS_INCLUDE}
 )
 target_link_libraries(blas_extensions_ct PUBLIC ONEMKL::SYCL::SYCL)
```
mmeterel: I think, you need to remove this line.

tests/unit_tests/rng/service/CMakeLists.txt
```diff
@@ -43,3 +47,8 @@ target_include_directories(rng_service_ct
     PUBLIC ${CMAKE_BINARY_DIR}/bin
 )
 target_link_libraries(rng_service_ct PUBLIC ONEMKL::SYCL::SYCL)
```
mmeterel: I think, you need to remove this line.

cmake/FindCompiler.cmake
sbalint98: I added this because I wanted to have the `-fsycl-targets=nvptx64-nvidia-cuda-sycldevice` and `-fsycl-unnamed-lambda` flags defined when targeting the cublas backend and compiling with dpc++. I don't see anywhere else where these flags could be added to the link and the compile line.

tests/unit_tests/blas/batch/CMakeLists.txt
sbalint98: thanks! you're right

cmake/FindCompiler.cmake
mmeterel: Right, I see that. My question was why do we need those flags for cuBLAS backend now? I am curious why we did not need them before and need them now.

sbalint98: Right I will need to look into why this change was necessary. I don't think that this change belongs to this PR necessarily, so I'll remove it from here and submit a new PR with a more detailed description of the issue at hand

sbalint98: Rebased to develop, and removed `ENABLE_CUBLAS_BACKEND` from this PR. I will open a separate PR with a more detailed explanation of this change

mmeterel (approved):

PR #103, opened 2021-06-09
Add abstraction for invocation of interop_task

sbalint98:
# Description
This PR abstracts away the `interop_task` calls inside the cuBLAS functions. This approach is intended to be very similar to the one for the MKLcpu functions. Putting the actual call inside the `cublas_task.hpp` allows us to cleanly define an alternative if we want to use a different interface. 

This PR also changes the interface for `CublasScopedContextHandler`. The `interop_handler` is passed with the constructor; this enables changing the  `CublasScopedContextHandler` implementation to be easily switched among different SYCL implementations.

For example, in the case of hipSYCL, this PR enables the following pattern: https://github.com/sbalint98/oneMKL/blob/90598195961ce6cfa4c4a6723176611a2a5487ea/src/blas/backends/cublas/cublas_task.hpp

related to #99 

## All Submissions


mkrainiuk (approved):
Thank you for PR! Looks good to me.

mmeterel:
Looks good to me.

mmeterel (approved):

PR #104, opened 2021-06-09
Replace get_cl_code() with what()

sbalint98:
# Description
`get_cl_code()` is not part of the SYCL standard, therefore this PR replaces it with `what()`

related to #99 

# Checklist

### DIFF
--- a/tests/unit_tests/blas/batch/axpy_batch_usm.cpp
+++ b/tests/unit_tests/blas/batch/axpy_batch_usm.cpp
@@ -54,7 +54,7 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t group_count) {
             catch (exception const &e) {
                 std::cout << "Caught asynchronous SYCL exception during AXPY_BATCH:\n"
                           << e.what() << std::endl
-                          << "OpenCL status: " << e.get_cl_code() << std::endl;
+                          << "OpenCL status: " << e.what() << std::endl;
             }
         }
     };
@@ -186,7 +186,7 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t group_count) {
     catch (exception const &e) {
         std::cout << "Caught synchronous SYCL exception during AXPY_BATCH:\n"
                   << e.what() << std::endl
-                  << "OpenCL status: " << e.get_cl_code() << std::endl;
+                  << "OpenCL status: " << e.what() << std::endl;
     }
 
     catch (const oneapi::mkl::unimplemented &e) {
--- a/tests/unit_tests/blas/batch/axpy_batch_usm.cpp
+++ b/tests/unit_tests/blas/batch/axpy_batch_usm.cpp
@@ -53,8 +53,8 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t group_count) {
             }
             catch (exception const &e) {
                 std::cout << "Caught asynchronous SYCL exception during AXPY_BATCH:\n"
-                          << e.what() << std::endl
-                          << "OpenCL status: " << e.what() << std::endl;
+                          << e.what() << std::endl;
+                print_error_code(e);
             }
         }
     };
@@ -185,8 +185,8 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t group_count) {
     }
     catch (exception const &e) {
         std::cout << "Caught synchronous SYCL exception during AXPY_BATCH:\n"
-                  << e.what() << std::endl
-                  << "OpenCL status: " << e.what() << std::endl;
+                  << e.what() << std::endl;
+        print_error_code(e);
     }
 
     catch (const oneapi::mkl::unimplemented &e) {
--- a/tests/unit_tests/blas/extensions/gemmt.cpp
+++ b/tests/unit_tests/blas/extensions/gemmt.cpp
@@ -76,7 +76,7 @@ int test(device* dev, oneapi::mkl::layout layout, oneapi::mkl::uplo upper_lower,
             catch (exception const& e) {
                 std::cout << "Caught asynchronous SYCL exception during GEMMT:\n"
                           << e.what() << std::endl
-                          << "OpenCL status: " << e.get_cl_code() << std::endl;
+                          << "OpenCL status: " << e.what() << std::endl;
             }
         }
     };
@@ -121,7 +121,7 @@ int test(device* dev, oneapi::mkl::layout layout, oneapi::mkl::uplo upper_lower,
     catch (exception const& e) {
         std::cout << "Caught synchronous SYCL exception during GEMMT:\n"
                   << e.what() << std::endl
-                  << "OpenCL status: " << e.get_cl_code() << std::endl;
+                  << "OpenCL status: " << e.what() << std::endl;
     }
 
     catch (const oneapi::mkl::unimplemented& e) {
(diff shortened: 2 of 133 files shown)


tests/unit_tests/blas/batch/axpy_batch_usm.cpp
```diff
@@ -54,7 +54,7 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t group_count) {
             catch (exception const &e) {
                 std::cout << "Caught asynchronous SYCL exception during AXPY_BATCH:\n"
                           << e.what() << std::endl
-                          << "OpenCL status: " << e.get_cl_code() << std::endl;
+                          << "OpenCL status: " << e.what() << std::endl;
```
mkrainiuk: I guess with new exception method first part of the error message `OpenCL status:` will be a little bit misleading. Should it be removed completely in this case?

mkrainiuk (commented):
Thank you for PR! I have one comment about potential improvement.

tests/unit_tests/blas/batch/axpy_batch_usm.cpp
sbalint98: What do you think about using `Error status`? I think it would be more elegant?

mmeterel: I don't think it makes sense to print e.what() twice. 
Also, I see get_cl_code is part of SYCL 1.2.1 spec (section 4.9). (Please correct me if I am missing something)
Anything you can do to support this from hipSYCL side? It can be a dummy call?
I think, providing the OpenCL error status is a valuable debug feature for DPC++

illuhad: It's part of the SYCL 1.2.1 OpenCL interoperability API which doesn't make sense in hipSYCL since hipSYCL does not have an OpenCL backend. This API was removed in SYCL 2020. We could add a dummy call if this is a blocker, but I'd prefer not to. 
I think the better strategy might be to switch to the SYCL 2020 exception interface in the mid term, which provides a replacement for `get_cl_code()`: https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:exception.class

mkrainiuk: I'm fine with using `Error status: e.code()`. Let me check if it works with DPC++ compiler as well.
Update: looks like this part is not there yet, can we wrap get_cl_code part to use only for DPC++ compiler for now and remove it later when both compilers will be able to support e.code()?

sbalint98: I've added the wrapper for the `get_cl_code`. note that sycl2020 exception handling is still missing. So we probably want to add that before this is merged.

mkrainiuk: Thank you! Looks good to me. Could you please clarify what exact changes we need for sycl2020 exception handling?

sbalint98: I am sorry for the confusion I meant we need to add the exception handling to hipSYCL, but that should be handled by this PR https://github.com/illuhad/hipSYCL/pull/595 . I don't think that there is any modification necessary from the side of oneMKL or dpc++

mkrainiuk: Oh, I see, thank you!

mmeterel: Looks good to me. Could you please rebase and add similar changes to newer tests?

mkrainiuk: Could you please re-run clang-format one more time? I've just noticed it doesn't like how the changes look after `print_error_code` function introduction.

mkrainiuk: Hi @sbalint98, did you have a chance to re-run clang-format for this PR? Everything else look good to me for integration.

sbalint98: yes, I did that, before rebasing. I also checked it again, everything seems to be well formatted.

mkrainiuk (approved):

tests/unit_tests/blas/extensions/gemmt.cpp
```diff
@@ -119,9 +119,8 @@ int test(device* dev, oneapi::mkl::layout layout, oneapi::mkl::uplo upper_lower,
 #endif
     }
     catch (exception const& e) {
-        std::cout << "Caught synchronous SYCL exception during GEMMT:\n"
-                  << e.what() << std::endl
-                  << "OpenCL status: " << e.get_cl_code() << std::endl;
+        std::cout << "Caught synchronous SYCL exception during GEMMT:\n" << e.what() << std::endl;
```
mmeterel, line 122: Let's keep the same format if possible.

tests/unit_tests/blas/extensions/gemmt_usm.cpp
```diff
@@ -121,9 +121,8 @@ int test(device* dev, oneapi::mkl::layout layout, oneapi::mkl::uplo upper_lower,
 #endif
     }
     catch (exception const& e) {
-        std::cout << "Caught synchronous SYCL exception during GEMMT:\n"
-                  << e.what() << std::endl
-                  << "OpenCL status: " << e.get_cl_code() << std::endl;
+        std::cout << "Caught synchronous SYCL exception during GEMMT:\n" << e.what() << std::endl;
```
mmeterel, line 124: Same here

tests/unit_tests/blas/level1/asum.cpp
```diff
@@ -101,9 +101,8 @@ int test(device* dev, oneapi::mkl::layout layout, int64_t N, int64_t incx) {
 #endif
     }
     catch (exception const& e) {
-        std::cout << "Caught synchronous SYCL exception during ASUM:\n"
-                  << e.what() << std::endl
-                  << "OpenCL status: " << e.get_cl_code() << std::endl;
+        std::cout << "Caught synchronous SYCL exception during ASUM:\n" << e.what() << std::endl;
```
mmeterel, line 104: same here

tests/unit_tests/blas/extensions/gemmt.cpp
sbalint98: If I understand correctly you would like to have the `<< e.what() << std::endl;` in a new line, like for example in  tests/unit_tests/blas/batch/axpy_batch_stride.cpp, is that correct? Changing this consistently would entail changing 104 occasions and it wouldn't be clang-format conformant.

mmeterel: Why wouldn't it be clang-format conformant? The code before your change was clang-format conformant. 
The concern here is, in case we want to do some global changes in all tests, we typically run some script and not having the same format across tests will break the script. Then, someone will have to go fix them.

sbalint98: I think it is because we have made the error printing statement much shorter by printing the OpenCL status in the `print_error_code` function. This makes the line break necessary for only the functions with longer names, such as the batched functions. For all the other tests, there is no line break. 

Should we add the line break for every function, or rather delete the new line for the few longer lines ~~(batched functions mainly)~~? I think either way the code won't adhere to the clang-format, but removing the new lines for the batched functions probably results in fewer errors. 

EDIT:
The line break seems to be triggered by the 1 character difference between asynchronous and synchronous

mmeterel: Thanks for looking at it in more detail. Based on your findings, i think we can try to handle this case in the script.

mmeterel (approved):

PR #105, opened 2021-06-09
Use appropiately qualified names in blas and rng

sbalint98:
# Description
Some types in the blas and rng domain did not have properly qualified names. This has caused trouble when using different SYCL implementations. This PR fixes this for these two domains. 

related to #99

## All Submissions


include/oneapi/mkl/detail/backend_selector.hpp
```diff
@@ -27,6 +27,8 @@
 namespace oneapi {
 namespace mkl {
 
+using namespace cl;
```
mkrainiuk, line 30: namespace cl is not part of SYCL 2020 anymore, is it still required for hipSYCL?

mkrainiuk (commented):
Thank you for PR! I have one comment regarding these changes.

include/oneapi/mkl/detail/backend_selector.hpp
sbalint98: In case the `CL/sycl.hpp` header is included only the `cl::sycl::` namespace is present in hipSYCL. Changing the includes to include `sycl/sycl.hpp` does not seem to affect dpc++, and that way we wouldn't need to use the `cl::` namespace. 
EDIT: Would you agree with changing the includes project-wide, or you'd prefer it to be changed only where it is necessary?

include/oneapi/mkl/blas.hxx
```diff
@@ -263,19 +263,22 @@ static inline void gemm(cl::sycl::queue &queue, transpose transa, transpose tran
 }
 
 static inline void gemm(cl::sycl::queue &queue, transpose transa, transpose transb, std::int64_t m,
-                        std::int64_t n, std::int64_t k, half alpha, cl::sycl::buffer<half, 1> &a,
-                        std::int64_t lda, cl::sycl::buffer<half, 1> &b, std::int64_t ldb, half beta,
-                        cl::sycl::buffer<half, 1> &c, std::int64_t ldc) {
+                        std::int64_t n, std::int64_t k, cl::sycl::half alpha,
+                        cl::sycl::buffer<cl::sycl::half, 1> &a, std::int64_t lda,
```
mmeterel: Initially, we were using a local implementation for half precision. We will be moving to sycl::half soon. You can remove these changes from PR I think. we would like to do the changes globally and validate them with CI.

src/rng/backends/mklcpu/mrg32k3a.cpp
```diff
@@ -33,6 +33,8 @@ namespace mkl {
 namespace rng {
 namespace mklcpu {
 
+using namespace cl;
```
mmeterel, line 36: Adding @aelizaro

tests/unit_tests/rng/include/statistics_check_test.hpp
```diff
@@ -44,6 +44,7 @@
 
 #define POISSON_ARGS 0.5
 
+using namespace cl;
```
mmeterel, line 47: Adding @aelizaro

include/oneapi/mkl/detail/backend_selector.hpp
illuhad: @mkrainiuk The `CL/sycl.hpp` header is deprecated in SYCL 2020 and only there for backwards compatibility. The SYCL specification as it is currently worded only says:

> For compatibility with SYCL 1.2.1, SYCL provides another standard header file: <CL/sycl.hpp>, which can be included in place of <sycl/sycl.hpp>. In that case, all SYCL classes, constants, types and functions defined by this specification should exist within the ::cl::sycl C++ namespace.
> For consistency, the programming API will only refer to the <sycl/sycl.hpp> header and the ::sycl namespace, but this should be considered synonymous with the SYCL 1.2.1 header and namespace.

The problem is that when including `CL/sycl.hpp`, DPC++ exposes `::sycl` as well, which is arguably incorrect, although the specification is not super clear here. hipSYCL follows a more literal interpretation of this section of the specification where `CL/sycl.hpp` exposes `::cl::sycl` and `sycl/sycl.hpp` exposes `::sycl`.

See here for a discussion of exactly this issue: https://github.com/KhronosGroup/SYCL-CTS/pull/108

I agree that the proper solution is to move to the SYCL 2020 header `sycl/sycl.hpp` and use `::sycl`, but I'm not sure how well this is already supported in DPC++.

mkrainiuk: Thanks @sbalint98 and @illuhad! Sorry for delay. I agree that we need to move to SYCL 2020, I also checked sycl/sycl.hpp in DPC++ compiler, and as I see it's not fully supported yet, so I'm fine with having cl namespace, we can change this part when both hipSYCL and DPC++ compilers will fully move to SYCL 2020.

tests/unit_tests/blas/include/test_common.hpp
```diff
@@ -29,8 +29,8 @@
 #include <CL/sycl.hpp>
 
 namespace std {
-static cl::sycl::half abs(cl::sycl::half v) {
-    if (v < cl::sycl::half(0))
+static half abs(half v) {
+    if (v < half(0))
```
mkrainiuk: Shouldn't the half precision changes be part of https://github.com/oneapi-src/oneMKL/pull/102?

sbalint98: Sorry, I think this shouldn't be a part of any PR. I accidentally removed these with all the other `cl::sycl::` qualifiers from before the half type as asked for by @mmeterel

mmeterel: @sbalint98 Can you remind why you needed to qualify half with cl::sycl:: namespaces to start with?

sbalint98: In hipSYCL we define half inside that namespace, and if I am correct it is also required by the spec that half is defined in the sycl namespace. 

In table 138. in section 4.14.1 `half` is listed as 

> Additional scalar data types which are supported by SYCL within the sycl namespace

mkrainiuk (approved):

include/oneapi/mkl/blas/detail/cublas/blas_ct.hxx
```diff
@@ -767,7 +767,6 @@ void gemm(backend_selector<backend::cublas> selector, transpose transa, transpos
     gemm_postcondition(selector.get_queue(), transa, transb, m, n, k, alpha, a, lda, b, ldb, beta,
                        c, ldc);
 }
-
```
mmeterel: Nit: Could you please remove these unnecessary changes? It occurs at multiple places.

src/blas/blas_loader.cpp
```diff
@@ -3495,7 +3495,6 @@ void gemm(oneapi::mkl::device libkey, cl::sycl::queue &queue, transpose transa,
     function_tables[libkey].row_major_zgemm_sycl(queue, transa, transb, m, n, k, alpha, a, lda, b,
                                                  ldb, beta, c, ldc);
 }
-
```
mmeterel: same here

tests/unit_tests/blas/include/reference_blas_templates.hpp
```diff
@@ -291,7 +291,6 @@ template <typename fpa, typename fpc>
 static void gemm(CBLAS_LAYOUT layout, CBLAS_TRANSPOSE transa, CBLAS_TRANSPOSE transb, const int *m,
                  const int *n, const int *k, const fpc *alpha, const fpa *a, const int *lda,
                  const fpa *b, const int *ldb, const fpc *beta, fpc *c, const int *ldc);
-
```
mmeterel: same here

tests/unit_tests/rng/include/statistics_check_test.hpp
mmeterel: @aelizaro looks like all real changes are in RNG domain. Could you please review?

mmeterel:
@sbalint98 Could you please resolve the conflicts?

tests/unit_tests/rng/include/statistics_check_test.hpp
aelizaro: Thanks @mmeterel, looks good to me!

aelizaro (approved):

PR #108, opened 2021-07-13
[cmake] temporary disable sycl 2020 deprecations warnings

aelizaro:
# Description
Temporary disable deprecations warnings for sycl 2020 which happens in build with llvm compiler, as functionality is not yet available in dpcpp compiler, so can't switch to the new version. Example: get_count for sycl::buffer

# Checklist

## All Submissions


mkrainiuk (approved):
Thank you! Looks good to me, but we need to make sure we will update affected functions before they will be completely removed.

mmeterel (approved):
@aelizaro Thanks for the PR. Can we open an issue to remove this workaround when everything is in place?

aelizaro:
sure, @mmeterel, a good point, I have created #110 issue.

PR #109, opened 2021-07-13
Add new BLAS APIs and functional tests

mmeterel:
# Description

This PR is adding new BLAS APIs along with their functional tests. 
It also fixes a few typos and does minor refactoring for better readability. 

# Checklist

## All Submissions

- [X] Do all unit tests pass locally? Please see CI.

## New interfaces

The tests in oneMKL are used by customers for regression testing, hence these new APIs will improve their coverage. 
oneAPI Spec will be updated after this PR. 

## New features

- [X] Have you provided motivation for adding a new feature?
- [X] Have you added relevant tests?


include/oneapi/mkl/bfloat16.hpp
```diff
@@ -0,0 +1,228 @@
...
+#define _BFLOAT16_HPP__
+
+#include <cmath>
+#include <cstdint>
+#include <type_traits>
+
+namespace oneapi {
+namespace mkl {
+
+namespace bfloat16_impl {
```
mkrainiuk, line 30: As far as I know there is no C++ language standard for bf16, so users can have their own custom type to work with bf16 feature. With introducing public bf16 type here users will have to cast their bf16 types to this one to work with oneMKL and back to work with custom kernels. Did you consider some void* approach to simplify this work for users?

mkrainiuk (commented):
Thanks for the PR! Changes look good to me. But I have one question about new data type that this PR introduces.

src/blas/backends/cublas/cublas_batch.cpp
```diff
@@ -197,46 +584,282 @@ inline cl::sycl::event gemm_batch(Func func, cl::sycl::queue &queue, transpose *
...
+                                  get_cublas_operation(trans[i]),
+                                  get_cublas_diag_type(unit_diag[i]), (int)m[i], (int)n[i],
+                                  (cuDataType *)&alpha[i], a_ + offset, (int)lda[i], b_ + offset,
+                                  (int)ldb[i], (int)group_size[i]);
+                offset += group_size[i];
+            }
+        });
+    });
+    return done;
+}
```
lhuot, line 662: Why did you choose to implement the group batch API for TRSM batch with a loop around cublas trsm_batch calls but not for other functions for which you could loop around the base function for example axpy_batch loop around axpy calls?

src/blas/backends/mklcpu/mklcpu_batch.cxx
```diff
@@ -19,399 +19,1999 @@
...
     queue.submit([&](cl::sycl::handler &cgh) {
-        auto a_acc = a.get_access<cl::sycl::access::mode::read>(cgh);
-        auto b_acc = b.get_access<cl::sycl::access::mode::read>(cgh);
-        auto c_acc = c.get_access<cl::sycl::access::mode::read_write>(cgh);
+        auto accessor_x = x.template get_access<cl::sycl::access::mode::read>(cgh);
+        auto accessor_y = y.template get_access<cl::sycl::access::mode::read_write>(cgh);
+        host_task<class mkl_kernel_scopy_batch_strided>(cgh, [=]() {
+            int64_t i;
+            for (i = 0; i < batch_size; i++) {
+                ::cblas_scopy(n, accessor_x.get_pointer() + i * stridex, incx,
```
lhuot, line 31: why not using copy_batch DPC++ API from oneMKL on CPU?

src/blas/backends/mklcpu/mklcpu_batch.cxx
```diff
@@ -602,244 +2537,585 @@ cl::sycl::event gemm_batch(cl::sycl::queue &queue, transpose transa, transpose t
...
+            }
+            for (int64_t i = 0; i < batch_size; i++) {
+                if (i == 0) {
+                    a_array[0] = (float *)a;
+                    b_array[0] = (float *)b;
+                }
+                else {
+                    a_array[i] = a_array[i - 1] + stride_a;
+                    b_array[i] = b_array[i - 1] + stride_b;
+                }
```
lhuot: what about something like a_array[i] = ((float *) a) + i * stride_a; so you do not need the if?

src/blas/backends/mklgpu/mklgpu_batch.cxx
```diff
@@ -19,160 +19,281 @@
...
+                int64_t incx, std::int64_t stridex, cl::sycl::buffer<std::complex<float>, 1> &y,
+                int64_t incy, std::int64_t stridey, std::int64_t batch_size) {
+    ::oneapi::mkl::gpu::ccopy_batch_sycl(&queue, n, &x, incx, stridex, &y, incy, stridey,
+                                         batch_size);
+}
+
+void copy_batch(cl::sycl::queue &queue, int64_t n, cl::sycl::buffer<std::complex<double>, 1> &x,
+                int64_t incx, std::int64_t stridex, cl::sycl::buffer<std::complex<double>, 1> &y,
+                int64_t incy, std::int64_t stridey, std::int64_t batch_size) {
+    ::oneapi::mkl::gpu::zcopy_batch_sycl(&queue, n, &x, incx, stridex, &y, incy, stridey,
```
lhuot, line 46: Why using the internal oneMKL API here and not the external API? 
There is no guarantee the internal oneMKL API won't change without notice and then this code might break. Also it is not helping for external contribution.

tests/unit_tests/blas/batch/axpy_batch_stride.cpp
```diff
@@ -0,0 +1,206 @@
...
+template <typename fp>
+int test(device *dev, oneapi::mkl::layout layout, int64_t incx, int64_t incy, fp alpha,
+         int64_t batch_size) {
+    // Prepare data.
+    int64_t n, i;
+
+    n = 1357;
+
+    int64_t stride_x, stride_y;
+    stride_x = n * std::abs(incx);
```
lhuot, line 55: should we tests for stride smaller than the array size for example 0?

src/blas/backends/cublas/cublas_batch.cpp
mmeterel: Here, my goal was to take advantage of the fact that trsm_batch is supported in CUDA. (Other routines like axpy does not have batch implementations in CUDA) And since trsm_batch in CUDA requires only 1 group, I had to add a loop to handle multiple groups. (Similar to what has been done for gemm_batch)

lhuot: Yes i understand for trsm_batch but why not doing it for other functions? There is no need for batch version in cublas to implement something similar.

src/blas/backends/mklcpu/mklcpu_batch.cxx
mmeterel: That could potentially work, I guess. Do you know what it was implemented this way in oneMKL?

lhuot: There is no cblas_?copy_batch in oneMKL so we cannot call it directly when defining the DPC++ copy_batch API.
Here you have access to the DPC++ copy_batch API directly so you do not need to reimplement what is done in oneMKL.

src/blas/backends/cublas/cublas_batch.cpp
mmeterel: I don't think there is any technical reason for not doing it. It can be added by community. It is not in my priority list. 

Edit: Rephrasing priority list: My goal here was to support the batch APIs that are already supported in cuBLAS. If cuBLAS does not support a batched version, I don't think it is that important to support it in cublas backend of oneMKL interfaces.

tests/unit_tests/blas/batch/axpy_batch_stride.cpp
mmeterel: Yes, we need to extend the coverage in our functional tests for different stride options and also for malloc_device in USM APIs. However, I suggest doing it for all tests as a separate PR.

src/blas/backends/mklgpu/mklgpu_batch.cxx
mmeterel: Right, that is my preference as well due to same reasons you mentioned that I had to work around before. We had some discussions about this before but did not reach to a conclusion. One concern with this could be having the exact same function signature for oneMKL and oneMKL open source interfaces at runtime. 
@mkrainiuk Do you have more info about this discussion?

Nevertheless, this change should be done as a separate PR since it impact all existing APIs as well.

src/blas/backends/mklcpu/mklcpu_batch.cxx
mmeterel: That would create inconsistency though (and additional dependency on mkl_sycl library?) For all CPU implementations (L1, L2, L3...) we are calling cblas_* interfaces from oneMKL.

src/blas/backends/mklcpu/mklcpu_batch.cxx
mmeterel: Sure that will work. I will update.

include/oneapi/mkl/bfloat16.hpp
mmeterel: That is a good suggestion. My concern is, I am not sure how to test it properly without a sample custom type. Maybe do this change when/if we need it?

src/blas/backends/mklcpu/mklcpu_batch.cxx
lhuot: Yes i was thinking this should probably be done for all functions :-) 
So for CPU backend there is some requirement to not have mkl_sycl library dependencies?

mmeterel: I am not aware of such requirement to not have mkl_sycl library dependency for mklcpu backend. @mkrainiuk do you know? (I was just pointing out the additional dependency :) )

Overall, I agree with your proposal. It will simplify the code. Since it is impacting all APIs, I think we should do this separately. :) We can bundle this with using DPC++ external APIs for mklgpu backend as well.

mkrainiuk: Initially we enabled mklcpu blas backend without dependencies on mkl_sycl library because opensource compiler runtime and Intel DPC++runtime were not aligned, now it's better, but mkl_sycl library still doesn't fully support opensource compiler runtime, so you can see LAPACK domain failures because it uses mkl_sycl in mklcpu.

include/oneapi/mkl/bfloat16.hpp
mkrainiuk: Can we add this custom type to our tests only and refer to them as an expected type to use for new API?

src/blas/backends/mklgpu/mklgpu_batch.cxx
mkrainiuk: Right, the main problem with using Intel oneMKL external API is a potential recursion, because opensource wrapper can resolve call to Intel oneMKL as opensource oneMKL symbol again and go to the same wrapper.

src/blas/backends/mklcpu/mklcpu_batch.cxx
mmeterel: Updated. Local tests pass.

include/oneapi/mkl/bfloat16.hpp
mkrainiuk: Since Intel oneMKL uses this type, I guess we don't have other choices but define the same type in opensource oneMKL.

src/blas/backends/mklcpu/mklcpu_batch.cxx
mmeterel: Thanks @mkrainiuk looks like we are not ready to switch.

mkrainiuk (approved):

src/blas/backends/mklgpu/mklgpu_batch.cxx
lhuot: Is there some tracker open for this?

tests/unit_tests/blas/batch/axpy_batch_stride.cpp
lhuot: Is there some tracker open for this?

mmeterel: Thanks for reminding. Opened an issue #113

src/blas/backends/mklgpu/mklgpu_batch.cxx
mmeterel: Thanks for reminding. Opened an issue: #114

lhuot (approved):

PR #120, opened 2021-08-19
Add missing include to mklcpu_batch.cpp

sbalint98:
# Description

This is a fix for the bug introduced in #102 that prevents compiling the blas domain. In #102 `exceptions.hpp` was not included even though it is used in this file.


mmeterel (approved):
I confirm that the develop is broken without this PR. Approved.

vmalia (approved):

PR #122, opened 2021-08-24
Add hipSYCL scope_handle and host_task

sbalint98:
# Description

This PR adds hipSYCL support to oneMKL by providing a specific scope_handler. As discussed [here](https://github.com/oneapi-src/oneMKL/pull/105#discussion_r673342827) hipSYCL has the `half` datatype in the `cl::sycl::` namespace; Therore to compile successfully the following patch needs to be applied: [half.txt](https://github.com/oneapi-src/oneMKL/files/7038184/half.txt)

# Checklist

## All Submissions

[hipsycl_cuda_host.log](https://github.com/oneapi-src/oneMKL/files/7038238/hipsycl_cuda_host.log)
[dpcpp-hand-test.log](https://github.com/oneapi-src/oneMKL/files/7038742/dpcpp-hand-test.log)

### DIFF
--- a/src/blas/backends/cublas/cublas_handle.hpp
+++ b/src/blas/backends/cublas/cublas_handle.hpp
@@ -16,8 +16,8 @@
 *  limitations under the License.
 *
 **************************************************************************/
-#ifndef CUBLAS_HANDLE_HIPSYCL_HPP
-#define CUBLAS_HANDLE_HIPSYCL_HPP
+#ifndef CUBLAS_HANDLE_HPP
+#define CUBLAS_HANDLE_HPP
 #include<atomic>
 #include<unordered_map>
 
@@ -53,4 +53,4 @@ struct cublas_handle {
 } // namespace mkl
 } // namespace oneapi
 
-#endif // CUBLAS_HANDLE_HIPSYCL_HPP
+#endif // CUBLAS_HANDLE_HPP
--- a/src/blas/backends/cublas/cublas_handle.hpp
+++ b/src/blas/backends/cublas/cublas_handle.hpp
@@ -1,0 +1,56 @@
+/***************************************************************************
+*  Copyright (C) Codeplay Software Limited
+*  Licensed under the Apache License, Version 2.0 (the "License");
+*  you may not use this file except in compliance with the License.
+*  You may obtain a copy of the License at
+*
+*      http://www.apache.org/licenses/LICENSE-2.0
+*
+*  For your convenience, a copy of the License has been included in this
+*  repository.
+*
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*
+**************************************************************************/
+#ifndef CUBLAS_HANDLE_HIPSYCL_HPP
+#define CUBLAS_HANDLE_HIPSYCL_HPP
+#include<atomic>
+#include<unordered_map>
+
+namespace oneapi {
+namespace mkl {
+namespace blas {
+namespace cublas {
+
+template<typename T>
+struct cublas_handle {
+    using handle_container_t = std::unordered_map<T, std::atomic<cublasHandle_t> *>;
+    handle_container_t cublas_handle_mapper_{};
+    ~cublas_handle() noexcept(false){
+    for (auto &handle_pair : cublas_handle_mapper_) {
+        cublasStatus_t err;
+        if (handle_pair.second != nullptr) {
+            auto handle = handle_pair.second->exchange(nullptr);
+            if (handle != nullptr) {
+                CUBLAS_ERROR_FUNC(cublasDestroy, err, handle);
+                handle = nullptr;
+            }
+            delete handle_pair.second;
+            handle_pair.second = nullptr;
+        }
+    }
+    cublas_handle_mapper_.clear();
+}
+};
+
+
+} // namespace cublas
+} // namespace blas
+} // namespace mkl
+} // namespace oneapi
+
+#endif // CUBLAS_HANDLE_HIPSYCL_HPP
--- a/src/blas/backends/cublas/cublas_scope_handle_hipsycl.cpp
+++ b/src/blas/backends/cublas/cublas_scope_handle_hipsycl.cpp
@@ -1,27 +1,12 @@
 #include "cublas_scope_handle_hipsycl.hpp"
+#include "cublas_handle.hpp"
 
 namespace oneapi {
 namespace mkl {
 namespace blas {
 namespace cublas {
 
-cublas_handle::~cublas_handle() noexcept(false) {
-    for (auto &handle_pair : cublas_handle_mapper_) {
-        cublasStatus_t err;
-        if (handle_pair.second != nullptr) {
-            auto handle = handle_pair.second->exchange(nullptr);
-            if (handle != nullptr) {
-                CUBLAS_ERROR_FUNC(cublasDestroy, err, handle);
-                handle = nullptr;
-            }
-            delete handle_pair.second;
-            handle_pair.second = nullptr;
-        }
-    }
-    cublas_handle_mapper_.clear();
-}
-
-thread_local cublas_handle CublasScopedContextHandler::handle_helper = cublas_handle{};
+thread_local cublas_handle<int> CublasScopedContextHandler::handle_helper = cublas_handle<int>{};
 
 CublasScopedContextHandler::CublasScopedContextHandler(cl::sycl::queue queue,
                                                        cl::sycl::interop_handle &ih)
--- a/src/blas/backends/cublas/cublas_scope_handle_hipsycl.cpp
+++ b/src/blas/backends/cublas/cublas_scope_handle_hipsycl.cpp
@@ -1,0 +1,71 @@
+#include "cublas_scope_handle_hipsycl.hpp"
+
+namespace oneapi {
+namespace mkl {
+namespace blas {
+namespace cublas {
+
+cublas_handle::~cublas_handle() noexcept(false) {
+    for (auto &handle_pair : cublas_handle_mapper_) {
+        cublasStatus_t err;
+        if (handle_pair.second != nullptr) {
+            auto handle = handle_pair.second->exchange(nullptr);
+            if (handle != nullptr) {
+                CUBLAS_ERROR_FUNC(cublasDestroy, err, handle);
+                handle = nullptr;
+            }
+            delete handle_pair.second;
+            handle_pair.second = nullptr;
+        }
+    }
+    cublas_handle_mapper_.clear();
+}
+
+thread_local cublas_handle CublasScopedContextHandler::handle_helper = cublas_handle{};
+
+CublasScopedContextHandler::CublasScopedContextHandler(cl::sycl::queue queue, cl::sycl::interop_handle& ih): interop_h(ih){}
+
+cublasHandle_t CublasScopedContextHandler::get_handle(const cl::sycl::queue &queue){
+    cl::sycl::device device = queue.get_device();
+    int current_device = interop_h.get_native_device<cl::sycl::backend::cuda>();
+    CUstream streamId = get_stream(queue);
+    cublasStatus_t err;
+    auto it = handle_helper.cublas_handle_mapper_.find(current_device);
+    if (it != handle_helper.cublas_handle_mapper_.end()) {
+        if (it->second == nullptr) {
+            handle_helper.cublas_handle_mapper_.erase(it);
+        }
+        else {
+            auto handle = it->second->load();
+            if (handle != nullptr) {
+                cudaStream_t currentStreamId;
+                CUBLAS_ERROR_FUNC(cublasGetStream, err, handle, &currentStreamId);
+                if (currentStreamId != streamId) {
+                    CUBLAS_ERROR_FUNC(cublasSetStream, err, handle, streamId);
+                }
+                return handle;
+            }
+            else {
+                handle_helper.cublas_handle_mapper_.erase(it);
+            }
+        }
+    }
+    cublasHandle_t handle;
+
+    CUBLAS_ERROR_FUNC(cublasCreate, err, &handle);
+    CUBLAS_ERROR_FUNC(cublasSetStream, err, handle, streamId);
+
+    auto insert_iter = handle_helper.cublas_handle_mapper_.insert(
+        std::make_pair(current_device, new std::atomic<cublasHandle_t>(handle)));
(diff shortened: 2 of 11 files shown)


mmeterel:
What is the purpose of half.txt? That seems like a global search and replace of all "half" texts in the library.

sbalint98:
Half.txt is only necessary as a temporary solution until `half` is replaced by `cl::sycl::half` in the library. This is necessary since in hipSYCL half (according to the SYCL standard) is only defined in the `cl::sycl` namespace. It is only there so that the library can be built with hipSYCL and the tests can be run with hipSYCL.

EDIT:
The attached patch is mainly a global replacement indeed, but it needed some tweaking therefore I thought it might be useful for anyone who wants to verify this PR.

mmeterel:
> Half.txt is only necessary as a temporary solution until `half` is replaced by `cl::sycl::half` in the library. This is necessary since in hipSYCL half (according to the SYCL standard) is only defined in the `cl::sycl` namespace. It is only there so that the library can be built with hipSYCL and the tests can be run with hipSYCL.
> 
> EDIT:
> The attached patch is mainly a global replacement indeed, but it needed some tweaking therefore I thought it might be useful for anyone who wants to verify this PR.

So, we need to apply the half.txt patch to be able to test this PR? If yes, I would suggest first submitting the PR needed for updating the half namespace. Once it is reviewed and merged, I suggest submitting the following PRs. 
Regarding the half namespace: Instead of changing everything globally in oneMKL, is there any minimal workaround that can be applied to hipSYCL?

sbalint98:
> So, we need to apply the half.txt patch to be able to test this PR?

You should be able to test this PR with dpc++ without applying the `half.txt`, it should only be necessary for testing with hipSYCL.

>If yes, I would suggest first submitting the PR needed for updating the half namespace. 

That was the original purpose of the #105 I believe you mentioned in the conversation that you would like to change the namespace yourself. Do you have any information on when this change will approximately get merged?

> Regarding the half namespace: Instead of changing everything globally in oneMKL, is there any minimal workaround that can be applied to hipSYCL?

Pinging @illuhad for information on that

illuhad:
> Regarding the half namespace: Instead of changing everything globally in oneMKL, is there any minimal workaround that can be applied to hipSYCL?

Of couse you could just have something like
```
#ifdef __HIPSYCL__
using ::sycl::half;
#endif
```
in some commonly used oneMKL header to import `half` into the global namespace when compiling with hipSYCL. I suspect this is the cleanest solution here to move forward.

I'd prefer not to apply such a workaround to hipSYCL itself, as hipSYCL follows the SYCL standard by putting `half` into the `sycl` namespace, and I think that unnecessary pollution of the global namespace should be avoided if possible.

sbalint98:
Sorry for the long silence. @mmeterel would you prefer the solution proposed by @illuhad or would you rather have a different PR replacing the `half` with `sycl::half`

mmeterel:
> Sorry for the long silence. @mmeterel would you prefer the solution proposed by @illuhad or would you rather have a different PR replacing the `half` with `sycl::half`

Hi @sbalint98 we also would like to adopt SYCL2020 going forward, however, the current oneMKL version supported this repo is 2021.4 base tool kit, and i am worried sycl::half might not be supported in this version yet. 
I would suggest going with the solution proposed by @illuhad right now, and when ready, we can remove that workaround and switch to sycl::half everywhere. What do you think?

sbalint98:
I agree completely, I have added the change proposed by @illuhad to the `types.hpp` and to the `test_helper.hpp`. Now I can compile and run the tests with hipSYCL without applying any patch.

sbalint98:
Hi @mmeterel , I think this PR is ready, could you have a look?

mmeterel:
> Hi @mmeterel , I think this PR is ready, could you have a look?

Hi @sbalint98 thanks for the PR. I am currently doing some local tests on the PR. Will get back to you soon.

src/blas/backends/cublas/cublas_scope_handle_hipsycl.cpp
```diff
@@ -0,0 +1,72 @@
+#include "cublas_scope_handle_hipsycl.hpp"
```
mmeterel, line 19: Please add the appropriate Copyright banner

src/blas/backends/cublas/cublas_scope_handle_hipsycl.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace oneapi {
+namespace mkl {
+namespace blas {
+namespace cublas {
+
+cublas_handle::~cublas_handle() noexcept(false) {
```
mmeterel: This section is common with cublas_scope_handle.cpp. Maybe put them in a common header file?

src/blas/backends/cublas/cublas_scope_handle_hipsycl.hpp
```diff
@@ -0,0 +1,85 @@
...
+*  repository.
+*
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*
+**************************************************************************/
+#ifndef _MKL_BLAS_CUBLAS_SCOPED_HANDLE_HPP_
```
mmeterel: Please use more appropriate header guard macro. _CUBLAS_SCOPED_HANDLE_HIPSYCL_HPP_

src/blas/backends/cublas/cublas_scope_handle_hipsycl.hpp
```diff
@@ -0,0 +1,85 @@
...
+    inline T get_mem(U acc) {
+        return reinterpret_cast<T>(interop_h.get_native_mem<cl::sycl::backend::cuda>(acc));
+    }
+};
+
+} // namespace cublas
+} // namespace blas
+} // namespace mkl
+} // namespace oneapi
+#endif //_MKL_BLAS_CUBLAS_SCOPED_HANDLE_HPP_
```
mmeterel: same here

src/blas/backends/mklcpu/mklcpu_common.hpp
```diff
@@ -47,7 +47,9 @@ static inline auto host_task_internal(H &cgh, F f, int) -> decltype(cgh.run_on_h
 
 template <typename K, typename H, typename F>
 static inline void host_task_internal(H &cgh, F f, long) {
+#ifndef SYCL_DEVICE_ONLY
```
mmeterel: Can you explain the reason for this change pls?

tests/unit_tests/main_test.cpp
```diff
@@ -116,7 +116,11 @@ int main(int argc, char** argv) {
                         if (dev.is_gpu() && vendor_id == NVIDIA_ID)
                             continue;
 #endif
+#ifdef __HIPSYCL__
```
mmeterel, line 119: What is the reason for this change?

illuhad: I think this is about a difference in classification of devices between hipSYCL and DPC++. The SYCL specification does not really define what "accelerator" is and when `is_accelerator()` should return true. In particular, it does not say that `is_accelerator()` and `is_gpu()` should be mutually exclusive. In hipSYCL, for a GPU both `is_gpu() == true` as well as `is_accelerator() == true`. On the other hand, for DPC++ `is_accelerator() == false` for a GPU.
This change therefore is required to allow selecting a  GPU in hipSYCL.

If we want to exclude FPGAs, maybe we could generalize it to `if(dev.is_gpu() || !dev.is_accelerator())`?

src/blas/backends/mklcpu/mklcpu_common.hpp
sbalint98: As in hipSYCL `host_task` is not implemented we use `single_task` instead. Since `single_task` could be used as GPU kernel it is not allowed to call CPU functions such as the cblas functions. In order to prevent that from happening we exclude it from the compilation in case we are compiling for the device.

src/blas/backends/cublas/cublas_scope_handle_hipsycl.cpp
sbalint98: Changed the cublas_handle to a templated task, and moved it to a separate header `cublas_handle.hpp`

src/blas/backends/mklcpu/mklcpu_common.hpp
mmeterel: Any reason why host_task is not implemented in hipSYCL?

illuhad: @mmeterel host tasks are generally inefficient for backend interop, which is however one of their primary use cases. Since hipSYCL provides more efficient alternatives, implementing host task is curently low on the priority list.

src/blas/backends/cublas/cublas_handle.hpp
```diff
@@ -0,0 +1,56 @@
...
+*  repository.
+*
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*
+**************************************************************************/
+#ifndef CUBLAS_HANDLE_HIPSYCL_HPP
```
mmeterel: Please remove HIPSYCL from header guard. This is common header file.

src/blas/backends/mklcpu/mklcpu_common.hpp
mmeterel: I am not an expert in this area, but AFAIK, we use single_task to call CPU functions if run_on_host_intel is not supported by the compiler. So, this is confusing to me. Are you saying, in hipSYCL implementation, single_task is not allowed to call CPU functions? 

Also, did you mean to use __SYCL_DEVICE_ONLY__ macro? Or are you proposing to add a new macro here?

tests/unit_tests/main_test.cpp
mmeterel: @illuhad Thanks for the explanation. I agree with you. I also see same behavior in my local tests. I am ok with the current implementation.

mmeterel:
@sbalint98 Could you please rebase your branch with latest develop and test with latest llvm compiler? Your branch still has sycl::vector_class calls which are not supported in recent llvm compilers.

src/blas/backends/mklcpu/mklcpu_common.hpp
sbalint98: > single_task is not allowed to call CPU functions?

HipSYCL compiles every kernel for all available platforms. For example, in the case, CPU and Cuda are targetted all kernels will be compiled for both. It is illegal to call host functions from device code; therefore we need to differentiate between when generating device and CPU code, which can be done using the `SYCL_DEVICE_ONLY` macro, so that we only compile the `single_task` kernels in case we are compiling for CPU. Since these functions are only meant to be executed on the CPU, it shouldn't be a problem that the GPU version does not exist.

> Also, did you mean to use SYCL_DEVICE_ONLY macro? Or are you proposing to add a new macro here?

`SYCL_DEVICE_ONLY` is set by hipSYCL in case generating device code. I think here this is the right macro to use. See [here](https://github.com/illuhad/hipSYCL/blob/develop/doc/macros.md)

illuhad: > I am not an expert in this area, but AFAIK, we use single_task to call CPU functions if run_on_host_intel is not supported by the compiler. So, this is confusing to me. Are you saying, in hipSYCL implementation, single_task is not allowed to call CPU functions?

As @sbalint98 points out, since `single_task()` is designed to execute kernel code, it will in general be also compiled for non-CPU devices. This is also the case with DPC++ and to be expected with any SYCL implementation.

> Also, did you mean to use `__SYCL_DEVICE_ONLY__` macro? Or are you proposing to add a new macro here?


hipSYCL currently cannot guarantee the existence of `__SYCL_DEVICE_ONLY__` on all backends. There is `SYCL_DEVICE_ONLY` instead, but you are right to point out that the SYCL spec says that it should be called `__SYCL_DEVICE_ONLY__`. This is an issue in hipSYCL that we should look at fixing.

For the sake of completeness, I'd like to point out that in general `__SYCL_DEVICE_ONLY__` is not sufficient to control whether code executes only on device, even though it is frequently misunderstood in that way (and this approach happens to work in DPC++ and most hipSYCL backends). `__SYCL_DEVICE_ONLY__` is defined by the spec to be set when the current compilation does not produce a host binary. This means that this macro only work for this purpose when the SYCL implementation relies on multiple independent compilation passes. If the SYCL implementation has a unified frontend compilation pass for host and device with late separation of device code at the IR level, this macro no longer works. In that case `__SYCL_SINGLE_SOURCE__` should be defined.

mmeterel: @sbalint98  @illuhad Thanks for your patience with me :) and for your explanations. If I understand correctly, you will update this code to use `__SYCL_DEVICE_ONLY__`? Am I right?

illuhad: @mmeterel Yes, I think this should be the course of action. We have just merged support for `__SYCL_DEVICE_ONLY__` into hipSYCL for this purpose :-)

sbalint98: Now using `__SYCL_DEVICE_ONLY__` :)

mmeterel:
> @sbalint98 Could you please rebase your branch with latest develop and test with latest llvm compiler? Your branch still has sycl::vector_class calls which are not supported in recent llvm compilers.

@sbalint98 Please let me know when you complete this. I would like to do another round of tests after this.

src/blas/backends/mklcpu/mklcpu_common.hpp
mmeterel: Thanks @sbalint98 Did you test if this breaks anything with DPC++ compiler?

sbalint98: Yes I tested it with both dpcpp and hipSYCL and everything seemed to be fine

sbalint98:
It has been rebased to develop already, let me know if something doesn't seem right!

src/blas/backends/mklcpu/mklcpu_common.hpp
mmeterel: Thanks for confirming.

mmeterel:
> It has been rebased to develop already, let me know if something doesn't seem right!

Oh ok, looks like I missed the rebase commit. Thanks.

mmeterel (approved):
Thanks for the PR and all the updates.

mkrainiuk (approved):
Looks good to me, thank you!

PR #125, opened 2021-09-01
[cublas] Fix build for CUDA 11.x

npmiller:
# Description

In CUDA 11 the type of one of the argument of `cublasGemmEx` has been
changed, but CUDA does provide an overload with the previous argument
type for backwards compatibility.

However because it is used as a template argument the introduction of
that overload seemingly broke the template resolution.

This patch simply removes that function from the template arguments and
calls it directly in the intermediary templated function.

This still uses the old signature for `cublasGemmEx` so should work fine
for CUDA 10.x and relies on the compatibility overload for CUDA 11.x.

With this patch oneMKL builds successfully with CUDA 11.4.

# Checklist

## All Submissions

### DIFF
--- a/src/blas/backends/cublas/cublas_level3.cpp
+++ b/src/blas/backends/cublas/cublas_level3.cpp
@@ -69,13 +69,12 @@ GEMM_LAUNCHER(std::complex<double>, cublasZgemm)
 
 #undef GEMM_LAUNCHER
 
-template <typename Func, typename T_A, typename T_B, typename T_C, typename DATATYPE_A,
-          typename DATATYPE_B, typename DATATYPE_C>
-inline void gemm(Func func, DATATYPE_A DT_A, DATATYPE_B DT_B, DATATYPE_C DT_C,
-                 cl::sycl::queue &queue, transpose transa, transpose transb, int64_t m, int64_t n,
-                 int64_t k, T_C alpha, cl::sycl::buffer<T_A, 1> &a, int64_t lda,
-                 cl::sycl::buffer<T_B, 1> &b, int64_t ldb, T_C beta, cl::sycl::buffer<T_C, 1> &c,
-                 int64_t ldc) {
+template <typename T_A, typename T_B, typename T_C, typename DATATYPE_A, typename DATATYPE_B,
+          typename DATATYPE_C>
+inline void gemm_ex(DATATYPE_A DT_A, DATATYPE_B DT_B, DATATYPE_C DT_C, cl::sycl::queue &queue,
+                    transpose transa, transpose transb, int64_t m, int64_t n, int64_t k, T_C alpha,
+                    cl::sycl::buffer<T_A, 1> &a, int64_t lda, cl::sycl::buffer<T_B, 1> &b,
+                    int64_t ldb, T_C beta, cl::sycl::buffer<T_C, 1> &c, int64_t ldc) {
     using cuDataType_A = typename CudaEquivalentType<T_A>::Type;
     using cuDataType_B = typename CudaEquivalentType<T_B>::Type;
     using cuDataType_C = typename CudaEquivalentType<T_C>::Type;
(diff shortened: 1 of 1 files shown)


mmeterel:
@npmiller Thanks for the PR. A few questions for clarity:

1. Why do we need to switch to CUDA 11.x? 
2. Has this change (CUDA version change) propagated to README? I did not see it in #126 
3. Did you attach the log? I could not find it. 
4. @mkrainiuk, @jasukhar If we are changing the supported CUDA version, I think, we need to update our CI system as well.

npmiller:
> @npmiller Thanks for the PR. A few questions for clarity:
> 
>     1. Why do we need to switch to CUDA 11.x?
> 
This is not switching to CUDA 11.x, it is simply making it possible to build and run with CUDA 11, it still works with CUDA 10, and we can keep CUDA 10 as the default and supported version for now.

I do think we might need to switch at some point, but I think we should simply follow what DPC++ is using, which I believe is still 10.2 right now but might change soon.

>     2. Has this change (CUDA version change) propagated to README? I did not see it in [Add build instruction for CUDA backends #126](https://github.com/oneapi-src/oneMKL/pull/126)
> 

As I said earlier this isn't really a version change and 10.2 should still be the default supported version so I didn't change that in the README.

>     3. Did you attach the log? I could not find it.
> 
Apologies, I didn't attach them, here's two cuBLAS ctest log files, one with CUDA 11.4 and one with CUDA 10.2:

* [cublas-cuda-10.log](https://github.com/oneapi-src/oneMKL/files/7128642/cublas-cuda-10.log)
* [cublas-cuda-11.log](https://github.com/oneapi-src/oneMKL/files/7128646/cublas-cuda-11.log)

>     4. @mkrainiuk, @jasukhar If we are changing the supported CUDA version, I think, we need to update our CI system as well.

This shouldn't be needed yet, but it might be good to start thinking about it if we want to be able to update at the same time as DPC++.

mmeterel:
> > @npmiller Thanks for the PR. A few questions for clarity:
> > ```
> > 1. Why do we need to switch to CUDA 11.x?
> > ```
> 
> This is not switching to CUDA 11.x, it is simply making it possible to build and run with CUDA 11, it still works with CUDA 10, and we can keep CUDA 10 as the default and supported version for now.
> 
> I do think we might need to switch at some point, but I think we should simply follow what DPC++ is using, which I believe is still 10.2 right now but might change soon.

Thanks for your replies and clarification. I guess my question is, if we are not changing CUDA version, why are we trying to make oneMKL interfaces compatible with 11.x? I am not objecting the changes, just trying to understand what is driving this PR. :)


> 
> > ```
> > 2. Has this change (CUDA version change) propagated to README? I did not see it in [Add build instruction for CUDA backends #126](https://github.com/oneapi-src/oneMKL/pull/126)
> > ```
> 
> As I said earlier this isn't really a version change and 10.2 should still be the default supported version so I didn't change that in the README.

> > ```
> > 3. Did you attach the log? I could not find it.
> > ```
> 
> Apologies, I didn't attach them, here's two cuBLAS ctest log files, one with CUDA 11.4 and one with CUDA 10.2:
> 
> * [cublas-cuda-10.log](https://github.com/oneapi-src/oneMKL/files/7128642/cublas-cuda-10.log)
> * [cublas-cuda-11.log](https://github.com/oneapi-src/oneMKL/files/7128646/cublas-cuda-11.log)
> 
> > ```
> > 4. @mkrainiuk, @jasukhar If we are changing the supported CUDA version, I think, we need to update our CI system as well.
> > ```
> 
> This shouldn't be needed yet, but it might be good to start thinking about it if we want to be able to update at the same time as DPC++.

npmiller:
> Thanks for your replies and clarification. I guess my question is, if we are not changing CUDA version, why are we trying to make oneMKL interfaces compatible with 11.x? I am not objecting the changes, just trying to understand what is driving this PR. :)

Well, I originally tried to build oneMKL with CUDA 11.4 as that's the default version on my system, and when I ran into this issue I had a quick look and it seemed small enough that it might be worth just fixing, especially since the fix would easily just work for both CUDA 10 and CUDA 11.

If it had been more involved to get it to work on CUDA 11 then it probably wouldn't be worth doing until we actually switch, but as it stands this small patch makes oneMKL usable on CUDA 11 even if we don't support it officially, and will make the transition to the new version easier if and when we decide to do it.

mmeterel (approved):

mmeterel:
> > Thanks for your replies and clarification. I guess my question is, if we are not changing CUDA version, why are we trying to make oneMKL interfaces compatible with 11.x? I am not objecting the changes, just trying to understand what is driving this PR. :)
> 
> Well, I originally tried to build oneMKL with CUDA 11.4 as that's the default version on my system, and when I ran into this issue I had a quick look and it seemed small enough that it might be worth just fixing, especially since the fix would easily just work for both CUDA 10 and CUDA 11.
> 
> If it had been more involved to get it to work on CUDA 11 then it probably wouldn't be worth doing until we actually switch, but as it stands this small patch makes oneMKL usable on CUDA 11 even if we don't support it officially, and will make the transition to the new version easier if and when we decide to do it.

Agreed on your argument. Thanks for your work on the PR!

akabalov (approved):
Looks good to me, thanks!

Issue #91, opened 2021-05-14
Compilation error when building cuRAND or cuBLAS tests


sbalint98:
# Summary
When compiling oneMKL with both tests and cuRAND or cuBLAS enabled a compilation error occurs.

# Version
The current oneMKL develop head is used eg: 1ed12c7270a68b78c29178ccc582d0239a4bd050 

# Environment

* HW you use
Intel Gold 6130 CPU with Nvidia gtx1080 GPUs
* Backend library version
Cuda 10.0
MKL, and TBB obtained via intel installer version 2021.1.1
* OS name and version
Ubuntu 20.04
* Compiler version
dpc++ compiler cloned from develop with hash: 4e26734cb87c451e0562559d5d6f83b7eabcaea3
```
compiled with buildbot/configure.py --cuda
and buildbot/compile.py
```

* CMake 
[cmake_log_cublas.log](https://github.com/oneapi-src/oneMKL/files/6478343/cmake_log_cublas.log)
[cmake_log_curand.log](https://github.com/oneapi-src/oneMKL/files/6478345/cmake_log_curand.log)

# Steps to reproduce
```
git clone https://github.com/oneapi-src/oneMKL.git
mkdir build && cd build

LD_LIBRARY_PATH=/root/hipSYCL-main/dpc++-hand/llvm/build/install/lib/:$LD_LIBRARY_PATH \
CXX=/root/hipSYCL-main/dpc++-hand/llvm/build/install/bin/clang++ \
CC=/root/hipSYCL-main/dpc++-hand/llvm/build/install/bin/clang \
cmake -G Ninja \
-DCMAKE_BUILD_TYPE=Debug \
-DTBB_ROOT=/root/hipSYCL-main/dpc++/tbb/latest \
-DMKL_ROOT=/root/hipSYCL-main/dpc++/mkl/latest \
-DREF_BLAS_ROOT=/root/spack/opt/spack/linux-centos7-skylake_avx512/gcc-9.3.1/openblas-0.3.14-npb5lv7dhfygc3lgh6zx3x6chlyt4kth/ \
-DENABLE_CUBLAS_BACKEND=OFF \
-DENABLE_CURAND_BACKEND=ON \
-DENABLE_MKLGPU_BACKEND=OFF ..

LD_LIBRARY_PATH=/root/hipSYCL-main/dpc++-hand/llvm/build/install/lib/:$LD_LIBRARY_PATH ninja
```

# Observed behavior
[...]

PR #126, opened 2021-09-02
Add build instruction for CUDA backends

npmiller:
# Description

This adds CMake instructions to build oneMKL for CUBLAS and CURAND, the current instructions are a little confusing since they only mention MKL build so this is aiming to make them a little clearer when targeting other backends.

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,7 +504,7 @@ conan build ..
 
 2. Build and install all required [dependencies](#software-requirements). 
 
-### Building for MKL
+### Building for oneMKL
 
 - On Linux*
 ```bash
--- a/README.md
+++ b/README.md
@@ -561,9 +561,6 @@ With:
 -DENABLE_CURAND_BACKEND=True   \
 ```
 
-Note that enabling both the cuBLAS and cuRAND backends breaks building the
-tests.
-
 ### Build Options
 All options specified in the Conan section are available to CMake. You can specify these options using `-D<cmake_option>=<value>`.
 
--- a/README.md
+++ b/README.md
@@ -546,7 +546,6 @@ cmake .. -DENABLE_CUBLAS_BACKEND=True                      \
          -DENABLE_MKLCPU_BACKEND=False                     \   # disable Intel MKL CPU backend
          -DENABLE_MKLGPU_BACKEND=False                     \   # disable Intel MKL GPU backend
          [-DREF_BLAS_ROOT=<reference_blas_install_prefix>] \   # required only for testing
-         [-DREF_LAPACK_ROOT=<reference_lapack_install_prefix>] # required only for testing
 cmake --build .
 ctest
 cmake --install . --prefix <path_to_install_dir>
--- a/README.md
+++ b/README.md
@@ -504,7 +504,7 @@ conan build ..
 
 2. Build and install all required [dependencies](#software-requirements). 
 
-Then:
+### Building for MKL
 
 - On Linux*
 ```bash
@@ -531,6 +531,36 @@ ninja
 ctest
 cmake --install . --prefix <path_to_install_dir>
 ```
+
+### Building for CUDA
+
+- On Linux*
+
+With the CUBLAS backend:
+
+```bash
+# Inside <path to onemkl>
+mkdir build && cd build
+export CXX=<path_to_dpcpp_compiler>/bin/dpcpp;
+cmake .. -DENABLE_CUBLAS_BACKEND=ON   \
+         -DENABLE_MKLCPU_BACKEND=OFF  \ # disable Intel MKL CPU backend
+         -DENABLE_MKLGPU_BACKEND=OFF    # disable Intel MKL GPU backend
+cmake --build .
+cmake --install . --prefix <path_to_install_dir>
+```
+
+To build with the CURAND backend instead simply replace:
+```bash
+-DENABLE_CUBLAS_BACKEND=ON   \
+```
+
+With:
+```bash
+-DENABLE_CURAND_BACKEND=ON   \
+```
+
+Note that enabling both the CUBLAS and CURAND backends breaks building the
+tests.
 
 ### Build Options
 All options specified in the Conan section are available to CMake. You can specify these options using `-D<cmake_option>=<value>`.
--- a/README.md
+++ b/README.md
@@ -536,7 +536,7 @@ cmake --install . --prefix <path_to_install_dir>
 
 - On Linux*
 
-With the CUBLAS backend:
+With the cuBLAS backend:
 
 ```bash
 # Inside <path to onemkl>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -504,7 +504,7 @@ conan build ..
 
 2. Build and install all required [dependencies](#software-requirements). 
 
-Then:
+### Building for MKL
```
akabalov: Could you please change "MKL" to "oneMKL" since that is the correct name of library?

npmiller: Updated to "oneMKL"

README.md
```diff
@@ -532,6 +532,36 @@ ctest
 cmake --install . --prefix <path_to_install_dir>
 ```
 
+### Building for CUDA
+
+- On Linux*
+
+With the CUBLAS backend:
```
akabalov: Let's align names 
In this readme above we have cuBLAS (not CUBLAS) and cuRAND (not CURAND)

npmiller: Good point, updated with your suggestion

README.md
```diff
@@ -532,6 +532,36 @@ ctest
...
+
+```bash
+# Inside <path to onemkl>
+mkdir build && cd build
+export CXX=<path_to_dpcpp_compiler>/bin/dpcpp;
+cmake .. -DENABLE_CUBLAS_BACKEND=ON   \
```
akabalov: For cmake options we use "True" or "False" instead of ON/OFF
-DENABLE_MKLCPU_BACKEND=False
-DENABLE_MKLGPU_BACKEND=False
-DENABLE_CUBLAS_BACKEND=True

Let's update them as well

README.md
```diff
@@ -532,6 +532,36 @@ ctest
...
+
+With the cuBLAS backend:
+
+```bash
+# Inside <path to onemkl>
+mkdir build && cd build
+export CXX=<path_to_dpcpp_compiler>/bin/dpcpp;
+cmake .. -DENABLE_CUBLAS_BACKEND=ON   \
+         -DENABLE_MKLCPU_BACKEND=OFF  \ # disable Intel MKL CPU backend
+         -DENABLE_MKLGPU_BACKEND=OFF    # disable Intel MKL GPU backend
```
akabalov: I tried cmake command manually with -DENABLE_CUBLAS_BACKEND=True and we also need to specify -DREF_BLAS_ROOT=<path_to_cblas>

It can be [-DREF_BLAS_ROOT=<reference_blas_install_prefix>]

README.md
```diff
@@ -532,6 +532,36 @@ ctest
...
+cmake --build .
+cmake --install . --prefix <path_to_install_dir>
```
akabalov, line 551: you don't mention ctest here as we have in oneMKL part. I believe we want to add it here as well

npmiller: Added ctest after the build step

README.md
npmiller: Swapped ON/OFF to "True" and "False"

README.md
npmiller: Added the same reference parameters as for the oneMKL build instructions

README.md
```diff
@@ -532,6 +532,39 @@ ctest
...
+With the cuBLAS backend:
+
+```bash
+# Inside <path to onemkl>
+mkdir build && cd build
+export CXX=<path_to_dpcpp_compiler>/bin/dpcpp;
+cmake .. -DENABLE_CUBLAS_BACKEND=True                      \
+         -DENABLE_MKLCPU_BACKEND=False                     \   # disable Intel MKL CPU backend
+         -DENABLE_MKLGPU_BACKEND=False                     \   # disable Intel MKL GPU backend
+         [-DREF_BLAS_ROOT=<reference_blas_install_prefix>] \   # required only for testing
```
akabalov, line 548: Let's have one common style: 
let's specify all options in [], as we have above for oneMKL building

README.md
```diff
@@ -532,6 +532,39 @@ ctest
...
+         [-DREF_LAPACK_ROOT=<reference_lapack_install_prefix>] # required only for testing
```
akabalov: We don't mention anywhere above that we have lapack with cuda backend. 
And I believe we cannot run lapack with cuda backend now. Let's not mention lapack here since for cublas/curand we don't need lapack paths and we don't mention anywhere that we have support of culapack.

mmeterel: Agreed, i dont think this is needed here.

README.md
vmalia: @akabalov is this about the formatting using `[]`?
If so, there are some community guidelines that you can find online, about best practices for markdown documentation, as we have here in our `README.md`. All options specified inside `[]` are optional, but all options without them are mandatory for the command to work. `<>` specifies the intended value for a given option. If specifying multiple valid options, one can use `|` to separate them. Something like this: `-DBUILD_DOC=<TRUE|FALSE>`.
Hope that helps! :)

README.md
```diff
@@ -532,6 +532,39 @@ ctest
...
+-DENABLE_CUBLAS_BACKEND=True   \
+```
+
+With:
+```bash
+-DENABLE_CURAND_BACKEND=True   \
+```
+
+Note that enabling both the cuBLAS and cuRAND backends breaks building the
+tests.
```
vmalia: Is this a known issue? I think we should have it open as a bug/known issue instead of specifying it in the README.

README.md
npmiller: Removed lapack from this command

README.md
npmiller: I'm not sure what you mean by that, if it's about `ENABLE_MKLGPU_BACKEND` and `ENABLE_MKLCPU_BACKEND`, I specifically made them non-optional because these are enabled by default, so if you simply omit them when trying to build for cuBLAS it will still try to build them as well, and that will fail on a system that doesn't have oneMKL installed.

Obviously on a system with both cuBLAS and oneMKL we could enable them both but for just the CUDA section I feel like it's less confusing to do it this way.

README.md
npmiller: I couldn't find an issue about this, I can file one if you want but I'm not too sure if this something that's broken or just not supposed to work.

akabalov (approved):
Looks good to me now. Thanks!

README.md
vmalia: @npmiller I think these changes are correct, but let's wait for @akabalov to clarify if this was her original question or something else.

README.md
mmeterel: Could you check if you are setting TARGET_DOMAINS properly? I remember seeing similar issue here: #91

vmalia: We may have to do a little refactoring and improve this mechanism, but this text should not be in the README. It will be misleading.

npmiller: Wording removed, and I filed a ticket about this: https://github.com/oneapi-src/oneMKL/issues/127

vmalia: Thank you @npmiller

vmalia (approved):

PR #134, opened 2021-10-08
all: replace unsupported sycl::vector_class with std::vector

mkrainiuk:
# Description

As DPC++ compiler moves from SYCL1.2.1 spec to SYCL2020, it doesn't support `sycl::vector_class` anymore. Initially it was an alias for `std::vector`, so this PR updates all sources to use `std::vector` to fix build with the latest compiler.
Please note since `sycl::vector_class` was an alias, these changes are compatible with older compilers. 
USM API from oneMKL Spec will be also updated.

# Checklist

## All Submissions

### DIFF
--- a/docs/domains/blas/asum.rst
+++ b/docs/domains/blas/asum.rst
@@ -103,7 +103,7 @@ asum (USM Version)
                         const T *x,
                         std::int64_t incx,
                         T_res *result,
-                        const sycl::vector_class<sycl::event> &dependencies = {})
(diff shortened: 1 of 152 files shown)


ericlars:
looks good

ericlars (approved):

mmeterel (approved):
Thanks for the PR. Looks good to me.

aelizaro (approved):

PR #136, opened 2021-10-20
[cublas] Fix race condition in cublas handle deletion

npmiller:
This was causing segmentation faults at the end of the program when
running tests.

Cublas handles need to be created per thread and per context, therefore
they're currently being stored in a `static thread_local` map, mapping
contexts to cublas handles.

These handles are deleted in two places, the map destructor called when
the thread is deleted, or the context destructor, where a callback is
registered on the context to delete the cublas handle.

Since there is two places where a handle can be deleted and that these
two will likely run in separate threads, the handle was placed in an
C++ atomic to avoid race conditions, and allocated on the heap so that
it can stay alive when one of either the thread or the context is
deleted.

But there was two issues with this, the main one is that the callback
registered on the context was not given the pointer to the atomic
handle, rather the pointer to the slot in the `thread_local` map where
the atomic pointer was stored. Which meant that whenever the thread
would be deleted before the context, the pointer available to the
callback would be invalid as the map was deleted. The solution for that
is to simply pass directly the pointer to the atomic handle to the
callback.

The second issue is that both of these places were deleting the atomic
in all cases, which means one of the deletion was invalid. The solution
for this is to only do the deletion in the last one being called, that
is to say the one which will get a `nullptr` when atomically accessing
the handle pointer.

With these two patches the tests are running fine.

Here's a log file running the cublas tests:
[cublas.log](https://github.com/oneapi-src/oneMKL/files/7379386/cublas.log)

### DIFF
--- a/include/oneapi/mkl/types.hpp
+++ b/include/oneapi/mkl/types.hpp
@@ -108,10 +108,8 @@ enum class order : char {
 } //namespace mkl
 } //namespace oneapi
 
-// Workaround for supporting ::half for hipSYCL
+// Workaround for supporting ::half for hipSYCL and DPC++
 // TODO: This should be removed after the interface is SYCL2020 conformant
-#ifdef __HIPSYCL__
 using ::cl::sycl::half;
-#endif
 
 #endif //_ONEMKL_TYPES_HPP_
--- a/src/blas/backends/cublas/cublas_handle.hpp
+++ b/src/blas/backends/cublas/cublas_handle.hpp
@@ -26,27 +26,32 @@ namespace mkl {
 namespace blas {
 namespace cublas {
 
-template<typename T>
+template <typename T>
 struct cublas_handle {
     using handle_container_t = std::unordered_map<T, std::atomic<cublasHandle_t> *>;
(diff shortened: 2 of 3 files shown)


mehdi-goli (approved):

mmeterel:
@npmiller have you applied clang-format?

npmiller:
> @npmiller have you applied clang-format?

I thought I did but maybe it missed the specific repo configuration, I believe it should be fixed now.

npmiller:
I also had to rebase this since the hipSYCL changes moved some of the code changed in here, I've attached an updated log file:

* [cublas-formatted.log](https://github.com/oneapi-src/oneMKL/files/7546046/cublas-formatted.log)

And I also had to extend the hipSYCL workaround for `half` as the type has also been removed from the global namespace in DPC++ (see: https://github.com/intel/llvm/pull/4818)

mmeterel (approved):
@npmiller Thanks a lot for the PR and cleaning the random segfaults I was seeing. It looks good to me.

mkrainiuk (approved):
Looks good to me, thank you!

PR #137, opened 2021-10-20
[curand] Use host_task rather than codeplay_host_task

npmiller:
`codeplay_host_task` is deprecated in SYCL 2020, `host_task` should be
used instead, this is already what is used by the cublas backend.

Additionally support for `codeplay_host_task` was recently removed from
DPC++, https://github.com/intel/llvm/pull/4426 so this code wouldn't
build with recent DPC++.

Here's a log file running the curand tests with this patch: 
[curand.log](https://github.com/oneapi-src/oneMKL/files/7379419/curand.log)


aelizaro (approved):
Looks good to me, thank you @npmiller!

mehdi-goli (approved):

mkrainiuk (approved):

PR #140, opened 2021-11-18
Update CXX compilers

dnhsieh-intel:
# Description

This PR updates the usage of CXX compilers in [README](https://github.com/dnhsieh-intel/oneMKL/blob/compiler_select/README.md) and [CMakeLists.txt](https://github.com/dnhsieh-intel/oneMKL/blob/compiler_select/CMakeLists.txt).

When building with CMake, a CXX compiler is now selected according to the following precedence:
1. A compiler specified by users;
2. `clang++` compiler if cuBLAS or cuRAND backend is enabled;
3. Intel(R) oneAPI DPC++ Compiler `dpcpp` if mklgpu backend is enabled;
4. `dpcpp` if it can be found in the environment variable PATH;
5. `clang++` on Linux and `clang-cl` on Windows if `dpcpp` is not available.

Locally builds were successful using the latest release of dpcpp, oneMKL, and clang++ and clang-cl from Intel project for LLVM* technology.


CMakeLists.txt
```diff
@@ -91,6 +74,56 @@ if(ENABLE_MKLCPU_BACKEND
...
+    endif()
+  endif()
+endif()
+
+# Define required C compilers for Windows before project
+if(WIN32)
+  set(CMAKE_C_COMPILER "clang-cl")
+endif()
+
+project(oneMKL VERSION 0.1.0 LANGUAGES CXX)
```
vmalia: We made a 0.2 release tag in the past. Does it make sense to update this version to 0.2.0 as well?

CMakeLists.txt
```diff
@@ -91,6 +74,56 @@ if(ENABLE_MKLCPU_BACKEND
...
+  endif()
+else()
+  if(ENABLE_CUBLAS_BACKEND OR ENABLE_CURAND_BACKEND)
+    set(CMAKE_CXX_COMPILER "clang++")
+  elseif(ENABLE_MKLGPU_BACKEND)
+    set(CMAKE_CXX_COMPILER "dpcpp")
+  else()
+    find_program(DPCPP_PATH dpcpp)
+    if(DPCPP_PATH)
+      set(CMAKE_CXX_COMPILER "dpcpp")
```
vmalia, line 91: Can you please share the purpose of using `find_program` here?

CMakeLists.txt
```diff
@@ -91,6 +74,56 @@ if(ENABLE_MKLCPU_BACKEND
...
+project(oneMKL VERSION 0.1.0 LANGUAGES CXX)
+
+# Override default CXX compile/link lines for Windows after project
+if(WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-function -w")
+  foreach (flag_var
+           CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
+           CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
+    string(REPLACE "/MD" "" ${flag_var} "${${flag_var}}")
+  endforeach()
```
vmalia, line 134: Did you check if any of the Debug build types work?
We had to remove a similar "override" of `CMAKE_CXX_FLAGS*` from our binary product's examples, to allow Visual Studio Debug runtime to work with CMake's automatic addition of `/MD` and `/MDd` compiler flags.

CMakeLists.txt
```diff
@@ -91,6 +74,56 @@ if(ENABLE_MKLCPU_BACKEND
...
+  set(CMAKE_CXX_COMPILE_OBJECT "<CMAKE_CXX_COMPILER> -fsycl /nologo <DEFINES> <INCLUDES> /EHsc <FLAGS> /Fo<OBJECT> -c <SOURCE>")
+  set(CMAKE_CXX_CREATE_STATIC_LIBRARY "lib /nologo <OBJECTS> /out:<TARGET>")
+  set(CMAKE_CXX_LINK_EXECUTABLE "<CMAKE_CXX_COMPILER> -fsycl /nologo <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
```
vmalia, line 137: In the past, overriding of `CMAKE_CXX_*` options was required, because Intel oneAPI DPC++ Compiler did not have official support in CMake, but things have changed a little since CMake 3.20.1. Which CMake version did you use for your local testing?

README.md
```diff
@@ -515,10 +528,10 @@ conan build ..
 ```bash
 # Inside <path to onemkl>
 mkdir build && cd build
-export CXX=<path_to_dpcpp_compiler>/bin/dpcpp;
```
vmalia, line 518: In line 78 of new changes, we have `if(CMAKE_CXX_COMPILER)`. Since `CMAKE_CXX_COMPILER` variable is cached by CMake, their documentation recommends using CC, FC and CXX to define compilers in the environment instead of defining `CMAKE_CXX_COMPILER` manually (it is still a supported mechanism). Can we keep the recommended `export CXX` approach and verify that this method works?

README.md
```diff
@@ -527,11 +540,10 @@ cmake --install . --prefix <path_to_install_dir>
 ```bash
 # Inside <path to onemkl>
 md build && cd build
-cmake .. -G Ninja
-                  [-DMKL_ROOT=<mkl_install_prefix>] \                    # required only if environment variable MKLROOT is not set
-                  [-DREF_BLAS_ROOT=<reference_blas_install_prefix>] \    # required only for testing
-                  [-DREF_LAPACK_ROOT=<reference_lapack_install_prefix>]  # required only for testing
-
+cmake .. -G Ninja [-DCMAKE_CXX_COMPILER=<path_to_dpcpp_compiler>\bin\dpcpp] \  # required only if dpcpp is not found in environment variable PATH
```
vmalia, line 544: Following the Linux approach and CMake recommendations, can we add `set CXX=<path_to_dpcpp_compiler>/bin/dpcpp` before the `cmake` command?

README.md
```diff
@@ -546,11 +558,11 @@ With the cuBLAS backend:
 ```bash
 # Inside <path to onemkl>
 mkdir build && cd build
-export CXX=<path_to_dpcpp_compiler>/bin/dpcpp;
```
vmalia, line 549: Same comment as before for `CXX`.

vmalia (changes requested):

CMakeLists.txt
dnhsieh-intel: I will update it to 0.2.0.

CMakeLists.txt
dnhsieh-intel: I use `find_program` to check if `dpcpp` is available, so `dpcpp` has a higher precedence than open source compilers in this implementation.

CMakeLists.txt
dnhsieh-intel: No, I didn't check debug build. I will check that.

CMakeLists.txt
dnhsieh-intel: I used CMake 3.16.3 on Linux and 3.15.7 on Windows. I think it would be better not to rely on this feature because the minimum version we require is 3.13.

README.md
dnhsieh-intel: Because now we want to auto-select compilers, we will encounter difficulties using this approach.

First, `CMAKE_CXX_COMPILER` is cached after `project` in CMakeLists.txt, but CMake also detects CXX compiler automatically after `project`. If we rely on cached `CMAKE_CXX_COMPILER`, we cannot tell if it is supplied by users or CMake.

Next, it is possible to detect `CXX` as an environment variable before `project`, but we may have a situation that the user sets `CXX` variable for other builds, then this `CXX` will be picked up mistakenly. I think it's safer to set CXX compiler in a self-contained way if necessary.

mmeterel:
@dnhsieh-intel My guess is this PR will impact hipSYCL integration, so I will ask hipSYCL team to provide feedback as well.

@sbalint98 @illuhad Can you please provide feedback if you have any?

sbalint98:
Thank you for pinging me, sorry for the delay, I have for some reason missed this PR completely. I'll look into it right away.

CMakeLists.txt
```diff
@@ -91,6 +74,56 @@ if(ENABLE_MKLCPU_BACKEND
   list(APPEND DOMAINS_LIST "rng")
 endif()
 
+# Define required CXX compilers before project
+if(CMAKE_CXX_COMPILER)
```
sbalint98: In case compiling with hipSYCL, I believe this part is not necessary, since the selection of the proper compiler will be handled by `syclcc` and the hipSYCL cmake integration. Would it be possible to modify this line like this:

```suggestion
if(CMAKE_CXX_COMPILER OR NOT ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
```
Since `ONEMKL_SYCL_IMPLEMENTATION` is set to `dpc++` by default this would not change the behaviour that `dpc++` is selected by default

CMakeLists.txt
```diff
@@ -91,6 +74,56 @@ if(ENABLE_MKLCPU_BACKEND
...
+      else()
+        message(STATUS "dpcpp was not found in PATH, using clang++ instead")
+        set(CMAKE_CXX_COMPILER "clang++")
+      endif()
+    endif()
+  endif()
+endif()
+
+# Define required C compilers for Windows before project
+if(WIN32)
```
sbalint98: I believe here as well would be preferable to only set the compiler explicitly in case we are using `dpc++` in hipSYCL we make efforts internally to find the proper settings even in case targeting windows

```suggestion
if(WIN32 AND ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
```

CMakeLists.txt
```diff
@@ -91,6 +74,56 @@ if(ENABLE_MKLCPU_BACKEND
...
+  set(CMAKE_C_COMPILER "clang-cl")
+endif()
+
+project(oneMKL VERSION 0.1.0 LANGUAGES CXX)
+
+# Override default CXX compile/link lines for Windows after project
+if(WIN32)
```
sbalint98: I would make these settings also specific for `dpc++` based on the reasons above

```suggestion
if(WIN32 AND ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
```

CMakeLists.txt
```diff
@@ -91,6 +74,56 @@ if(ENABLE_MKLCPU_BACKEND
...
+           CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
+    string(REPLACE "/MD" "" ${flag_var} "${${flag_var}}")
+  endforeach()
+  set(CMAKE_CXX_COMPILE_OBJECT "<CMAKE_CXX_COMPILER> -fsycl /nologo <DEFINES> <INCLUDES> /EHsc <FLAGS> /Fo<OBJECT> -c <SOURCE>")
+  set(CMAKE_CXX_CREATE_STATIC_LIBRARY "lib /nologo <OBJECTS> /out:<TARGET>")
+  set(CMAKE_CXX_LINK_EXECUTABLE "<CMAKE_CXX_COMPILER> -fsycl /nologo <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
+endif()
+
+# Temporary disable sycl 2020 deprecations warnings
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DSYCL2020_DISABLE_DEPRECATION_WARNINGS")
```
sbalint98: This is also not necessary for hipSYCL, therefore I would also only add this flag when compiling with `dpc++`

```suggestion
if( ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DSYCL2020_DISABLE_DEPRECATION_WARNINGS")
endif()
```

README.md
```diff
@@ -366,6 +369,16 @@ Python | 3.6 or higher | No | *N/A* | *Pre-installed or Installed by user* | [PS
 
 ---
 
+## Selection of Compilers
+
+A compiler needs to be chosen according to the required backend of your application.
+
+- If your application requires Intel GPU, use [Intel(R) oneAPI DPC++ Compiler](https://software.intel.com/en-us/oneapi/dpc-compiler) `dpcpp`.
+- If your application requires NVIDIA GPU, use the latest release of `clang++` from [Intel project for LLVM* technology](https://github.com/intel/llvm/releases/tag/2021-07).
+- If neither Intel GPU nor NVIDIA GPU is required, you can use `dpcpp` or `clang++` on Linux and `dpcpp` or `clang-cl` from [Intel project for LLVM* technology](https://github.com/intel/llvm/releases/tag/2021-07) on Windows.
```
mkrainiuk: We also can use Intel oneAPI DPC++ compiler for this one, not only LLVM build.

CMakeLists.txt
```diff
@@ -91,6 +74,58 @@ if(ENABLE_MKLCPU_BACKEND
...
+    set(CMAKE_CXX_COMPILER "clang++")
+  elseif(ENABLE_MKLGPU_BACKEND)
+    set(CMAKE_CXX_COMPILER "dpcpp")
+  else()
+    find_program(DPCPP_PATH dpcpp)
+    if(DPCPP_PATH)
+      set(CMAKE_CXX_COMPILER "dpcpp")
+    else()
+      if(WIN32)
+        message(STATUS "dpcpp was not found in PATH, using clang-cl instead")
```
mkrainiuk: Should we use STATUS in case we find dpcpp as well?

CMakeLists.txt
```diff
@@ -91,6 +74,58 @@ if(ENABLE_MKLCPU_BACKEND
...
+        message(STATUS "dpcpp was not found in PATH, using clang++ instead")
+        set(CMAKE_CXX_COMPILER "clang++")
+      endif()
+    endif()
+  endif()
+endif()
+
+# Define required C compilers for Windows before project
+if(WIN32 AND ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
+  set(CMAKE_C_COMPILER "clang-cl")
```
mkrainiuk: `clang-cl` is not a part of Intel oneAPI DPC++ compiler environment starting from 2022.0, need to replace it with something from the package, like `icx` or `dpcpp-cl`

README.md
```diff
@@ -546,11 +558,11 @@ With the cuBLAS backend:
...
 mkdir build && cd build
-export CXX=<path_to_dpcpp_compiler>/bin/dpcpp;
-cmake .. -DENABLE_CUBLAS_BACKEND=True                      \
-         -DENABLE_MKLCPU_BACKEND=False                     \   # disable Intel MKL CPU backend
-         -DENABLE_MKLGPU_BACKEND=False                     \   # disable Intel MKL GPU backend
-         [-DREF_BLAS_ROOT=<reference_blas_install_prefix>] \   # required only for testing
+cmake .. -DENABLE_CUBLAS_BACKEND=True  \
+         -DENABLE_MKLCPU_BACKEND=False \                                  # disable Intel MKL CPU backend
+         -DENABLE_MKLGPU_BACKEND=False \                                  # disable Intel MKL GPU backend
+         [-DCMAKE_CXX_COMPILER=<path_to_clang++_compiler>/bin/clang++] \  # required only if clang++ is not found in environment variable PATH
```
mkrainiuk: Could you please move CMAKE_CXX option to the beginning of the command to align it with previous examples?

mmeterel:
@dnhsieh-intel I have two general questions:

1. What is the motivation/background for this PR? Is there a customer request?
2. How does CI impacted with this PR? Do we rely on the fact that CI always specifies which compiler needs to be used?

dnhsieh-intel:
@mmeterel 
1. As far as my understanding, the auto-selection of compilers is for the convenience of usage. @mkrainiuk Could you provide more background please?
2. With this PR, we will rely on the environment that is set by CI, i.e., `PATH`. PCm shows that the environment has been set correctly, so tests will pick either intel compilers or open source compilers correctly.

mkrainiuk:
> @mmeterel
> 
> 1. As far as my understanding, the auto-selection of compilers is for the convenience of usage. @mkrainiuk Could you provide more background please?

Right, user still can select a compiler manually, but we want to have a default compiler that will be selected based on what is available in the environment and what is a build targets.

mkrainiuk (approved):
Changes look good to me. Thank you!

CMakeLists.txt
dnhsieh-intel: @sbalint98 Thank you for the comments! Could you please review the changes and mark the conversations as resolved after verifying the fix?

sbalint98: I think I don't have the permission to mark the conversation resolved... at least I don't see the button for that.

The way cmake works with this change looks good to me, however, I couldn't fully test this PR because #149. Would it be possible to wait until that issue is resolved? If not I can create a local solution for that problem and test with that...

dnhsieh-intel: @sbalint98 Could you test this PR using a local solution please?

CMakeLists.txt
vmalia: thanks!

CMakeLists.txt
vmalia: Does CMake not find `dpcpp` from only line 90? CMake should do that search automatically and find `dpcpp` in the environment variables like `PATH`. Can you check this code section without find_program and if-check?

CMakeLists.txt
vmalia: Let's verify debug builds later, since we have not tested them on the main branch either.

CMakeLists.txt
vmalia: My point is that we saw some warnings and CMake policy errors when using 3.20.1 when it was released, so it is better to check with CMake 3.20.1 also, that overriding default CMake behavior is fine.

README.md
vmalia: got it, thanks!

CMakeLists.txt
```diff
@@ -91,6 +74,71 @@ if(ENABLE_MKLCPU_BACKEND
...
+        set(CMAKE_CXX_COMPILER "clang-cl")
+      else()
+        message(STATUS "dpcpp was not found in PATH, using clang++ instead")
+        set(CMAKE_CXX_COMPILER "clang++")
+      endif()
+    endif()
+  endif()
+endif()
+
+# Define required C compilers before project
```
sbalint98, line 104: I need to set `CMAKE_CXX_COMPILER` and `CMAKE_C_COMPILER` since dpc++ is not in my path. However, cmake complains that it cannot find clang, even though I have provided a correct path. Please correct me if I am wrong, but here the `CMAKE_C_COMPILER` set by the user will be overwritten, won't it? I believe a similar check might need to be added here as on line 78 to check if the user has set the `CMAKE_C_COMPILER` variable. 

This does not effect hipSYCL I just wanted to share my experience with this change.

CMakeLists.txt
sbalint98: Compilation with hipSYCL works as expected with this change!

dnhsieh-intel: Thank you! I will mark the conversations as resolved.

CMakeLists.txt
dnhsieh-intel: Thank you very much for catching this problem! We are working on this.

CMakeLists.txt
dnhsieh-intel: Line 90 is not sufficient. The `set` command only sets the variable literally. The search and check of compilers are triggered by `project`.

CMakeLists.txt
dnhsieh-intel: It works with CMake 3.20.1. I was wrong about the version on Windows. `setvars.bat` changes the version to 3.20.21032501-MSVC_2.

CMakeLists.txt
vmalia: okay thanks!

CMakeLists.txt
vmalia: okay thanks!

vmalia (approved):
Looks good to me. Thanks for the PR.

PR #151, opened 2021-12-20
Remove tautological constant compare warnings for isnan

andrewtbarker:
# Description

Remove thousands of warnings related to `isnan` that report `tautological constant compare` when using DPC++.

# Checklist

## All Submissions

[test.txt](https://github.com/oneapi-src/oneMKL/files/7748481/test.txt)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,10 +137,9 @@ if(WIN32 AND ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
   set(CMAKE_CXX_LINK_EXECUTABLE "<CMAKE_CXX_COMPILER> -fsycl /nologo <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
 endif()
 
+# Temporary disable sycl 2020 deprecations warnings
 if(ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
-  # Temporary disable sycl 2020 deprecations warnings
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DSYCL2020_DISABLE_DEPRECATION_WARNINGS")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-tautological-constant-compare")
 endif()
 
 # Target domains
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,9 +137,10 @@ if(WIN32 AND ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
   set(CMAKE_CXX_LINK_EXECUTABLE "<CMAKE_CXX_COMPILER> -fsycl /nologo <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
 endif()
 
-# Temporary disable sycl 2020 deprecations warnings
 if(ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
+  # Temporary disable sycl 2020 deprecations warnings
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DSYCL2020_DISABLE_DEPRECATION_WARNINGS")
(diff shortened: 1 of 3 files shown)


mmeterel:
@andrewtbarker  Is there a way to actually update the code to eliminate these warnings, rather than disabling them?

andrewtbarker:
@mmeterel I am not sure. Many of these warnings come from `iamin` in `reference_blas_templates.hpp` which seems to be included by many files. I suspect the `isnan` checks are actually required here to get the right reference behavior if your array has nans, and by default dpcpp will complain every time it sees an `isnan` because of its default floating point mode. 

I am not sure what the best solution is here but I don't think the current status with thousands of warnings in the CI is very good.

dnhsieh-intel:
Right now simply calling `std::isnan` generates the warning. I think the long term solution is to have Compiler team address this issue. As for short term solutions, I suggest either suppressing the warning specific to `std::isnan` (see below) or replacing `std::isnan` if you have an available implementation of `isnan` that won't trigger the warning. Fortunately, `std::isnan` only appears in `src/blas/backends/netlib/netlib_level1.cpp` and `tests/unit_tests/blas/include/reference_blas_templates.hpp`.
```cpp
#include <cmath>

int main()
{
   #pragma CLANG diagnostic push
   #pragma CLANG diagnostic ignored "-Wtautological-constant-compare"
   bool is_nan = std::isnan(0.0 / 0.0);
   #pragma CLANG diagnostic pop

   return 0;
}
```

mmeterel:
> Right now simply calling `std::isnan` generates the warning. I think the long term solution is to have Compiler team address this issue. As for short term solutions, I suggest either suppressing the warning specific to `std::isnan` (see below) or replacing `std::isnan` if you have an available implementation of `isnan` that won't trigger the warning. Fortunately, `std::isnan` only appears in `src/blas/backends/netlib/netlib_level1.cpp` and `tests/unit_tests/blas/include/reference_blas_templates.hpp`.
> 
> ```c++
> #include <cmath>
> 
> int main()
> {
>    #pragma CLANG diagnostic push
>    #pragma CLANG diagnostic ignored "-Wtautological-constant-compare"
>    bool is_nan = std::isnan(0.0 / 0.0);
>    #pragma CLANG diagnostic pop
> 
>    return 0;
> }
> ```

How about we suppress the warnings as in PR and open an issue to the compiler and link it here? So, the compiler team is aware and we can remove the suppressing when/if compiler fixes it (or recommends how to handle it)

andrewtbarker:
I've changed the PR to implement @dnhsieh-intel 's solution.

According to existing compiler tickets the non-standard `isnan` behavior and the warnings are expected (and hotly debated in the comments to the tickets), so I am afraid this temporary solution may be the best we can do.

Another possible solution is `#pragma float_control(precise, on)`, but when I try that I get the warning:
```
test.cpp:40:17: warning: '#pragma float_control' is not supported on this target - ignored [-Wignored-pragmas]
```

mmeterel:
> I've changed the PR to implement @dnhsieh-intel 's solution.
> 
> According to existing compiler tickets the non-standard `isnan` behavior and the warnings are expected (and hotly debated in the comments to the tickets), so I am afraid this temporary solution may be the best we can do.
> 
> Another possible solution is `#pragma float_control(precise, on)`, but when I try that I get the warning:
> 
> ```
> test.cpp:40:17: warning: '#pragma float_control' is not supported on this target - ignored [-Wignored-pragmas]
> ```

These tickets are internal and are not applicable to this open source project. I think, we need an issue filed against https://github.com/intel/llvm. If this is an expected behavior from compiler due to fp-model used, then, we should evaluate if using isnan is safe for our purpose. Testing with a vector with NaN could be an option. Once we know it is safe to use isnan, we can ignore the warnings.

dnhsieh-intel:
@andrewtbarker Thank you pointing to those tickets. `isnan(0.0 / 0.0)` returns false, because now `isnan` always returns false. This will lead to unexpected results. Suppressing warnings is no longer a good idea given this information.

andrewtbarker:
Since we don't think suppressing warnings is the right solution, I'm closing this PR but documenting the issue as #157.

PR #158, opened 2022-01-27
[BLAS] Fix syrk test bug

mmeterel:
This PR is fixing a test bug found during hipSYCL testing. Removed unsupported conjugate_transpose test conditions from SYRK. 

Also capped max error printing to 20 to avoid large log files when there are errors. 

- [X] Do all unit tests pass locally?

### DIFF
--- a/tests/unit_tests/blas/batch/syrk_batch_stride.cpp
+++ b/tests/unit_tests/blas/batch/syrk_batch_stride.cpp
@@ -63,14 +63,13 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t batch_size) {
 
     upper_lower = (oneapi::mkl::uplo)(std::rand() % 2);
     if ((std::is_same<fp, float>::value) || (std::is_same<fp, double>::value)) {
-        trans = (oneapi::mkl::transpose)(std::rand() % 2);
+        trans = (std::rand() % 2) == 0 ? oneapi::mkl::transpose::nontrans
+                                       : (std::rand() % 2) == 0 ? oneapi::mkl::transpose::trans
+                                                                : oneapi::mkl::transpose::conjtrans;
     }
     else {
-        tmp = std::rand() % 3;
-        if (tmp == 2)
-            trans = oneapi::mkl::transpose::conjtrans;
-        else
-            trans = (oneapi::mkl::transpose)tmp;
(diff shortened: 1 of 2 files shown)


vmalia:
Hi @mmeterel 
Seems like the count-based change will affect other domains too. Do we need to check with them to see if this is okay for their domains? What kind of error do you see that gives a long log file?

mmeterel:
> Hi @mmeterel Seems like the count-based change will affect other domains too. Do we need to check with them to see if this is okay for their domains? What kind of error do you see that gives a long log file?

@vmalia I did not see this file included in other domains. Where did you see it?
If there is a correctness error in gemm (e.g.) calculation, the current implementation will print all of the matrix indexes with wrong value - which can be quite long (more than 100K). Limiting to 20 will help with log size and run time. We need to fail the test and debug even if there is a single error anyways.

tests/unit_tests/blas/batch/syrk_batch_stride.cpp
```diff
@@ -63,14 +63,13 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t batch_size) {
 
     upper_lower = (oneapi::mkl::uplo)(std::rand() % 2);
     if ((std::is_same<fp, float>::value) || (std::is_same<fp, double>::value)) {
-        trans = (oneapi::mkl::transpose)(std::rand() % 2);
+        trans = (std::rand() % 2) == 0 ? oneapi::mkl::transpose::nontrans
```
andrewtbarker, line 66: I am confused by these changes. What exactly is the unsupported test condition (ie, what datatype/trans combination caused problems)?

The [docs](https://oneapi-src.github.io/oneMKL/domains/blas/syrk.html) say "Conjugation is never performed, even if trans = transpose::conjtrans" which is maybe strange but it's not the same as unsupported.

mmeterel: mkl::transpose::conjtrans caused issue for complex precisions. It is not supported for complex precisions, hence removed from possible test conditions.

andrewtbarker (approved):
Looks good to me.

mkrainiuk (approved):
Looks good to me, thank you!

PR #165, opened 2022-02-09
[RNG] Add scripts to add new backend

kcenia4010:
# Description

Added scripts for creating new backends for RNG. Also updated an instruction on how to use them and fix the generated files.

### DIFF
--- a/scripts/rng/seed_type.txt
+++ b/scripts/rng/seed_type.txt
@@ -1,3 +1,0 @@
-uint64_t
-uint32_t
-uint64_t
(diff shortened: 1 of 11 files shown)


scripts/rng/seed_type.txt
```diff
@@ -0,0 +1,3 @@
+uint64_t
```
aelizaro: should it be done by some kind of json file to match engines and their properties (such as seed types (some has several ones))?

kcenia4010: Done

aelizaro:
Closing this PR as outdated one

PR #174, opened 2022-03-05
improve/fix references to the documentation within the readme

fitchbe:
Signed-off-by: Benjamin Fitch <benjamin.fitch@intel.com>

Based on feedback, improved the references from the readme to the documentation and fixed a couple of links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ $> clang++ -fsycl –I$ONEMKL/include app.cpp
 $> clang++ -fsycl app.o –L$ONEMKL/lib –lonemkl_blas_mklcpu –lonemkl_blas_cublas
 ```
 
-*Refer to [Selection of Compilers](#selection-of-compilers) for the choice between `dpcpp` and `clang++` compilers.*
+*Refer to [Selecting a Compiler](https://oneapi-src.github.io/oneMKL/selecting_a_compiler.html) for the choice between `dpcpp` and `clang++` compilers.*
 
 ### Supported Configurations:
 
@@ -377,9 +377,9 @@ Microsoft Windows* Server | 2016, 2019 | *Not supported*
     </tbody>
 </table>
 
-*If [Building with Conan](#building-with-conan), above packages marked as "No" must be installed manually.*
+*If [Building with Conan](https://oneapi-src.github.io/oneMKL/building_the_project.html#building-with-conan), above packages marked as "No" must be installed manually.*
 
-*If [Building with CMake](#building-with-cmake), above packages must be installed manually.*
+*If [Building with CMake](https://oneapi-src.github.io/oneMKL/building_the_project.html#building-with-cmake), above packages must be installed manually.*
 
 #### Notice for Use of Conan Package Manager
 **LEGAL NOTICE: By downloading and using this container or script as applicable (the "Software Package") and the included software or software made available for download, you agree to the terms and conditions of the software license agreements for the Software Package, which may also include notices, disclaimers, or license terms for third party software (together, the "Agreements") included in this README file.**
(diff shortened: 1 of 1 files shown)


mmeterel (approved):
Looks good.

README.md
```diff
@@ -116,7 +116,7 @@ $> clang++ -fsycl –I$ONEMKL/include app.cpp
 $> clang++ -fsycl app.o –L$ONEMKL/lib –lonemkl_blas_mklcpu –lonemkl_blas_cublas
 ```
 
-*Refer to [Selection of Compilers](#selection-of-compilers) for the choice between `dpcpp` and `clang++` compilers.*
+*Refer to [Selecting a Compiler](https://oneapi-src.github.io/oneMKL/selecting_a_compiler.html) for the choice between `dpcpp` and `clang++` compilers.*
```
vmalia, line 119: Can you please share the reason for using links instead of Markdown references?
From what I understand, if you are browsing a non-default branch, the new approach will take me to the develop/main branch, instead of the non-default branch's own `README.md`. Is this the intended behavior we want?
Edit: As an example, you can check how Conan C++ Package manager handles RST documentation with custom domain hosting of those docs,
https://github.com/conan-io/docs/tree/master
https://docs.conan.io/en/latest/

A similar approach is followed by Python and many other open-source projects. Let me know if you need more examples.

mmeterel: @vmalia  Good point. I think we should stay in the non-default branch.

vmalia: @mmeterel that is my thought as well. Let's wait for the rationale for this change from the documentation team.

fitchbe: The links that you're referring to used to go to Markdown content; now those targets are in the reStructuredText content which gets published on GitHub Pages. Therefore, this was not an (intentional) change in approach on my part. I assumed that we'd want to take the reader to the generated documentation. If there is a better way, I will be quite happy to use a reference from the readme to the appropriate reStructuredText file, but the reader won't have the context of the full documentation. Perhaps you could give an example of the syntax that you want. Thanks!

mmeterel: I guess, the question is, if the user makes any change to these files/links in their PR, will they have any way to test/verify it in their own branch? I am not fully understanding "publishing on GitHub Pages", but it seems to me that, these changes will not be visible/working until we "publish them on Github Pages". I might be missing something though.

fitchbe: Even we ourselves couldn't test such changes until we published to GitHub Pages, so you're correct. We merged the PR (along with the gh-pages PR) and then tested the links. The choice is between external links that take a regular user (versus a contributor) to the current generated documentation, or internal links that are more convenient for a non-Intel contributor. I'm not sure that there's a "right" answer, but I'm happy to make the switch.

vmalia: I see, so you intended to connect the new RST-built-and-hosted documentation to the `README.md` via links.
When we created the basic docs, we created markdown documentation that remains in GitHub as markdown, and also RST docs that can be built and hosted: https://github.com/oneapi-src/oneMKL/tree/develop/docs

Are we hosting the same RST docs here or different ones? https://oneapi-src.github.io/oneMKL/
Edit: I see we are using the same ones. In that case, do you think we could provide a link at the top of the main `README.md` once pointing to https://oneapi-src.github.io/oneMKL ? All other links can still be markdown references.

vmalia: The use-cases would look like this:

1. For an end-user: Go to https://oneapi-src.github.io/oneMKL and read the latest info for the latest release.
2. For a contributor: Go to Github and follow markdown docs for the latest "development" documentation.

fitchbe: Yes, there is (already) a link to the full documentation earlier in the readme (in fact, a few links to it under Documentation). I could change the later references to be Markdown references. I might ask the InfoDev members who maintain the Intel open-source guidelines what they think about such references. Perhaps then I could submit a fresh pull request to switch those two or three links over. Does that sound good?

vmalia: Yeah, that sounds good to me.

vmalia: If you find that our project is using an unusual approach as compared to other Intel open-source projects, please point that out. Alignment in documentation is important for good customer experience.

fitchbe: Great! If you would approve this PR, I'll merge it and then make myself an AR to research the links.

vmalia: Sure, @fitchbe. Does it make sense to also create a new Github issue to reflect that? We do not want to miss those updates at a later time. We would prefer to keep markdown references as is if possible.

fitchbe: Our preference in InfoDev is Jira (Documentation project, for tracking purposes), but either way would be fine.

vmalia: Okay, please send me the JIRA offline.

vmalia (approved):
Conditionally approved - expecting a JIRA offline to track fixes for external links in `README.md`

Issue #175, opened 2022-03-08
cuBLAS tests failed after aligning with SYCL 2020 specification


dnhsieh-intel:
# Summary
After cuBLAS backend was aligned with SYCL 2020 specification, a large number of cuBLAS tests failed.

# Environment
* OS: Linux
* Hardware: NVIDIA TITAN RTX
* Backend library version: CUDA 10.2
* Compiler version: sycl-nightly 20220210

# Steps to reproduce
Clone the branch: https://github.com/dnhsieh-intel/oneMKL/tree/SYCL_2020_cuBLAS
```
$ mkdir build && cd build
$ cmake .. -DENABLE_CUBLAS_BACKEND=True -DENABLE_MKLCPU_BACKEND=False -DENABLE_MKLGPU_BACKEND=False -DREF_BLAS_ROOT=<reference_blas_install_prefix> -DTARGET_DOMAINS=blas
$ cmake --build .
$ ctest --output-on-failure
```

# Observed behavior
Some cuBLAS tests using dynamic libraries reported segmentation faults in column major cases.
Output of cuBLAS unit tests: [cublas_unit_tests.txt](https://github.com/oneapi-src/oneMKL/files/8202605/cublas_unit_tests.txt) (81% tests passed, 321 tests failed out of 1668)

Examples of failed tests:
```
          Start    1: BLAS/RT/Nrm2TestSuite/Nrm2Tests.RealSinglePrecision/Column_Major_TITAN_RTX
   1/1668 Test    #1: BLAS/RT/Nrm2TestSuite/Nrm2Tests.RealSinglePrecision/Column_Major_TITAN_RTX ..................................***Exception: SegFault  1.42 sec
```
```
          Start  155: BLAS/RT/CopyUsmTestSuite/CopyUsmTests.RealSinglePrecision/Column_Major_TITAN_RTX
 155/1668 Test  #155: BLAS/RT/CopyUsmTestSuite/CopyUsmTests.RealSinglePrecision/Column_Major_TITAN_RTX ............................***Exception: SegFault  0.58 sec
Run this program with --terse_output to change the way it prints its output.
Note: Google Test filter = CopyUsmTestSuite/CopyUsmTests.RealSinglePrecision/Column_Major_TITAN_RTX
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from CopyUsmTestSuite/CopyUsmTests
[...]

PR #189, opened 2022-04-25
Adding rocBLAS support for BLAS domain through Intel DPCPP compiler with HIP backend support

TejaX-Alaghari:
# Description

This PR is intended to extend the HIP backend of BLAS domain, which is currently supported for hipSYCL compiler to Intel’s DPC++/SYCL compiler.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? [hipSYCL.log](https://github.com/oneapi-src/oneMKL/files/9651382/hipSYCL.log) [llvm.log](https://github.com/oneapi-src/oneMKL/files/9651384/llvm.log)

## New features


ajaypanyala:
@TejaX-Alaghari Thank you for this PR! I have built and successfully tested it on an application containing SYCL code. Just had to adjust the following line in [FindrocBLAS.cmake](https://github.com/TejaX-Alaghari/oneMKL/blob/rocblas_hip_support/cmake/FindrocBLAS.cmake#L61) to ```IMPORTED_LOCATION "${HIP_PATH}/../rocblas/lib/librocblas.so"``` in order to handle non-standard ROCM paths. Could you please incorporate this small fix before the PR gets merged? Also filed a related [issue](https://github.com/intel/llvm/issues/6091) on the intel llvm page.

TejaX-Alaghari:
Thanks for the suggestion @ajaypanyala. I've added a commit for that change. 
Do let me know if there are any other suggestions.

CMakeLists.txt
```diff
@@ -143,8 +144,8 @@ if(WIN32 AND ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
   set(CMAKE_CXX_LINK_EXECUTABLE "<CMAKE_CXX_COMPILER> -fsycl /nologo <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
 endif()
 
-# Temporary disable sycl 2020 deprecations warnings for cuBLAS and cuSOLVER
-if(ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++" AND (ENABLE_CUBLAS_BACKEND OR ENABLE_CUSOLVER_BACKEND))
+# Temporary disable sycl 2020 deprecations warnings for cuBLAS, rocBLAS and cuSOLVER
+if(ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++" AND (ENABLE_CUBLAS_BACKEND OR ENABLE_CUSOLVER_BACKEND OR ENABLE_ROCBLAS_BACKEND))
```
mmeterel: Could you explain which deprecation warnings are being disabled here? An example would be useful.

cmake/FindCompiler.cmake
```diff
@@ -57,3 +62,4 @@ if(is_dpcpp)
   endif()
 
 endif()
+
```
mmeterel: Nit: please remove extra line

cmake/FindrocBLAS.cmake
```diff
@@ -0,0 +1,67 @@
...
+set(CMAKE_MODULE_PATH "${HIP_PATH}/cmake" ${CMAKE_MODULE_PATH})
+list(APPEND CMAKE_PREFIX_PATH
+    "${HIP_PATH}/lib/cmake"
+    "${HIP_PATH}/../lib/cmake"
+)
+
+find_package(HIP QUIET)
+find_package(rocblas REQUIRED)
+get_filename_component(SYCL_BINARY_DIR ${CMAKE_CXX_COMPILER} DIRECTORY)
+# the OpenCL include file from hip is opencl 1.1 and it is not compatible with DPC++
```
mmeterel: I see the same comments for FindcuBLAS.cmake as well. Is it just copy/paste or is this really the case for HIP as well?

cmake/FindrocBLAS.cmake
```diff
@@ -0,0 +1,67 @@
...
+# the OpenCL include headers 1.2 onward is required. This is used to bypass NVIDIA OpenCL headers
```
mmeterel: same comment as before. Did you mean AMD instead of NVIDIA??

cmake/FindrocBLAS.cmake
```diff
@@ -0,0 +1,67 @@
...
+find_path(OPENCL_INCLUDE_DIR CL/cl.h OpenCL/cl.h 
+HINTS 
+${OPENCL_INCLUDE_DIR}
+${SYCL_BINARY_DIR}/../include/sycl/
+)
+# this is work around to avoid duplication half creation in both hip and SYCL
```
mmeterel: Is this a valid argument for hip as well?

src/blas/backends/rocblas/rocblas_handle.hpp
```diff
@@ -0,0 +1,62 @@
...
+*  repository.
+*
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*
+**************************************************************************/
+#ifndef RBLAS_HANDLE_HPP
```
mmeterel: Could you please match the name of the file? `_ROCBLAS_HANDLE_HPP_`

src/blas/backends/rocblas/rocblas_handle.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include <unordered_map>
+#include "rocblas_helper.hpp"
+
+namespace oneapi {
+namespace mkl {
+namespace blas {
+namespace rocblas {
+
+template <typename T>
+struct rblas_handle {
```
mmeterel: would prefer if the struct name matches the file name: `rocblas_handle`

src/blas/backends/rocblas/rocblas_handle.hpp
```diff
@@ -0,0 +1,62 @@
...
+    using handle_container_t = std::unordered_map<T, std::atomic<rocblas_handle> *>;
+    handle_container_t rblas_handle_mapper_{};
```
mmeterel: same here: `rocblas_handle_mapper_`

src/blas/backends/rocblas/rocblas_handle.hpp
```diff
@@ -0,0 +1,62 @@
...
+        rblas_handle_mapper_.clear();
+    }
+};
+
+} // namespace rocblas
+} // namespace blas
+} // namespace mkl
+} // namespace oneapi
+
+#endif // RBLAS_HANDLE_HPP
```
mmeterel: match file name

src/blas/backends/rocblas/rocblas_level1.cpp
```diff
@@ -971,6 +971,9 @@ cl::sycl::event sdsdot(cl::sycl::queue &queue, int64_t n, float sb, const float
         });
     });
     done.wait();
+#if(defined(ENABLE_ROCBLAS_BACKEND) && !defined(__HIPSYCL__))
+    queue.wait();
```
mmeterel: Why did you need to add this? Wouldn't done.wait() be enough?

src/blas/backends/rocblas/rocblas_scope_handle.cpp
```diff
@@ -0,0 +1,154 @@
...
+    }
+    else {
+        // if the handle is nullptr it means the handle was already destroyed by
+        // the rocblas_handle destructor and we're free to delete the atomic
+        // object.
+        delete ptr;
+    }
+}
+
+rocblas_handle RocblasScopedContextHandler::get_handle(const cl::sycl::queue &queue) {
```
mmeterel: Please use sycl:: namespace instead of cl::sycl everywhere.

src/blas/backends/rocblas/rocblas_scope_handle.hpp
```diff
@@ -0,0 +1,93 @@
...
+*  repository.
+*
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*
+**************************************************************************/
+#ifndef ROCBLAS_SCOPED_HANDLE_HIPSYCL_HPP
```
mmeterel: The header guard looks wrong. `_ROCBLAS_SCOPE_HANDLE_HPP_`

src/blas/backends/rocblas/rocblas_scope_handle.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+template <typename T>
+struct rocblas_handle_container {
+    using handle_container_t = std::unordered_map<T, std::atomic<rocblas_handle> *>;
+    handle_container_t rocblas_handle_container_mapper_{};
+    ~rocblas_handle_container() noexcept(false);
+};
+
+/**
+* @brief NVIDIA advise for handle creation:
```
mmeterel: Are these comments still valid for rocBLAS backend?

mmeterel (commented):
Thanks for the PR. Looking forward to add this feature. 

I added some initial feedback. General comments:
1. Please check if the comments added for cuBLAS backend still makes sense here as well. And if yes, update them accordingly.
2. Please use sycl:: namespace throughout the project.
3. Please make sure header guards match header names.
4. Please test rocBLAS backend with hipSYCL so make sure nothing is broken. Update logs. 
5. I have some concerns about Copyrights - will add my feedback later ==> feedback added
6. Please update README and docs folder as needed. 
7. In the test logs, please test MKLCPU backend as well. Also, I see segfault errors. Can you fix them?

cmake/FindrocBLAS.cmake
```diff
@@ -0,0 +1,67 @@
+#==========================================================================
+#  Copyright (C) Codeplay Software Limited
```
mmeterel, line 3: For the new files added, could you please update Copyright notice as follows:


`#==========================================================================`
`#  Copyright 2020-2022 Intel Corporation`
`#  Copyright (C) Codeplay Software Limited`
rest of the copyright...

ajaypanyala:
@TejaX-Alaghari @mmeterel  I get a link error in the final step. 

```
[100%] Linking CXX shared library ../../../../lib/libonemkl_blas_rocblas.so
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: cannot find -lclang_rt.builtins-x86_64
```
Not sure if something needs to be fixed in [this](https://github.com/TejaX-Alaghari/oneMKL/blob/rocblas_hip_support/src/blas/backends/rocblas/CMakeLists.txt) file.

For now, I got it to work by manually adding `-L$ROCM_PATH/llvm/lib/clang/13.0.0/lib/linux` to the link line where `libclang_rt.builtins-x86_64.a` exists. The link step tries to find `clang_rt.builtins-x86_64` in the folder `$INTEL_LLVM_INSTALL_PATH/lib/clang/15.0.0/lib/linux` which doesn't exist. Only the include folder `$INTEL_LLVM_INSTALL_PATH/lib/clang/15.0.0/include` exists. I am building the `intel-llvm` compiler as described [here](https://github.com/intel/llvm/issues/6091)

**[EDIT]** This issue probably belongs in the [intel-llvm](https://github.com/intel/llvm) repo. There is a related [issue](https://github.com/intel/llvm/issues/6140) that was just opened.

cmake/FindrocBLAS.cmake
TejaX-Alaghari: Yes, [here](https://github.com/ROCm-Developer-Tools/hipamd/blob/develop/include/hip/amd_detail/hip_fp16_gcc.h#L250) is one of the ref

cmake/FindCompiler.cmake
TejaX-Alaghari: Sure, we'll address this in the next PR

cmake/FindrocBLAS.cmake
TejaX-Alaghari: Sure, we'll incorporate the new line into the copyright

cmake/FindrocBLAS.cmake
TejaX-Alaghari: Thanks for pointing this out, its indeed been included from FindcuBLAS.cmake to support older versions of ROCm (1.2+). But later, we decided to only focus on releases after 5.0 (with OCL 2.0). Hence this inclusion is not relevant now.

src/blas/backends/rocblas/rocblas_scope_handle.hpp
TejaX-Alaghari: Thanks for pointing out. We'll change this as per your suggestion in the next PR.

src/blas/backends/rocblas/rocblas_scope_handle.cpp
TejaX-Alaghari: Sure, we'll incorporate the change in the next PR

src/blas/backends/rocblas/rocblas_handle.hpp
TejaX-Alaghari: Sure, we'll incorporate the change in the next PR

src/blas/backends/rocblas/rocblas_level1.cpp
TejaX-Alaghari: The event object returned by the 'queue' doesn't handle hipEvents in the current LLVM compiler with HIP backend. We've worked around this issue with an internal developed patch but its not yet upstreamed. So, this workaround was implemented for the PR

src/blas/backends/rocblas/rocblas_scope_handle.hpp
TejaX-Alaghari: Yes, this is true ([ref](https://rocblas.readthedocs.io/en/rocm-4.3.0/handle_stream_device.html#stream-and-device-management)).

TejaX-Alaghari: The comments need to be adjusted accordingly for rocBLAS. We'll add relevant info to this.

src/blas/backends/rocblas/rocblas_handle.hpp
TejaX-Alaghari: We'll address this in the next PR

cmake/FindCompiler.cmake
mmeterel: @TejaX-Alaghari  This change is simple enough that, it does not require a separate PR. Could you please update the current PR?

cmake/FindrocBLAS.cmake
mmeterel: Please remove if not relevant.

src/blas/backends/rocblas/rocblas_handle.hpp
mmeterel: @TejaX-Alaghari  This change is simple enough that, it does not require a separate PR. Could you please update the current PR?

src/blas/backends/rocblas/rocblas_handle.hpp
mmeterel: @TejaX-Alaghari  This change is simple enough that, it does not require a separate PR. Could you please update the current PR?

src/blas/backends/rocblas/rocblas_level1.cpp
mmeterel: Thanks for the explanation. Do you know when will the LLVM compiler will be updated? Could you please paste the PR for LLVM here?

src/blas/backends/rocblas/rocblas_scope_handle.cpp
mmeterel: @TejaX-Alaghari  This change is simple enough that, it does not require a separate PR. Could you please update the current PR?

src/blas/backends/rocblas/rocblas_scope_handle.hpp
mmeterel: @TejaX-Alaghari  This change is simple enough that, it does not require a separate PR. Could you please update the current PR?

mmeterel:
> @TejaX-Alaghari @mmeterel I get a link error in the final step.
> 
> ```
> [100%] Linking CXX shared library ../../../../lib/libonemkl_blas_rocblas.so
> /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: cannot find -lclang_rt.builtins-x86_64
> ```
> 
> Not sure if something needs to be fixed in [this](https://github.com/TejaX-Alaghari/oneMKL/blob/rocblas_hip_support/src/blas/backends/rocblas/CMakeLists.txt) file.
> 
> For now, I got it to work by manually adding `-L$ROCM_PATH/llvm/lib/clang/13.0.0/lib/linux` to the link line where `libclang_rt.builtins-x86_64.a` exists. The link step tries to find `clang_rt.builtins-x86_64` in the folder `$INTEL_LLVM_INSTALL_PATH/lib/clang/15.0.0/lib/linux` which doesn't exist. Only the include folder `$INTEL_LLVM_INSTALL_PATH/lib/clang/15.0.0/include` exists. I am building the `intel-llvm` compiler as described [here](https://github.com/intel/llvm/issues/6091)
> 
> **[EDIT]** This issue probably belongs in the [intel-llvm](https://github.com/intel/llvm) repo. There is a related [issue](https://github.com/intel/llvm/issues/6140) that was just opened.

@ajaypanyala @TejaX-Alaghari Thanks for the detailed explanation. I have seen the same issue earlier. I am leaning towards waiting for LLVM issue to be fixed, so we don't have to recommend users to use temporary hack. What do you think?

CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ set(DOMAINS_LIST "")
 if(ENABLE_MKLCPU_BACKEND
         OR ENABLE_MKLGPU_BACKEND
         OR ENABLE_CUBLAS_BACKEND
+	OR ENABLE_ROCBLAS_BACKEND
```
vmalia: Can you please fix the alignment?

CMakeLists.txt
```diff
@@ -86,7 +87,7 @@ if(CMAKE_CXX_COMPILER OR NOT ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
     string(REPLACE "\\" "/" CMAKE_CXX_COMPILER ${CMAKE_CXX_COMPILER})
   endif()
 else()
-  if(ENABLE_CUBLAS_BACKEND OR ENABLE_CURAND_BACKEND)
+	if(ENABLE_CUBLAS_BACKEND OR ENABLE_CURAND_BACKEND OR ENABLE_ROCBLAS_BACKEND)
```
vmalia: We use 2 space alignments throughout the project. Can you please fix this one?

cmake/FindCompiler.cmake
```diff
@@ -38,8 +38,13 @@ if(is_dpcpp)
         -fsycl-targets=nvptx64-nvidia-cuda -fsycl-unnamed-lambda)
       list(APPEND UNIX_INTERFACE_LINK_OPTIONS
         -fsycl-targets=nvptx64-nvidia-cuda)
+    elif(ENABLE_ROCBLAS_BACKEND)
```
vmalia: I don't think `elif` is the right syntax for CMake. `elseif` is what you are looking for. https://cmake.org/cmake/help/v3.15/command/if.html

ajaypanyala:
> > @TejaX-Alaghari @mmeterel I get a link error in the final step.
> > ```
> > [100%] Linking CXX shared library ../../../../lib/libonemkl_blas_rocblas.so
> > /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: cannot find -lclang_rt.builtins-x86_64
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Not sure if something needs to be fixed in [this](https://github.com/TejaX-Alaghari/oneMKL/blob/rocblas_hip_support/src/blas/backends/rocblas/CMakeLists.txt) file.
> > For now, I got it to work by manually adding `-L$ROCM_PATH/llvm/lib/clang/13.0.0/lib/linux` to the link line where `libclang_rt.builtins-x86_64.a` exists. The link step tries to find `clang_rt.builtins-x86_64` in the folder `$INTEL_LLVM_INSTALL_PATH/lib/clang/15.0.0/lib/linux` which doesn't exist. Only the include folder `$INTEL_LLVM_INSTALL_PATH/lib/clang/15.0.0/include` exists. I am building the `intel-llvm` compiler as described [here](https://github.com/intel/llvm/issues/6091)
> > **[EDIT]** This issue probably belongs in the [intel-llvm](https://github.com/intel/llvm) repo. There is a related [issue](https://github.com/intel/llvm/issues/6140) that was just opened.
> 
> @ajaypanyala @TejaX-Alaghari Thanks for the detailed explanation. I have seen the same issue earlier. I am leaning towards waiting for LLVM issue to be fixed, so we don't have to recommend users to use temporary hack. What do you think?

@mmeterel Sounds good. That would be perfect! Thanks!

pengtu:
> > @TejaX-Alaghari @mmeterel I get a link error in the final step.
> > ```
> > [100%] Linking CXX shared library ../../../../lib/libonemkl_blas_rocblas.so
> > /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: cannot find -lclang_rt.builtins-x86_64
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Not sure if something needs to be fixed in [this](https://github.com/TejaX-Alaghari/oneMKL/blob/rocblas_hip_support/src/blas/backends/rocblas/CMakeLists.txt) file.
> > For now, I got it to work by manually adding `-L$ROCM_PATH/llvm/lib/clang/13.0.0/lib/linux` to the link line where `libclang_rt.builtins-x86_64.a` exists. The link step tries to find `clang_rt.builtins-x86_64` in the folder `$INTEL_LLVM_INSTALL_PATH/lib/clang/15.0.0/lib/linux` which doesn't exist. Only the include folder `$INTEL_LLVM_INSTALL_PATH/lib/clang/15.0.0/include` exists. I am building the `intel-llvm` compiler as described [here](https://github.com/intel/llvm/issues/6091)
> > **[EDIT]** This issue probably belongs in the [intel-llvm](https://github.com/intel/llvm) repo. There is a related [issue](https://github.com/intel/llvm/issues/6140) that was just opened.
> 
> @ajaypanyala @TejaX-Alaghari Thanks for the detailed explanation. I have seen the same issue earlier. I am leaning towards waiting for LLVM issue to be fixed, so we don't have to recommend users to use temporary hack. What do you think?

@mmeterel : According to the compiler team, SYCL compiler doesn't use clang-builtins library.  It uses libgcc_s.so instead. In this case, the dependency for libclang_rt is from ROCM blas library.  The clang version of libclang_rt should match the ROCM library built version.  It suggests that adding -L$ROCM_PATH/llvm/lib/clang/13.0.0/lib/linux is the right permanent solution.

mmeterel:
@TejaX-Alaghari 
Sorry for the delay in getting back to this. I have a few follow-up requests and questions about this PR. 
1. Could you please update each conversation with your feedback? 
2. I left 7 comments on May 11. Could you please respond to each of these comments with the status? Let's do our best to have every question/comment answered.

mmeterel:
@TejaX-Alaghari 
I tested this PR on an AMD system with ROCM version 5.0.0 and card version gfx90a. 
LLVM with HIP backend is built from this commit: 66361038b63caaae566fc9648f5da50b74222b83

I am getting the warning/errors/crashes as shown in the attached files. I have tested only Level1 APIs for simplicity. 

Could you please advise how to resolve them? Feel free to schedule a meeting to discuss in detail.

[amd_llvm_level1.txt](https://github.com/oneapi-src/oneMKL/files/9595182/amd_llvm_level1.txt)
[amd_llvm_level1_verbose.txt](https://github.com/oneapi-src/oneMKL/files/9595183/amd_llvm_level1_verbose.txt)

CMakeLists.txt
TejaX-Alaghari: Below is an example for the disabled warnings

`/home/ubuntu/PR/oneMKL/src/blas/backends/rocblas/rocblas_task.hpp:57:58: warning: 'hip' is deprecated: use 'ext_oneapi_hip' instead [-Wdeprecated-declarations]
    hipStream_t stream = sycl::get_native<sycl::backend::hip>(queue);
                                                         ^
/opt/compiler/llvm/build_3/include/sycl/CL/sycl/backend_types.hpp:34:7: note: 'hip' has been explicitly marked deprecated here
  hip __SYCL2020_DEPRECATED("use 'ext_oneapi_hip' instead") = ext_oneapi_hip,
      ^
/opt/compiler/llvm/build_3/include/sycl/CL/sycl/detail/defines_elementary.hpp:54:40: note: expanded from macro '__SYCL2020_DEPRECATED'
#define __SYCL2020_DEPRECATED(message) __SYCL_DEPRECATED(message)
                                       ^
/opt/compiler/llvm/build_3/include/sycl/CL/sycl/detail/defines_elementary.hpp:45:38: note: expanded from macro '__SYCL_DEPRECATED'
#define __SYCL_DEPRECATED(message) [[deprecated(message)]]`

TejaX-Alaghari:
> Thanks for the PR. Looking forward to add this feature.
> 
> I added some initial feedback. General comments:
> 
> 1. Please check if the comments added for cuBLAS backend still makes sense here as well. And if yes, update them accordingly.
> 2. Please use sycl:: namespace throughout the project.
> 3. Please make sure header guards match header names.
> 4. Please test rocBLAS backend with hipSYCL so make sure nothing is broken. Update logs.
> 5. I have some concerns about Copyrights - will add my feedback later ==> feedback added
> 6. Please update README and docs folder as needed.
> 7. In the test logs, please test MKLCPU backend as well. Also, I see segfault errors. Can you fix them?

Thanks for the detailed feedback on the PR. Please find below, the response for above comments:
1. The existing comments are relevant for the changes introduced in this PR
2. All the earlier "cl::sycl::"namespaces have been updated to "sycl::"
3. All header guards match their respective file names
4. TBD
5. Updated the copyrights per the suggestion for newly introduced files
6. Updated README and docs with the build configuration for rocBLAS backend through clang++
7. We haven't observed any issues while testing MKLCPU backend for BLAS with "ctest"

CMakeLists.txt
mmeterel: Why can't we use the recommended APIs instead of disabling the warnings?

TejaX-Alaghari: I agree with your suggestion. Mesut. The deprecated warning shown above could be fixed. The only concern is whether it breaks with hipSYCL config or not. 

I'll explore the other kinds of deprecated warnings being thrown as well and fix them wherever possible after validating on hipSYCL as well.

CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ option(ENABLE_CURAND_BACKEND "" OFF)
 option(ENABLE_ROCRAND_BACKEND "" OFF)
 option(ENABLE_NETLIB_BACKEND "" OFF)
 set(ONEMKL_SYCL_IMPLEMENTATION "dpc++" CACHE STRING "Name of the SYCL compiler")
+set(HIP_TARGETS "gfx90a" CACHE STRING "Target HIP architectures")
```
mmeterel: I don't think we should hard code the architecture. Are we saying, HIP backend only works with this architecture? I think, user should be able to set this during build time.

CMakeLists.txt
```diff
@@ -151,8 +153,8 @@ if(WIN32 AND ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
   set(CMAKE_CXX_CREATE_SHARED_LIBRARY "<CMAKE_CXX_COMPILER> -fsycl -fsycl-device-code-split=per_kernel /nologo <OBJECTS> ${MKL_SYCL_LIB} /link /out:<TARGET> /implib:<TARGET_IMPLIB> /pdb:<TARGET_PDB> /dll /version:<TARGET_VERSION_MAJOR>.<TARGET_VERSION_MINOR> <LINK_FLAGS> <LINK_LIBRARIES>")
 endif()
 
-# Temporary disable sycl 2020 deprecations warnings for cuBLAS and cuSOLVER
-if(ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++" AND (ENABLE_CUBLAS_BACKEND OR ENABLE_CUSOLVER_BACKEND))
+# Temporary disable sycl 2020 deprecations warnings for cuBLAS, rocBLAS and cuSOLVER
+if(ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++" AND (ENABLE_CUBLAS_BACKEND OR ENABLE_CUSOLVER_BACKEND OR ENABLE_ROCBLAS_BACKEND))
```
mmeterel: We should resolve the depreciations as much as possible, instead of disabling them. 
For cuBLAS, as far as I remember, we had functionality issue when we updated the deprecated APIs. That is why we had to keep old APIs and disable the warnings. We opened #175 to track that.

docs/building_the_project.rst
```diff
@@ -382,6 +382,30 @@ With:
...
+
+   # Inside <path to onemkl>
+   mkdir build && cd build
+   cmake .. [-DCMAKE_CXX_COMPILER=<path_to_clang++_compiler>/bin/clang++]  # required only if clang++ is not found in environment variable PATH
+            [-DCMAKE_C_COMPILER=<path_to_clang_compiler>/bin/clang]        # required only if clang is not found in environment variable PATH
+            -DENABLE_CUBLAS_BACKEND=False                                \
+            -DENABLE_MKLCPU_BACKEND=False                                \ # disable Intel MKL CPU backend
+            -DENABLE_MKLGPU_BACKEND=False                                \ # disable Intel MKL GPU backend
+            -DENABLE_ROCBLAS_BACKEND=True                                \
+            -DHIP_TARGETS=gfx90a                                         \ # Specify the targetted device architectures
```
mmeterel, line 403: Please specify a few examples for target device along with the public arch names (e.g, MI210) Also, this is giving the impression that, HIP_TARGET can be changed. But earlier, it was embedded to the build. Please correct me if I am wrong.

docs/building_the_project.rst
```diff
@@ -490,6 +514,12 @@ CMake.
   ``-DHIPSYCL_TARGETS`` according to the targeted hardware. For the options,
   see the tables in the hipSYCL-specific sections.
 
+
+.. note::
+  When building with clang++ for AMD backends, you must additionally provide
+  ``-DHIP_TARGETS`` according to the targeted hardware. For the options,
```
mmeterel: Does this HIP_TARGETS used in LLVM compiler? I could not find it after a quick search.

docs/building_the_project.rst
```diff
@@ -490,6 +514,12 @@ CMake.
...
+  see the tables in the hipSYCL-specific sections.
```
mmeterel: I would suggest adding your own list. Not relying on hipSYCL side.

mmeterel:
> 

Tested the PR again after latest commits. Still seeing major failures. Please reach out for a debug session.

mmeterel:
> > 
> 
> Tested the PR again after latest commits. Still seeing major failures. Please reach out for a debug session.

After discussing with @TejaX-Alaghari offline, this issue is most likely due to missing synchronization. Similar issue was fixed in #228 for cuBLAS backend. @TejaX-Alaghari will work on updating this PR. Please note that, similar updates will likely needs to be done for rocRAND (#218 ) and rocSOLVER (#208 ) PRs.

mmeterel:
Attached are the results for my local tests using LLVM compiler: 
[rocblas_llvm_log.txt](https://github.com/oneapi-src/oneMKL/files/9686943/rocblas_llvm_log.txt)

I see three problems with these results:
1. SCAL seg fault: Tracked in issue #233 
2. Warnings due to hipCtxGetCurrent: Tracked in issue #234 
3. Warnings due to linking module: Tracked in https://github.com/intel/llvm/issues/6922

I also ran this PR using hipSYCL. Attached are the results:
[rocblas_hipsycl_log.txt](https://github.com/oneapi-src/oneMKL/files/9686957/rocblas_hipsycl_log.txt)
The errors I see are intermittent and do not appear then we run gtest directly. I don't think, these are related to this PR. 

Based on the above discussion, I don't have concerns with merging this PR. The mentioned issues can be tracked separately. 

Thanks for all your efforts on enabling rocBLAS backend with LLVM-HIP compiler.

mmeterel (approved):

mmeterel:
@vmalia @mkrainiuk I just added my test results along with related issues. The PR is ok to merge IMHO. Could you please take a look from your side?

src/blas/backends/rocblas/rocblas_task.hpp
```diff
@@ -55,7 +55,7 @@ static inline void host_task_internal(H &cgh, sycl::queue queue, F f) {
 #else
 template <typename H, typename F>
 static inline void host_task_internal(H &cgh, sycl::queue queue, F f) {
-    cgh.interop_task([f, queue](sycl::interop_handler ih) {
+    cgh.host_task([f, queue](sycl::interop_handle ih) {
```
mkrainiuk, line 58: Could it be affected by changes in DPC++ compiler for dropping host support?

mkrainiuk (approved):
Thank you for the PR! It looks good to me with one minor question.

src/blas/backends/rocblas/rocblas_task.hpp
mmeterel: @mkrainiuk In my tests with latest LLVM, I did not see any deprecation warnings related to host_task(). I guess, if/when the compiler drops support, we can switch to recommended API.

PR #200, opened 2022-05-17
Update CMake for Windows DLL and rocRAND

mcao59:
# Description

This PR 
- defines compile/link line for windows DLL creation (fixes error on Windows GPU)
- updates cmake for new rocRAND backend
- removes windows GPU from RNG examples ctest because it's not supported

### More details on Windows GPU error:
SYCL exception is caught when running lapack example on Windows GPU:
```
Running LAPACK getrs example on GPU device.

Device name is: Intel(R) Graphics [0x9b41]
Running with single precision real data type:
Caught synchronous SYCL exception:
No kernel named _ZTSZN6oneapi3mkl6lapack8internal12slaswp_esimdERN2cl4sycl5queueERKSt6vectorINS4_5eventESaIS8_EEPS8_xPfxxxxPxxxE19slaswp_esimd_kernel was found -46 (CL_INVALID_KERNEL_NAME)
SYCL error code: 1
```
This error occurs because the shared library for `onemkl_lapack_mklgpu.dll` is not properly created with `mkl_sycl.lib`
####  Solution: set CMAKE_CXX_CREATE_SHARED_LIBRARY in top-level CMakeList.txt

Before this fix, CMake uses the following line to create onemkl_lapack_mklgpu.dll (paths to oneMKL libraries are omitted):

`link /nologo bin\lapack\backends\mklgpu\CMakeFiles\onemkl_lapack_mklgpu_obj.dir\mkl_lapack.cpp.obj bin\lapack\backends\mklgpu\CMakeFiles\onemkl_lapack_mklgpu_obj.dir\lapack_gpu_wrappers.cpp.obj  /out:bin\onemkl_lapack_mklgpu.dll /implib:lib\onemkl_lapack_mklgpu.lib /pdb:bin\onemkl_lapack_mklgpu.pdb /dll /version:0.0 /machine:x64 /INCREMENTAL:NO mkl_sycl.lib mkl_intel_ilp64.lib mkl_tbb_thread.lib mkl_core.lib tbb.lib OpenCL.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib`

After this fix, CMake uses the following line to create onemkl_lapack_mklgpu.dll

`dpcpp-cl -fsycl /nologo bin\lapack\backends\mklgpu\CMakeFiles\onemkl_lapack_mklgpu_obj.dir\mkl_lapack.cpp.obj bin\lapack\backends\mklgpu\CMakeFiles\onemkl_lapack_mklgpu_obj.dir\lapack_gpu_wrappers.cpp.obj   mkl_sycl.lib   /link /out:bin\onemkl_lapack_mklgpu.dll /implib:lib\onemkl_lapack_mklgpu.lib /pdb:bin\onemkl_lapack_mklgpu.pdb /dll /version:0.0 /machine:x64 /INCREMENTAL:NO mkl_sycl.lib mkl_intel_ilp64.lib mkl_tbb_thread.lib mkl_core.lib tbb.lib OpenCL.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib`

# Checklist

## All Submissions

[win_examples_ctest_results.txt](https://github.com/oneapi-src/oneMKL/files/8711976/win_examples_ctest_results.txt)

No because this PR only changed CMake files

## Bug fixes
      GitHub issue or in this PR)?


mkrainiuk (approved):
Looks good to me, thank you!

examples/rng/run_time_dispatching/CMakeLists.txt
```diff
@@ -29,12 +29,16 @@ set(DEVICE_FILTERS "")
 if(ENABLE_MKLCPU_BACKEND)
   list(APPEND DEVICE_FILTERS "cpu")
 endif()
-if(ENABLE_MKLGPU_BACKEND)
+# RNG only supports CPU backend on Windows
```
mmeterel: CPU => MKLCPU

examples/rng/run_time_dispatching/CMakeLists.txt
```diff
@@ -29,12 +29,16 @@ set(DEVICE_FILTERS "")
...
+if(UNIX AND ENABLE_MKLGPU_BACKEND)
   list(APPEND DEVICE_FILTERS "gpu")
 endif()
 if(ENABLE_CURAND_BACKEND)
```
mmeterel: Why don't we have UNIX condition for this (and the next if cases?

CMakeLists.txt
```diff
@@ -145,6 +145,10 @@ if(WIN32 AND ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
   set(CMAKE_CXX_COMPILE_OBJECT "<CMAKE_CXX_COMPILER> -fsycl /nologo <DEFINES> <INCLUDES> /EHsc <FLAGS> /Fo<OBJECT> -c <SOURCE>")
   set(CMAKE_CXX_CREATE_STATIC_LIBRARY "lib /nologo <OBJECTS> /out:<TARGET>")
   set(CMAKE_CXX_LINK_EXECUTABLE "<CMAKE_CXX_COMPILER> -fsycl /nologo <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
+  find_library(MKL_SYCL_LIB NAMES mkl_sycl
+    HINTS $ENV{MKLROOT} ${MKL_ROOT}
+    PATH_SUFFIXES lib/intel64)
+  set(CMAKE_CXX_CREATE_SHARED_LIBRARY "<CMAKE_CXX_COMPILER> -fsycl /nologo <OBJECTS> ${MKL_SYCL_LIB} /link /out:<TARGET> /implib:<TARGET_IMPLIB> /pdb:<TARGET_PDB> /dll /version:<TARGET_VERSION_MAJOR>.<TARGET_VERSION_MINOR> <LINK_FLAGS> <LINK_LIBRARIES>")
```
mmeterel, line 151: can you fix the indentation?

mmeterel: Curious, why do we need this special flag for lapack example and not for unit tests? (assuming unit tests were working fine?)

examples/rng/run_time_dispatching/CMakeLists.txt
mcao59: Because we don't support cuda and hip backends on windows. So if ENABLE_CURAND_BACKEND or ENABLE_ROCRAND_BACKEND is on, it has to be Linux

mmeterel: Right, but how does this Cmake file know that? If I build examples on Windows and enable rocRAND backend, it would still try to add hip::gpu device? I might be missing something. 

Also, in `src/rng/backends/CMakeLists.txt` I don't see that MKLGPU backend is only for UNIX. But in README Windows table shows RNG only support MKLCPU backend. I also checked mklgpu backend for RNG and did not see any OS dependency. Are we sure about this missing support?

mcao59: I see that on windows unit_tests on mklgpu are skipped for rng. Also running examples executables gives `function is not implemented` error on mklgpu
So the README should be correct about windows rng only supporting mklcpu backend.

Additionally in commit [effbf47](https://github.com/oneapi-src/oneMKL/pull/200/commits/effbf4756f37a98d7280b924cdf23da85c1f9a51) I added UNIX condition in the other if statements. I was assuming that we will not pass invalid arguments to cmake (like ENABLE_ROCRAND on windows) since in CI we don't have this concern. But I guess it's better to add this additional check in case users pass in invalid settings and see the failure in ctest

examples/rng/run_time_dispatching/CMakeLists.txt
mcao59: updated in [effbf47](https://github.com/oneapi-src/oneMKL/pull/200/commits/effbf4756f37a98d7280b924cdf23da85c1f9a51)

CMakeLists.txt
mcao59: Do you mean indentation in line 149/150? They have two more spaces because they are continuation of the `find_library` function on line 148
As for unit_tests, I'm seeing problem in lapack/getrf on windows (the test hangs for me, the cause of this problem is still in investigation). Therefore unit_test could have the same "no kernel.." problem as examples

mmeterel: Oh, you are right about indentation. My bad. 

Regarding the need for flag in examples: @ericlars do you have any feedback?

mmeterel (approved):

PR #203, opened 2022-06-01
skip the rot and rotg tests when the blas library cannot be found

### LINKED ISSUE #195: [BUG] csrot and zdrot tests fail silently
# Summary
I tried to build the oneMKL tests. But some rot/rotg tests failed because they don't find (and therefore don't load) `libblas.so`, and therefore the functions are not executed and the reference vectors remain unchanged.
Also, we don't have any error message (exception for example) that could help us to fix the bug.

I think you should just need to load the library specified in a new cmake variable, for example `REF_BLAS_LIBNAME` instead of the hardcoded library `libblas.so` [here](https://github.com/oneapi-src/oneMKL/blob/f516ad18ed78268d8c1a8bfcded1bbce27c05235/tests/unit_tests/blas/include/reference_blas_templates.hpp#L56).

You can also add an exception if the library is not found [here](https://github.com/oneapi-src/oneMKL/blob/f516ad18ed78268d8c1a8bfcded1bbce27c05235/tests/unit_tests/blas/include/reference_blas_templates.hpp#L73).

And if the function was not found in the library, [here](https://github.com/oneapi-src/oneMKL/blob/f516ad18ed78268d8c1a8bfcded1bbce27c05235/tests/unit_tests/blas/include/reference_blas_templates.hpp#L72)

# Version
**oneMKL version:** v0.2
**githash:** f516ad1

# Environment
**OS:** Debian 10
**Compiler:** clang version 15.0.0 (intel/llvm 7b2fb02)

# Steps to reproduce
Build oneMKL with the mklcpu backend and use a blas library whose name does not match with `libblas.so`.

# Observed behavior
The tests fail not because the result is different , but because the reference functions were not executed because the library `libblas.so` does not exist. (the reference vectors remain unchanged because the reference functions have not been executed)

# Expected behavior
Tests pass if they find the `csrot` and `zdrot` functions. Fails with **an error and/or an exception** if the library or functions were not found.

Sholde:
# Description

I use the `oneapi::mkl::unimplemented` exception to specify that the blas library can't be found or that the symbols can't be found in the library, because that's the best exception to say what's going on right now in oneMKL.

So I use the *try catch* clause in the rot test files to handle the exception and skip the test when we trigger it. I think this is the best solution because it's not a computational error, so it's not viable to fail the test.

Below you will see a file that shows what happens when the blas library cannot be found:
[log_rotg_skip.log](https://github.com/oneapi-src/oneMKL/files/8815264/log_rotg_skip.log)

Fixes #195 (GitHub issue)

# Checklist

## All Submissions

[logfile.log](https://github.com/oneapi-src/oneMKL/files/8815834/logfile.log)
(All tests passed, except GEMMT but it is not related to this PR, it is due to the library I tested)

## Bug fixes

- [ ] ~~Have you added relevant regression tests?~~ (does not need it)
- [X] Have you included information on how to reproduce the issue (either in a
      GitHub issue or in this PR)?

### DIFF
--- a/tests/unit_tests/blas/level1/rot.cpp
+++ b/tests/unit_tests/blas/level1/rot.cpp
@@ -59,8 +59,14 @@ int test(device *dev, oneapi::mkl::layout layout, int N, int incx, int incy, fp_
     using fp_ref = typename ref_type_info<fp>::type;
     const int N_ref = N, incx_ref = incx, incy_ref = incy;
 
-    ::rot(&N_ref, (fp_ref *)x_ref.data(), &incx_ref, (fp_ref *)y_ref.data(), &incy_ref,
-          (fp_scalar *)&c, (fp_scalar *)&s);
+    try {
+        ::rot(&N_ref, (fp_ref *)x_ref.data(), &incx_ref, (fp_ref *)y_ref.data(), &incy_ref,
+              (fp_scalar *)&c, (fp_scalar *)&s);
+    }
+    catch (const oneapi::mkl::unimplemented &e) { // Did not find the library blas or the symbol
+        std::cout << "Caught synchronous SYCL exception during ROT:\n" << e.what() << std::endl;
+        return test_skipped;
+    }
 
     // Call DPC++ ROT.
 
--- a/tests/unit_tests/blas/include/reference_blas_templates.hpp
+++ b/tests/unit_tests/blas/include/reference_blas_templates.hpp
@@ -56,6 +56,13 @@ static LIB_TYPE cblas_library() {
         h = GET_LIB_HANDLE("libblas.so");
 #endif
     }
+    if (h == NULL) {
+#ifdef _WIN64
+        throw oneapi::mkl::unimplemented("blas", "", "- could not find libblas.dll or blas.dll\n");
+#else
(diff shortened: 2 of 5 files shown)


Sholde:
My pull request is ready for a review by the way.

tests/unit_tests/blas/level1/rot.cpp
```diff
@@ -59,8 +59,14 @@ int test(device *dev, oneapi::mkl::layout layout, int N, int incx, int incy, fp_
     using fp_ref = typename ref_type_info<fp>::type;
     const int N_ref = N, incx_ref = incx, incy_ref = incy;
 
-    ::rot(&N_ref, (fp_ref *)x_ref.data(), &incx_ref, (fp_ref *)y_ref.data(), &incy_ref,
-          (fp_scalar *)&c, (fp_scalar *)&s);
+    try {
```
andrewtbarker, line 62: I would leave this code as is, so the exception goes through and the test fails if the library is not found.

mmeterel:
@Sholde  As discussed in #195, I would recommend to update the build system and catch the described issue at build time. If no objections, I would like to close this PR. If we can't find a built time solution, we can re-open this PR.

Sholde:
No problem.

PR #209, opened 2022-06-17
[LAPACK][CUSOLVER] Add potrf and getrs batch functions to cuSolver

AidanBeltonS:
# Description

This PR extends `potrf_batch` to additional overloads. Additionally it implements all `getrs_batch` overloads. This change fully implements both `potrf_batch` and `getrs_batch` for USM and Buffer for both group and strided batch operations.
In neither case does cuSolver have a direct equivalent to the oneMKL function.

So in both cases some manipulation has to occur to make things work. `potrf_batch` implements the strided batch operation with the cuSolver group batch operation. This should be quite efficient, as it is just reformatting the input. `getrs_batch` is implemented using non-batched functions so this just loops over the cuSolver function for each matrix in the batch. The performance of this would not be significantly different than the user just looping over the oneMKL function themselves. This I believe this is okay, as it would provide additional convenience and portability between backends to the user.

The motivation for this change is to add greater functionality to the cuSolver backend and to improve portability between cuSolver and other lapack backends. Additionally this would provide a framework for further implementations of missing batch functions which do not align well with cuSolver functions.

Edit: Follow up commit implements remaining batched functions, with the exception of getrf group batched due to an issue with test.

Test logs:
[potrf_and_getrs_results.txt](https://github.com/oneapi-src/oneMKL/files/8926876/potrf_and_getrs_results.txt)


AidanBeltonS:
Remaining batch functions implemented using same method as potrf and getrs functions.

Getrf group batched operation commented out and set to unimplemented due to test segfaulting. Error appears to not be related to oneMKL but lapack. Further investigation will be done on it.

Test results: Some non-batched tests are failing locally, this occurs with and without this patch.
[test_result.txt](https://github.com/oneapi-src/oneMKL/files/8968798/test_result.txt)

ericlars:
Thanks for the PR @AidanBeltonS

Can you provide more details on the non batch failures? Setting the environment variable CTEST_OUTPUT_ON_FAILURE=1 with ctest can provide more details. Unusual errors are sometimes due to linking against Netlib libraries compiled with 32 bit integers, can you verify your reference libraries were compiled for 64bit integers?

I did see some batch failures relating to illegal memory accesses and invalid parameters passed to cuda. Log attached [here](https://github.com/oneapi-src/oneMKL/files/9014792/log_titan.txt). I didn't see anything immediately wrong in the parameter passing but I'll continue to look.

AidanBeltonS:
> Can you provide more details on the non batch failures? Setting the environment variable CTEST_OUTPUT_ON_FAILURE=1 with ctest can provide more details. Unusual errors are sometimes due to linking against Netlib libraries compiled with 32 bit integers, can you verify your reference libraries were compiled for 64bit integers?

I have attached the results below with ctest output on failures. I have double checked I have built the netlib libraries for 64bits and am linking with the correct ones.

[results.txt](https://github.com/oneapi-src/oneMKL/files/9140623/results.txt)




> I did see some batch failures relating to illegal memory accesses and invalid parameters passed to cuda. Log attached [here](https://github.com/oneapi-src/oneMKL/files/9014792/log_titan.txt). I didn't see anything immediately wrong in the parameter passing but I'll continue to look.

Thanks for the log. I will look into the failing tests.

AidanBeltonS:
@ericlars I found the issue with `getrf_batch_group` the problem was with my implementation, which was causing an odd failure in the reference lapack check. This has been resolved.


> I did see some batch failures relating to illegal memory accesses and invalid parameters passed to cuda. Log attached [here](https://github.com/oneapi-src/oneMKL/files/9014792/log_titan.txt). I didn't see anything immediately wrong in the parameter passing but I'll continue to look.

I have not been able to reproduce the errors you have found however I have looked further into the issue of non-batched operations failing. The problem appears to be with the addition of multiple CUDA streams per SYCL queue. This likely messes with the cusolver scope handler which assumes there is one stream per device per thread. It may be that the batch test failures you are seeing are also caused by the multistream changes. Would you be able to re-run the tests on a commit before the addition of multiple streams per queue? The DPCPP commit just before multi-streams is: d149ec39e7791a2d70858a7cf10261d6353b01be

I have attached the test logs below.
Commit: d149ec39e7791a2d70858a7cf10261d6353b01be
[results-no-multistream.txt](https://github.com/oneapi-src/oneMKL/files/9157988/results-no-multistream.txt)

Commit: dd418459868a976cd2eeae367fea6b92795ea611
[results-with-multistream.txt](https://github.com/oneapi-src/oneMKL/files/9157990/results-with-multistream.txt)

ericlars:
@AidanBeltonS My logs were from a build of llvm from 3/19, and it looks like the multistream patch was committed on 5/17, but for sanity I'll try building from your suggested commits. Thanks for  the sleuthing.

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -937,77 +1690,106 @@ std::int64_t getri_batch_scratchpad_size<std::complex<double>>(sycl::queue &queu
...
+        geqrf_batch_scratchpad_size(#CUSOLVER_ROUTINE, CUSOLVER_ROUTINE, queue, m, n, lda, \
+                                    group_count, group_sizes, &scratch_size);              \
+        return scratch_size;                                                               \
+    }
+
+GEQRF_GROUP_LAUNCHER_SCRATCH(float, cusolverDnSgetrf_bufferSize)
+GEQRF_GROUP_LAUNCHER_SCRATCH(double, cusolverDnDgetrf_bufferSize)
+GEQRF_GROUP_LAUNCHER_SCRATCH(std::complex<float>, cusolverDnCgetrf_bufferSize)
+GEQRF_GROUP_LAUNCHER_SCRATCH(std::complex<double>, cusolverDnZgetrf_bufferSize)
+
```
ericlars, line 1781: getrf -> geqrf

Found the source for geqrf failures.

AidanBeltonS: Thank you for catching the mistake! I have fixed the scratch function

AidanBeltonS: Is `GetrsBatchStride` still failing? I found a mistake where there was a call to free in the queue submit and not the host_task, this was fixed in previous commit.

JackAKirk:
> @AidanBeltonS My logs were from a build of llvm from 3/19, and it looks like the multistream patch was committed on 5/17, but for sanity I'll try building from your suggested commits. Thanks for the sleuthing.

New failures confirmed as resulting from multi-streams: although it is really a bug in oneMKL because interop streams are not synced. Proposed fix here: https://github.com/oneapi-src/oneMKL/pull/215

AerialMantis:
@ericlars now that https://github.com/oneapi-src/oneMKL/pull/215 has been merged the issues resulting from the multiple CUDA stream implementation in DPC++ should be resolved, would you be able to review this again?

AidanBeltonS:
I have updated this PR to use the new functionality introduced in #215

ericlars:
Looks good to me!

[log_llvm_cusolver_.txt](https://github.com/oneapi-src/oneMKL/files/9561469/log_llvm_cusolver_.txt)

ericlars (approved):

src/lapack/backends/cusolver/cusolver_helper.hpp
```diff
@@ -274,26 +280,43 @@ struct CudaEquivalentType<std::complex<double>> {
...
+                val);
+    }
+}
+
+/* batched helpers */
+
+// Creates list of matrix/vector pointers from initial ptr and stride
+// Note: user is responsible for deallocating memory
+template <typename T>
+T **create_ptr_list_from_stride(T *ptr, int64_t ptr_stride, int64_t batch_size) {
```
Alexander-Kleymenov, line 314: it is introduced, but never used?

ericlars: @Alexander-Kleymenov I believe this func is used in cusolver_batch.cpp in the same directory for translating the strided APIs.

src/lapack/backends/cusolver/cusolver_helper.hpp
```diff
@@ -274,26 +280,43 @@ struct CudaEquivalentType<std::complex<double>> {
...
+            throw oneapi::mkl::lapack::computation_error(
+                func_name, std::string(cufunc_name) + " failed with info = " + std::to_string(val),
+                val);
+    }
+}
+
+/* batched helpers */
+
+// Creates list of matrix/vector pointers from initial ptr and stride
+// Note: user is responsible for deallocating memory
```
Alexander-Kleymenov, line 312: Looks like none of the usages follow this?

src/lapack/backends/cusolver/cusolver_helper.hpp
Alexander-Kleymenov: ok, I see

src/lapack/backends/cusolver/cusolver_helper.hpp
AidanBeltonS: Thanks for catching this. I have updated this to free all the instances where this allocates memory

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -79,251 +100,645 @@ void getri_batch(sycl::queue &queue, std::int64_t n, sycl::buffer<std::complex<d
...
+        onemkl_cusolver_host_task(cgh, queue, [=](CusolverScopedContextHandler &sc) {
+            auto handle = sc.get_handle(queue);
+            auto a_ = sc.get_mem<cuDataType *>(a_acc);
+            auto ipiv_ = sc.get_mem<int *>(ipiv32_acc);
+            auto devInfo_ = sc.get_mem<int *>(devInfo_acc);
+            auto scratch_ = sc.get_mem<cuDataType *>(scratch_acc);
+            cusolverStatus_t err;
+
+            // Uses scratch so sync between each cuSolver call
+            for (int64_t i = 0; i < batch_size; ++i) {
```
Alexander-Kleymenov: I don't know how strict it is, but int64_t here is used without std namespace, a bit different vs everywhere else, and 32 bit int type is used not as int32_t.

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -79,251 +100,645 @@ void getri_batch(sycl::queue &queue, std::int64_t n, sycl::buffer<std::complex<d
...
+                                           lda, scratchpad_, ipiv_ + stride_ipiv * i, devInfo_ + i);
+            }
+        });
+    });
+
+    // Copy from 32-bit USM to 64-bit
+    sycl::event done_casting = queue.submit([&](sycl::handler &cgh) {
+        cgh.depends_on(done);
+        cgh.parallel_for(sycl::range<1>{ ipiv_size }, [=](sycl::id<1> index) {
+            ipiv[index] = static_cast<std::int64_t>(ipiv32[index]);
```
Alexander-Kleymenov: Original ipiv could be passed as *int32_t, and then resulting 32-bit values could be expanded to 64-bit (in-place).

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -79,251 +100,645 @@ void getri_batch(sycl::queue &queue, std::int64_t n, sycl::buffer<std::complex<d
...
+    for (int64_t group_id = 0, global_id = 0; group_id < group_count; ++group_id) {
+        uint64_t ipiv_size = n[group_id];
+        for (int64_t local_id = 0; local_id < group_sizes[group_id]; ++local_id, ++global_id) {
+            int64_t *d_ipiv = ipiv[global_id];
+            int *d_ipiv32 = ipiv32[global_id];
+
+            sycl::event e = queue.submit([&](sycl::handler &cgh) {
+                cgh.depends_on(done);
+                cgh.parallel_for(sycl::range<1>{ ipiv_size }, [=](sycl::id<1> index) {
+                    d_ipiv[index] = static_cast<std::int64_t>(d_ipiv32[index]);
```
Alexander-Kleymenov: why using static_cast here?

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -79,251 +100,645 @@ void getri_batch(sycl::queue &queue, std::int64_t n, sycl::buffer<std::complex<d
...
+        }
+    }
+
+    // Enqueue free memory
+    sycl::event done_freeing = queue.submit([&](sycl::handler &cgh) {
+        int64_t num_events = casting_dependencies.size();
+        for (int64_t i = 0; i < num_events; i++) {
+            cgh.depends_on(casting_dependencies[i]);
+        }
+        cgh.host_task([=](sycl::interop_handle ih) {
```
Alexander-Kleymenov, line 708: this would not be required if ipiv was passed as **int32_t, and converted in-pace later,

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -79,251 +100,645 @@ void getri_batch(sycl::queue &queue, std::int64_t n, sycl::buffer<std::complex<d
...
+                        oneapi::mkl::uplo uplo, std::int64_t n, std::int64_t nrhs,
+                        sycl::buffer<T> &a, std::int64_t lda, std::int64_t stride_a,
+                        sycl::buffer<T> &b, std::int64_t ldb, std::int64_t stride_b,
+                        std::int64_t batch_size, sycl::buffer<T> &scratchpad,
+                        std::int64_t scratchpad_size) {
+    using cuDataType = typename CudaEquivalentType<T>::Type;
+
+    overflow_check(n, nrhs, lda, ldb, stride_a, stride_b, batch_size, scratchpad_size);
+
+    // cusolver function only supports nrhs = 1
```
Alexander-Kleymenov: cuSolver is referred with different letter casing across the comments.

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -79,251 +100,645 @@ void getri_batch(sycl::queue &queue, std::int64_t n, sycl::buffer<std::complex<d
...
+        cgh.depends_on(done);
+        cgh.parallel_for(sycl::range<1>{ ipiv_size }, [=](sycl::id<1> index) {
+            ipiv[index] = static_cast<std::int64_t>(ipiv32[index]);
+        });
+    });
+
+    // Enqueue free memory, don't return event as not-neccessary for user to wait for ipiv32 being released
+    queue.submit([&](sycl::handler &cgh) {
+        cgh.depends_on(done_casting);
+        cgh.host_task([=](sycl::interop_handle ih) { sycl::free(ipiv32, queue); });
```
Alexander-Kleymenov, line 607: host_task lambda here can capture queue by reference, avoiding object copying.

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -550,37 +1186,76 @@ POTRF_BATCH_LAUNCHER_USM(std::complex<double>, cusolverDnZpotrfBatched)
...
+                               std::int64_t stride_b, std::int64_t batch_size, T *scratchpad,
+                               std::int64_t scratchpad_size,
+                               const std::vector<sycl::event> &dependencies) {
+    using cuDataType = typename CudaEquivalentType<T>::Type;
+
+    overflow_check(n, nrhs, lda, ldb, stride_a, stride_b, batch_size, scratchpad_size);
+
+    // cusolver function only supports nrhs = 1
+    if (nrhs != 1)
+        throw unimplemented("lapack", "potrs_batch", "cusolver potrs_batch only supports nrhs = 1");
```
Alexander-Kleymenov, line 1199: if at some point the support will be expanded, is there any way to bypass this check?

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -652,68 +1330,153 @@ POTRS_BATCH_LAUNCHER_USM(double, cusolverDnDpotrsBatched)
...
+        .submit([&](sycl::handler &cgh) {
+            onemkl_cusolver_host_task(cgh, queue, [=](CusolverScopedContextHandler &sc) {
+                auto handle = sc.get_handle(queue);
+                cusolverStatus_t err;
+
+                CUSOLVER_ERROR_FUNC_T(func_name, func, err, handle, m, n, nullptr, lda,
+                                      scratch_size);
+            });
+        })
+        .wait();
```
Alexander-Kleymenov: something happened to code formatting in this function

Alexander-Kleymenov (approved):
Looks good.

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: Could you explain further how the 32-bit values can be expanded to 64-bit in place? Are you suggesting use the 64-bit memory as if it were 32-bit then sequentially shift each value to the correct place?

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: I have added the `std` namespace. 

> and 32 bit int type is used not as int32_t.

Could you clarify what you are referring to, which 32 bit int type is not used as a int32_t and in what way?

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: I have removed the static_cast

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: Done

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: Done

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: There is no way to check if cuSolver supports this nrhs > 1 case. If support is expanded this check will have to be modified or deleted depending on the circumstances at the time.

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: Fixed formatting

src/lapack/backends/cusolver/cusolver_batch.cpp
Alexander-Kleymenov: The comment regarding legacy API makes use of 32-bit int a special case vs use of int64_t, so this special case could be explicitly highlighted with explicit fixed width type int32_t, not implementation dependent int type. In the current building environment int equals to int32_t. Use of int32_t would be just for code readability (highlight special case) and consistency (use only fixed-width integer types). This is just for your consideration.

src/lapack/backends/cusolver/cusolver_batch.cpp
Alexander-Kleymenov: Yes, this is the idea. It would reduce amount of code (especially in array-of-pointers case), and runtime overhead.

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -79,251 +100,642 @@ void getri_batch(sycl::queue &queue, std::int64_t n, sycl::buffer<std::complex<d
...
+    sycl::event done_casting = queue.submit([&](sycl::handler &cgh) {
+        cgh.depends_on(done);
+        cgh.parallel_for(sycl::range<1>{ ipiv_size },
+                         [=](sycl::id<1> index) { ipiv[index] = ipiv32[index]; });
+    });
+
+    // Enqueue free memory, don't return event as not-neccessary for user to wait for ipiv32 being released
+    queue.submit([&](sycl::handler &cgh) {
+        cgh.depends_on(done_casting);
+        cgh.host_task([&](sycl::interop_handle ih) { sycl::free(ipiv32, queue); });
```
Alexander-Kleymenov: This will lead to sporadic crashes: passing queue by reference makes sense as queue object is not going to be destroyed before host_tasl lambda is finished, but ipiv32 variable, which is automatic stack-allocated, can be destroyed by the time of host_task execution, and contain some other data, so free it will crash. `[=,&queue]`

src/lapack/backends/cusolver/cusolver_batch.cpp
Alexander-Kleymenov: Could it be some single build-time constant, or preprocessor macro, switching it in single place. Like `static constexpr bool multiple_nrhs_support = false;` ?

src/lapack/backends/cusolver/cusolver_batch.cpp
Alexander-Kleymenov: Thank you for the updates.

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: Okay, this would be an interested and likely more efficient approach. 

I think however it should be separated to this PR as the non-batched operations use a kernel for casting between types, same as here, so this change would be out of scope of this batch PR. Additionally, the change would be done for performance reasons so it would be good to measure if there is any performance benefit to the expanding method before implementing.

My thought is that it may make sense to create an issue to track this change so a future PR can be made once performance measurements have confirmed this approach is better and the PR can address both batched and non-batched implementations. Would this approach work for you?

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: Okay, I have wrapped the check with a macros preprocessor, so if the user defines `POTRS_BATCHED_MULTIPLE_NRHS_SUPPORTED` then the check will be removed.

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: I have reverted this change

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: I agree it would be better to have this as explicitly stated as `int32_t` rather than `int`. I think it should be a separate PR as the non-batched implementation also uses `int` in this way so it would be out of scope of this PR. I am happy to make a follow up PR to change this for both implementations.

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -79,251 +100,644 @@ void getri_batch(sycl::queue &queue, std::int64_t n, sycl::buffer<std::complex<d
...
+                        sycl::buffer<T> &a, std::int64_t lda, std::int64_t stride_a,
+                        sycl::buffer<T> &b, std::int64_t ldb, std::int64_t stride_b,
+                        std::int64_t batch_size, sycl::buffer<T> &scratchpad,
+                        std::int64_t scratchpad_size) {
+    using cuDataType = typename CudaEquivalentType<T>::Type;
+
+    overflow_check(n, nrhs, lda, ldb, stride_a, stride_b, batch_size, scratchpad_size);
+
+    // cuSolver function only supports nrhs = 1
+#ifndef POTRS_BATCHED_MULTIPLE_NRHS_SUPPORTED
```
Alexander-Kleymenov: There is C++ way of doing this, not involving preprocessor's macro (`if constexpr (not cusolver_batched_potrs_supports_multiple_nrhs) { if (nrhs != 1) { throw } }`).
Here it is also not clear from the code if it is supported, or not, without explicit definition.
Comment on line 346 got detached. Better to remove it as the code is quite self-documenting here.

src/lapack/backends/cusolver/cusolver_batch.cpp
Alexander-Kleymenov: Batch case is different from regular by the size of problems: regular API is intended for larger cases, and this pre/post call overheads amortized by long kernel computations time. Batch case works on smaller problems, and run-time is also pretty short (due to high parallelism allowing to handle lots of small problems simultaneously). So any additional overhead is significant. Do you have any measurements of corresponding CUDA API on NV HW to compare it with the performance done through these SYCL interfaces?

Alexander-Kleymenov: > Would this approach work for you?

Would work fine.

AidanBeltonS: > Do you have any measurements of corresponding CUDA API on NV HW to compare it with the performance done through these SYCL interfaces?

I currently do not have any measurements to compare the performance of this backend to native cuSolver

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: I think the issue with this approach is where is `cusolver_batched_potrs_supports_multiple_nrhs` going to reside? If this is to be set in a single place it would have to be a global variable, which is undesirable, as each of these are free functions.

Personally I think having this check done this way makes sense as we are using the legacy API. It seems very unlikely that NVIDIA will come out with an update eliminating this constraint for an API they are probably going to deprecate at some point. So any support to allow `nrhs > 1` would require a change to the code and therefore the check can be simply deleted at that time. However, I still do not have any issues with making a way to by pass it. Just noting that it is unlikely to be needed.

Alexander-Kleymenov: > I think the issue with this approach is where is cusolver_batched_potrs_supports_multiple_nrhs going to reside? If this is to be set in a single place it would have to be a global variable, which is undesirable, as each of these are free functions.

This is not a variable, but a constant, leaving no trace in the binary. Could be declared at the beginning of the source.

> It seems very unlikely that NVIDIA will come out with an update eliminating this constraint for an API they are probably going to deprecate at some point. 

This is a good point. So we can omit on this expectation.
Documentation explicitly states `The routine will be removed in the next major release.` Why we are using it here if it won't work in a year?

Alexander-Kleymenov (approved):

mkrainiuk (approved):

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: I have removed the macro.
The reason to use the legacy API is that the new API is not yet fleshed out. There would be a lot of missing functions, I would expect that when the legacy API is deprecated they will also bring in more supported operations for the new API.

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: Issue to track future change: https://github.com/oneapi-src/oneMKL/issues/230

src/lapack/backends/cusolver/cusolver_batch.cpp
Alexander-Kleymenov: > The reason to use the legacy API is that the new API is not yet fleshed out.

I skimmed over the spec and all API, having deprecation notices, also have a proposed new API to use. What is missed?

AidanBeltonS: If you take a look at the number of supported lapack operations, the 64-bit API does not support nearly as many operations. For example the 64-bit dense eigenvalue solver section only really contains two lapack operations. You can then see that the legacy API has many more. 
Legacy API eigenvalue solvers: https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverDN-eigensolver-reference
64-bit API eigenvalue solvers: https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverDN-eigensolver-reference-64bit

Alexander-Kleymenov: Isn't this update only concerned with potrs and getrs which are being deprecated and have new API proposed?

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -79,251 +100,642 @@ void getri_batch(sycl::queue &queue, std::int64_t n, sycl::buffer<std::complex<d
...
+                        sycl::buffer<T> &b, std::int64_t ldb, std::int64_t stride_b,
+                        std::int64_t batch_size, sycl::buffer<T> &scratchpad,
+                        std::int64_t scratchpad_size) {
+    using cuDataType = typename CudaEquivalentType<T>::Type;
+
+    overflow_check(n, nrhs, lda, ldb, stride_a, stride_b, batch_size, scratchpad_size);
+
+    // cuSolver function only supports nrhs = 1
+    if (nrhs != 1)
+        throw unimplemented("lapack", "potrs_batch", "cusolver potrs_batch only supports nrhs = 1");
```
Alexander-Kleymenov, line 348: BTW, I don't see any deprecation notices for cusolverDnDpotrsBatched

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -79,251 +100,642 @@ void getri_batch(sycl::queue &queue, std::int64_t n, sycl::buffer<std::complex<d
...
+    // To get around the limitation.
+    // Create new buffer and convert 64-bit values.
+    std::uint64_t ipiv_size = stride_ipiv * batch_size;
+    sycl::buffer<int, 1> ipiv32(sycl::range<1>{ ipiv_size });
+
+    queue.submit([&](sycl::handler &cgh) {
+        auto ipiv32_acc = ipiv32.template get_access<sycl::access::mode::write>(cgh);
+        auto ipiv_acc = ipiv.template get_access<sycl::access::mode::read>(cgh);
+        cgh.parallel_for(sycl::range<1>{ ipiv_size }, [=](sycl::id<1> index) {
+            ipiv32_acc[index] = static_cast<std::int32_t>(ipiv_acc[index]);
```
Alexander-Kleymenov, line 126: Do you think if it makes sense to have overflow checks for integer arrays same as on line 114?

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -79,251 +100,642 @@ void getri_batch(sycl::queue &queue, std::int64_t n, sycl::buffer<std::complex<d
...
+    using cuDataType = typename CudaEquivalentType<T>::Type;
+
+    overflow_check(m, n, lda, stride_a, stride_ipiv, batch_size, scratchpad_size);
+
+    // cuSolver legacy api does not accept 64-bit ints.
+    // To get around the limitation.
+    // Create new buffer with 32-bit ints then copy over results
+    std::uint64_t ipiv_size = stride_ipiv * batch_size;
+    sycl::buffer<int, 1> ipiv32(sycl::range<1>{ ipiv_size });
+    sycl::buffer<int> devInfo{ batch_size };
```
Alexander-Kleymenov, line 187: `sycl::buffer<int>` is equivalent to `sycl::buffer<int, 1>`

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: The legacy API functions are not "deprecated" functions. They just follow an old interface, have fewer features, and will be made deprecated in the future.

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: I think we can leave this without checks. While it would be nice to provide the user more error information, if the user provided valid dimensions but then requests out of bound `ipiv` values then that is more of an issue on their end.
There is also downside of having to check every `ipiv` value across the batches then copy the results back to the host to decide to throw an exception or not. So this means more memory being copied from device to host and the host cannot just queue up as much asynchronous work as possible and move on.

src/lapack/backends/cusolver/cusolver_batch.cpp
AidanBeltonS: Thanks for catching this, I have removed the dim value.

Alexander-Kleymenov (approved):
Thanks for the updates. Looks good!

AerialMantis:
@ericlars it looks like we have approvals for this, could this be merged now, or is there any further discussions to which need to be resolved?

Issue #204, opened 2022-06-03
BLAS: tests may not be executing the right backend


adegomme:
# Summary
When trying a new library we encountered an unexpected behaviour, as it seems that for BLAS (at least), most tests are actually comparing the reference libblas to itself (but wrapped in sycl), not to the actual intended backend. This can be reproduced with MKL.

As I understand, unit_tests such as axpy_usm.cpp are compiled and linked with reference libblas, hence cblas_xxx symbols are already resolved at the start of execution.

When the test ( for example RT test AxpyUsmTestSuite/AxpyUsmTests.RealSinglePrecision/Column_Major_SYCL_host_device ) is executed, it runs the reference call, and then tries the sycl backend tested (mklcpu for example), by dlopening the corresponding libonemkl_blas_mklcpu.so lib. But as this dlopen is made lazily and not deeply, the cblas_saxpy symbol still remains the one from libcblas. This is for the RT variant of the test, as the CT variant does not dynamically load anything, so the problem is worse for these.

gdb trace, with breaks on cblas_saxpy and dlopen

gdb --args ./bin/test_main_blas_rt "--gtest_filter=AxpyUsmTestSuite/AxpyUsmTests.RealSinglePrecision/Column_Major_SYCL_host_device" "--gtest_also_run_disabled_tests"

first breakpoint is the reference call, 2 is the dlopen and 3 is the call which should be from the mkl :

```
[ RUN      ] AxpyUsmTestSuite/AxpyUsmTests.RealSinglePrecision/Column_Major_SYCL_host_device

Breakpoint 1, 0x00007ffff7f9cdd0 in cblas_saxpy () from /foo/lapack/build/lib/libcblas.so.3
(gdb)
Continuing.

Breakpoint 2, ___dlopen (file=0x7ffff7f0c020 "libonemkl_blas_mklcpu.so", mode=257) at ./dlfcn/dlopen.c:77
77      in ./dlfcn/dlopen.c
(gdb) c
Continuing.
[New Thread 0x7fffe79ff640 (LWP 19332)]
[Switching to Thread 0x7fffe79ff640 (LWP 19332)]

Thread 2 "test_main_blas_" hit Breakpoint 1, 0x00007ffff7f9cdd0 in cblas_saxpy () from /foo/lapack/build/lib/libcblas.so.3
(gdb) bt
#0  0x00007ffff7f9cdd0 in cblas_saxpy () from /foo/lapack/build/lib/libcblas.so.3
[...]

PR #210, opened 2022-06-29
Use external interfaces in oneMKL for mklcpu and mklgpu backends

### LINKED ISSUE #114: Investigate switching to external oneMKL DPC++ APIs for mklcpu/mklgpu backend
# Summary
This issues is related to Intel oneMKL symbols used in mklgpu backend. Currently internal symbols are used to call oneMKL routines. However, these symbols can chance between releases, which will cause build failures in oneMKL open source interfaces. 

# Problem statement
The internal symbols can change between releases and cause build failures. 

# Details
Switching to external Intel oneMKL APIs will improve stability of the repo between releases.

andrewtbarker:
# Description

Replaces calls to internal oneMKL functions with calls to the usual external interface. This should make the mklcpu and mklgpu backends more robust and predictable.

Fixes #114 

112 out of 3338 tests fail after these changes due to the underlying issue in #204. In fact these tests were never really passing because we were previously testing the reference Netlib BLAS against itself rather than testing the mklcpu backend. 

The tests are intended to check the `cblas_` interface to reference BLAS against DPC++ interfaces to oneMKL BLAS. Previously these used the same symbols, so we tested reference BLAS against itself. In most cases ( [example](https://github.com/oneapi-src/oneMKL/blob/722f6c51b528a2bc09bf549b1097f7d8fdc37826/src/blas/backends/mklcpu/mklcpu_level1.cxx#L76) ) we tested the `cblas_` interface against itself. In a few cases that use an increment parameter ( [example](https://github.com/oneapi-src/oneMKL/blob/722f6c51b528a2bc09bf549b1097f7d8fdc37826/src/blas/backends/mklcpu/mklcpu_level1.cxx#L29) ), we tested instead the fortran interface. These latter tests are exactly the ones that fail with the changes in this PR because negative values for the increment parameter are handled differently in the two cases.

Fixing the tests will require more serious changes to how the project as a whole handles dynamic libraries as documented in #204.

# Checklist

## All Submissions

[test.txt](https://github.com/oneapi-src/oneMKL/files/9014351/test.txt)


src/blas/backends/mkl_common/mkl_batch.cxx
```diff
@@ -0,0 +1,736 @@
+/*******************************************************************************
+* Copyright 2020-2021 Intel Corporation
```
mmeterel: 2020-2021 ==> 2022

src/blas/backends/mkl_common/mkl_extensions.cxx
```diff
@@ -0,0 +1,156 @@
+/*******************************************************************************
+* Copyright 2020-2021 Intel Corporation
```
mmeterel: 2020-2021 => 2022

src/blas/backends/mkl_common/mkl_level1.cxx
```diff
@@ -0,0 +1,645 @@
+/*******************************************************************************
+* Copyright 2020-2021 Intel Corporation
```
mmeterel: 2020-2021 => 2022

src/blas/backends/mkl_common/mkl_level2.cxx
```diff
@@ -0,0 +1,862 @@
+/*******************************************************************************
+* Copyright 2020-2021 Intel Corporation
```
mmeterel: 2020-2021 => 2022

src/blas/backends/mkl_common/mkl_level3.cxx
```diff
@@ -0,0 +1,519 @@
+/*******************************************************************************
+* Copyright 2020-2021 Intel Corporation
```
mmeterel: 2020-2021 => 2022

src/blas/backends/mklcpu/mklcpu_level3.cpp
```diff
@@ -23,34 +23,26 @@
...
 
-#define CBLASMAJOR CblasColMajor
-#define COLUMN_MAJOR
-#define MKLMAJOR MKL_COL_MAJOR
-#include "mklcpu_level3.cxx"
-#undef CBLASMAJOR
-#undef COLUMN_MAJOR
-#undef MKLMAJOR
+namespace blas_major = ::oneapi::mkl::blas::column_major;
+// TODO: some macro that asserts the queue is a cpu queue?
```
mmeterel: What is the concern here? For compile time, we rely on the user to specify the matching device and environment variable (ENABLE_MKLCPU_BACKEND). For runtime, we select the backend based on device type in queue. Are you referring to compile time case? Anyhow, we would need a global solution for that case.

src/blas/backends/mklgpu/mklgpu_level3.cpp
```diff
@@ -24,26 +24,25 @@
...
 namespace mkl {
 namespace blas {
 namespace mklgpu {
 namespace column_major {
 
-#define MAJOR MKL_COL_MAJOR
-#include "mklgpu_level3.cxx"
-#undef MAJOR
+namespace blas_major = ::oneapi::mkl::blas::column_major;
+// TODO: some macro that asserts the queue is a gpu queue?
```
mmeterel: same comment as before

src/blas/backends/mkl_common/mkl_batch.cxx
andrewtbarker: Fixed, thanks.

src/blas/backends/mklcpu/mklcpu_level3.cpp
andrewtbarker: Not sure what I was worried about, comment removed.

mmeterel (approved):
Thanks for the PR Andrew. Could you please add a note to #204 about the fact that, nrm2/scal/asum tests will start failing once we merge this PR? These failures needs to be fixed as part of #204.

andrewtbarker:
@mkrainiuk Can you take a look at this PR please when you get a chance?

src/blas/backends/mkl_common/mkl_extensions.cxx
```diff
@@ -0,0 +1,156 @@
...
+* SPDX-License-Identifier: Apache-2.0
+*******************************************************************************/
+
+// Buffer APIs
+
+void gemm_bias(sycl::queue &queue, transpose transa, transpose transb, offset offsetc, int64_t m,
+               int64_t n, int64_t k, float alpha, sycl::buffer<int8_t, 1> &a, int64_t lda,
+               int8_t ao, sycl::buffer<int8_t, 1> &b, int64_t ldb, int8_t bo, float beta,
+               sycl::buffer<int32_t, 1> &c, int64_t ldc, sycl::buffer<int32_t, 1> &co) {
+    blas_major::gemm_bias(queue, transa, transb, offsetc, m, n, k, alpha, a, lda, ao, b, ldb, bo,
```
mkrainiuk, line 26: Does new DPC++ API require Intel oneMKL sycl library for MKLCPU backend? Because as far as I remember only LAPACK adds sycl library for MKLCPU backend, all other domains build with only 3 libraries.

mkrainiuk: Here is how sycl library is enabled in LAPACK MKLCPU backend https://github.com/oneapi-src/oneMKL/blob/3fe1f9bb897256d8c4d8fcb0de258b12c17c5a87/src/lapack/backends/mklcpu/CMakeLists.txt#L23

andrewtbarker: Good catch!

I've made similar changes in `src/blas/backends/mklcpu/CMakeLists.txt`. I'm testing now building with only mklcpu backend - this worked for me before probably because I always built with GPU backend enabled, even if I ran tests on CPU backend.

andrewtbarker: With change to `CMakeLists.txt`, a build with only mklcpu backend builds and the tests run.

andrewtbarker:
Ping @mkrainiuk can you approve if you don't have additional concerns? Thanks!

mkrainiuk (approved):
Looks good to me, thank you!

PR #213, opened 2022-07-26
[LAPACK][CUSOLVER] Add devInfo check to sytrd

AidanBeltonS:
# Description

This is a simple fix to an issue within cuSolvers sytrd implementation. For some cuSolver functions not passing `devInfo` causes an error to be thrown. In these cases `devInfo` is passed to the function and checked if the result is acceptable. Sytrd was missing this, depending on the machine this would fail. This change modifies the Sytrd function to match its Hetrd, its complex equivalent. See: https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverDN-lt-t-gt-sytrd

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Attach a log. [Sytrdresults.txt](https://github.com/oneapi-src/oneMKL/files/9188885/Sytrdresults.txt)


ericlars:
Thanks for the fix!

ericlars (approved):

sknepper (approved):
LGTM - thanks!

PR #214, opened 2022-07-28
[RNG] Enable mklgpu backend for Windows

aelizaro:
# Description

Enable RNG domain for mklgpu backend

# Checklist

## All Submissions

[rng_ctest_log.txt](https://github.com/oneapi-src/oneMKL/files/9207830/rng_ctest_log.txt)
Note: Double precision tests are skipped as card doesn't support it


mkrainiuk (approved):
Looks good to me, thank you!

CMakeLists.txt
```diff
@@ -144,11 +144,11 @@ if(WIN32 AND ONEMKL_SYCL_IMPLEMENTATION STREQUAL "dpc++")
   endforeach()
   set(CMAKE_CXX_COMPILE_OBJECT "<CMAKE_CXX_COMPILER> -fsycl /nologo <DEFINES> <INCLUDES> /EHsc <FLAGS> /Fo<OBJECT> -c <SOURCE>")
   set(CMAKE_CXX_CREATE_STATIC_LIBRARY "lib /nologo <OBJECTS> /out:<TARGET>")
-  set(CMAKE_CXX_LINK_EXECUTABLE "<CMAKE_CXX_COMPILER> -fsycl /nologo <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
+  set(CMAKE_CXX_LINK_EXECUTABLE "<CMAKE_CXX_COMPILER> -fsycl -fsycl-device-code-split=per_kernel /nologo <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
```
mmeterel, line 147: Could you explain why you need this flag? How does this impact other domains?

aelizaro: it is needed for graphic cards that don't support double precision; otherwise, all tests fail if mkl is linked statically (as I remember, we have such cases). It affects only DPC++ kernels - so if other domains have DPC++ kernels, it would help, but the link time would be increased. If domains don't rely on DPC++ kernels - there should be no effect at all.

mmeterel: @aelizaro Thanks for the explanation!

mmeterel (approved):

Issue #216, opened 2022-07-29
[CUSOLVER] cuSOLVER handler does not support multiple streams


ericlars:
# Summary
[Intel llvm](https://github.com/intel/llvm/commit/dd418459868a976cd2eeae367fea6b92795ea611
) introduced multiple CUDA streams per SYCL queue which broke the cuSOLVER scope handler that assumed one stream per device per thread. This broke asynchronous submissions, in the short term synchronous cuSOLVER submissions should be used. See #209 and #215 for further details.

# Version
All versions built with intel llvm after dd418459868a976cd2eeae367fea6b92795ea611.

# Environment
cuSOLVER backend

mmeterel:
@ericlars is this fixed and can be closed?

PR #217, opened 2022-07-30
[CUSOLVER] workaround for issue #216

ericlars:
RE: #215

### DIFF
--- a/src/lapack/backends/cusolver/cusolver_task.hpp
+++ b/src/lapack/backends/cusolver/cusolver_task.hpp
@@ -43,6 +43,7 @@ static inline void host_task_internal(H &cgh, sycl::queue queue, F f) {
 template <typename H, typename F>
 static inline void onemkl_cusolver_host_task(H &cgh, sycl::queue queue, F f) {
     (void)host_task_internal(cgh, queue, f);
+    queue.wait(); //temporary workaround for issue #216
 }
 
 } // namespace cusolver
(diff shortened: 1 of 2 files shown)


src/lapack/backends/cusolver/cusolver_task.hpp
```diff
@@ -43,6 +43,7 @@ static inline void host_task_internal(H &cgh, sycl::queue queue, F f) {
 template <typename H, typename F>
 static inline void onemkl_cusolver_host_task(H &cgh, sycl::queue queue, F f) {
     (void)host_task_internal(cgh, queue, f);
+    queue.wait(); //temporary workaround for issue #216
```
JackAKirk, line 46: Unfortunately this does not work because it is actually syncing the queue before the command group associated with this `onemkl_cusolver_host_task` is submitted: The result is that I find this does not solve #216

JackAKirk: Interestingly I find that replacing `queue.wait()` with a change to `host_task_internal`:

```
    cgh.interop_task([f, queue](sycl::interop_handler ih) {
        auto sc = CusolverScopedContextHandler(queue, ih);
        f(sc);
        cusolverStatus_t err;
        cudaStream_t currentStreamId;
        CUSOLVER_ERROR_FUNC(cusolverDnGetStream, err, handle, &currentStreamId);
        cuStreamSynchronize(currentStreamId);
    });
}
```

is also not solving #216. It seems that you have to either call cuStreamSynchronize from within the user facing `onemkl_cusolver_host_task` or call `queue.wait()` after the command group is submitted.

PR #144, opened 2021-11-20
Add rocBLAS backend

sbalint98:
# Description
As discussed in #99 with the newly added hipSYCL support oneMKL is able to target AMD GPUs. This PR adds the rocBLAS backend to the BLAS domain. The changes were based on the Cuda backend, and the set of supported BLAS functions matches the functionality of the Cuda backend.

Testing was carried out using rocm 4.5 and rocBLAS 4.5, and hipSYCL built from develop [cc794e3](https://github.com/illuhad/hipSYCL/tree/cc794e3291b5f365d6241689da16b5098f4a7a4a), on a system with AMD MI Instinct 100 GPUs. 

[rocm_tests_out.log](https://github.com/oneapi-src/oneMKL/files/7574723/rocm_tests_out.log)

- [x] Do all unit tests pass locally? Attach a log.
- [x] Have you formatted the code using clang-format?
- [x] Update Readme

PR #226, opened 2022-09-11
Add hipSYCL curand support

nilsfriess:
# Description

This PR adds support to use the cuRAND backend with hipSYCL. 
The approach is similar though simpler than #144: Since hipSYCL does not support `host_task` but instead implements the extension `hipSYCL_enqueue_custom_operation`, a wrapper function `onemkl_curand_host_task` is introduced that calls either `host_task` or `hipSYCL_enqueue_custom_operation` depending on which compiler is used.

Further, calls to `curandSetStream` before calling the cuRAND `generate`-functions were added. This is important since previously, the call to `wait_and_throw` after submitting the CUDA calls might not wait for the cuRAND random number generation to finish if it was started on a different stream. In fact, increasing the amount of random numbers generated in the unit tests (and thereby increasing the time spent generating numbers) made some tests fail (both when using hipSYCL and dpc++), since the result might be read before random number generation is finished. Now all tests pass also when increasing the number of random samples to be generated.

## Test logs
The tests were executed on Ubuntu 20.04 using CUDA 11.6.
[curand_dpcpp_test.log](https://github.com/oneapi-src/oneMKL/files/9543303/curand_dpcpp_test.log)
[curand_hipsycl_test.log](https://github.com/oneapi-src/oneMKL/files/9543304/curand_hipsycl_test.log)

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ Supported domains: BLAS, LAPACK, RNG
             <td align="center">NVIDIA GPU</td>
             <td align="center">NVIDIA cuRAND</td>
             <td align="center">Dynamic, Static</td>
-            <td align="center">LLVM*</td>
+            <td align="center">LLVM*, hipSYCL</td>
         </tr>
         <tr>
             <td align="center">AMD GPU</td>
--- a/cmake/FindcuRAND.cmake
+++ b/cmake/FindcuRAND.cmake
@@ -58,6 +58,8 @@
 
 find_package(CUDA 10.0 REQUIRED)
 get_filename_component(SYCL_BINARY_DIR ${CMAKE_CXX_COMPILER} DIRECTORY)
+
+if (NOT (ONEMKL_SYCL_IMPLEMENTATION STREQUAL "hipsycl"))
 # the OpenCL include file from cuda is opencl 1.1 and it is not compatible with DPC++
 # the OpenCL include headers 1.2 onward is required. This is used to bypass NVIDIA OpenCL headers
 find_path(OPENCL_INCLUDE_DIR CL/cl.h OpenCL/cl.h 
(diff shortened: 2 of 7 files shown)


aelizaro (approved):
Hi @nilsfriess, thank you for the PR, looks good to me! @mkrainiuk, could you, please, also take a look?

mkrainiuk (approved):
Thank you for the PR, it also looks good to me. Does it make sense to update also our documentation with RNG supported compilers?

nilsfriess:
I think there are multiple places where the documentation is not quite up to date with regard to hipSYCL support (e.g. [here](https://oneapi-src.github.io/oneMKL/selecting_a_compiler.html) and [here](https://oneapi-src.github.io/oneMKL/building_the_project.html) it is not mentioned that hipSYCL can also be used for the BLAS backend with CUDA). 
Should I make an extra PR for the documentation updates or just include it in this one?

mkrainiuk:
> I think there are multiple places where the documentation is not quite up to date with regard to hipSYCL support (e.g. [here](https://oneapi-src.github.io/oneMKL/selecting_a_compiler.html) and [here](https://oneapi-src.github.io/oneMKL/building_the_project.html) it is not mentioned that hipSYCL can also be used for the BLAS backend with CUDA). Should I make an extra PR for the documentation updates or just include it in this one?

Good catch, we need to update them too. I'm fine with both options, if you prefer extra PR we can merge this PR as is.

nilsfriess:
There weren't too many places where the documentation was not up to date, so I just added the changes to this PR.

mkrainiuk:
> There weren't too many places where the documentation was not up to date, so I just added the changes to this PR.

Thank you! Documentation looks good to me. Please let me know if there anything else you want to add or we can merge it now.

nilsfriess:
Thanks, you can merge it :)

PR #227, opened 2022-09-14
[BLAS] add interfaces to matrix copy/transposition routines

andrewtbarker:
# Description

This PR adds interfaces for batch strided matrix copy/transposition routines `omatcopy`, `imatcopy`, and `omatadd`.

Currently the non-batch routines are not fully document/implemented in the oneMKL product, when they are we will add interfaces to them as well. At that point we will also implement the cublas and rocblas backends via their `geam` interfaces where appropriate.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Logs attached, some failures expected from #204

## New interfaces

it was accepted? [#420](https://github.com/oneapi-src/oneAPI-spec/pull/420)
- [x] What version of oneAPI spec the interface is targeted? (1.2 provisional rev 1)

## New features

- [x] Have you added relevant tests? Added tests are "smoke tests" to test for compilation, linking, and execution, but do not test correctness because there is no reference code for the added functionality.

[test-cublas.txt](https://github.com/oneapi-src/oneMKL/files/9568422/test-cublas.txt)
[test-mklbackend.txt](https://github.com/oneapi-src/oneMKL/files/9568423/test-mklbackend.txt)


src/blas/backends/cublas/cublas_wrappers.cpp
```diff
@@ -634,6 +658,18 @@ extern "C" blas_function_table_t mkl_blas_table = {
     oneapi::mkl::blas::cublas::row_major::gemm_bias,
     oneapi::mkl::blas::cublas::row_major::gemm_bias,
     oneapi::mkl::blas::cublas::row_major::gemm_bias,
+    oneapi::mkl::blas::cublas::column_major::omatcopy_batch,
```
mmeterel: wrong major namespace in this section

src/blas/backends/rocblas/rocblas_wrappers.cpp
```diff
@@ -219,6 +219,18 @@ extern "C" blas_function_table_t mkl_blas_table = {
     oneapi::mkl::blas::rocblas::column_major::gemm_bias,
     oneapi::mkl::blas::rocblas::column_major::gemm_bias,
     oneapi::mkl::blas::rocblas::column_major::gemm_bias,
+    oneapi::mkl::blas::rocblas::row_major::omatcopy_batch,
```
mmeterel: wrong major namespace in this section

tests/unit_tests/blas/batch/imatcopy_batch_stride.cpp
```diff
@@ -0,0 +1,193 @@
...
+using namespace sycl;
+using std::vector;
+
+extern std::vector<sycl::device *> devices;
+
+namespace {
+
+template <typename fp>
+int test(device *dev, oneapi::mkl::layout layout, int64_t batch_size) {
+    // smoke test to make sure routine runs; there is no netlib reference
```
mmeterel: I think we need to write our own reference implementation and compare the results with that.

tests/unit_tests/blas/batch/imatcopy_batch_stride_usm.cpp
```diff
@@ -0,0 +1,214 @@
...
+using namespace sycl;
+using std::vector;
+
+extern std::vector<sycl::device *> devices;
+
+namespace {
+
+template <typename fp>
+int test(device *dev, oneapi::mkl::layout layout, int64_t batch_size) {
+    // smoke test to make sure routine runs; there is no netlib reference
```
mmeterel: I think we need to write our own reference implementation and compare the results with that.

mmeterel (commented):
Thanks for the PR! It looks good overall. I have two comments:
1. Could you please fix the layout namespaces in couple places?
2. We need to write our own reference implementation for new APIs, so that we can actually check correctness.

src/blas/backends/cublas/cublas_wrappers.cpp
andrewtbarker: Done, thanks for spotting this.

tests/unit_tests/blas/batch/imatcopy_batch_stride.cpp
andrewtbarker: Done for all the added routines, let me know what you think.

andrewtbarker:
Thanks for the review @mmeterel , I think I have addressed your comments.

tests/unit_tests/blas/batch/imatcopy_batch_stride.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+using namespace sycl;
+using std::vector;
+
+extern std::vector<sycl::device *> devices;
+
+namespace {
+
+template <typename fp>
+void imatcopy_ref(oneapi::mkl::layout layout, oneapi::mkl::transpose trans, int64_t m, int64_t n,
```
mmeterel: Can we move the reference implementations to this common place (`tests/unit_tests/blas/include/reference_blas_templates.hpp`), along with other reference implementation?

andrewtbarker: Done, good idea!

mmeterel (approved):
Looks good. Thanks!

tests/unit_tests/blas/batch/omatadd_batch_stride.cpp
```diff
@@ -0,0 +1,232 @@
...
+template <typename fp>
+int test(device *dev, oneapi::mkl::layout layout, int64_t batch_size) {
+    // Prepare data.
+    int64_t m, n;
+    int64_t lda, ldb, ldc;
+    oneapi::mkl::transpose transa, transb;
+    fp alpha, beta;
+    int64_t i, tmp;
+
+    batch_size = 1 + std::rand() % 20;
```
aaronjohnson, line 59: I am unsure this be the same random number every time.    

Do we want reproducibility of random batch size?

aaronjohnson: In general, if something in a test were to fail.   Is knowing the exact values useful for reproducing.

aaronjohnson: In this case, transpose APIs, there is probably not a concern.

aaronjohnson (commented):
A few comments but nothing blocking.

And, `#pragma once` may suffice for header guard.  But to date, both ways are used across the project.

aaronjohnson (approved):
And lastly, _precondition and _postcondition may not be necessary since they are stubs.  

Then again, to match the rest of pre-existing BLAS it makes sense to include them here.


No changes required.   Just some observations.   Thank you, @andrewtbarker

mkrainiuk (approved):

Issue #223, opened 2022-08-29
cuBLAS backend (and potentially other domains) fails with latest LLVM builds 


mmeterel:
# Summary
As first observed in #219 many tests in cuBLAS backend is failing with latest LLVM builds. 

# Version
I have tried LLVM commit: 66361038b63caaae566fc9648f5da50b74222b83 and got the below tests failing (showing only a few of them)

         1 - BLAS/RT/Nrm2TestSuite/Nrm2Tests.RealSinglePrecision/Column_Major_TITAN_RTX (Failed)
          3 - BLAS/RT/Nrm2TestSuite/Nrm2Tests.RealDoublePrecision/Column_Major_TITAN_RTX (Failed)
          7 - BLAS/RT/Nrm2TestSuite/Nrm2Tests.ComplexDoublePrecision/Column_Major_TITAN_RTX (Failed)
         17 - BLAS/RT/IamaxTestSuite/IamaxTests.RealSinglePrecision/Column_Major_TITAN_RTX (Failed)
         19 - BLAS/RT/IamaxTestSuite/IamaxTests.RealDoublePrecision/Column_Major_TITAN_RTX (Failed)
         23 - BLAS/RT/IamaxTestSuite/IamaxTests.ComplexDoublePrecision/Column_Major_TITAN_RTX (Failed)
         27 - BLAS/RT/DotuTestSuite/DotuTests.ComplexDoublePrecision/Column_Major_TITAN_RTX (Failed)
         35 - BLAS/RT/DotcTestSuite/DotcTests.ComplexSinglePrecision/Column_Major_TITAN_RTX (Failed)
         67 - BLAS/RT/AsumTestSuite/AsumTests.ComplexSinglePrecision/Column_Major_TITAN_RTX (Failed)
         81 - BLAS/RT/ScalTestSuite/ScalTests.RealSinglePrecision/Column_Major_TITAN_RTX (Failed)
         85 - BLAS/RT/ScalTestSuite/ScalTests.ComplexSinglePrecision/Column_Major_TITAN_RTX (Failed)
         87 - BLAS/RT/ScalTestSuite/ScalTests.ComplexDoublePrecision/Column_Major_TITAN_RTX (Failed)
        103 - BLAS/RT/RotgTestSuite/RotgTests.RealDoublePrecision/Column_Major_TITAN_RTX (Failed)
[...]

PR #228, opened 2022-09-14
[BLAS - cuBLAS] Fix stream synchronization

mmeterel:
# Description
With recent LLVM builds, multiple streams (instead of single CUstream) are enabled in each queue - which basically broke the in-order property of the queue. [see here](https://github.com/intel/llvm/commit/dd418459868a976cd2eeae367fea6b92795ea611)
As a result, we no longer had synchronization in cuBLAS streams. This PR is adding synchronization for cuBLAS backend. 
Similar fix was implemented for cuSolver in #215 

Fixes # (GitHub issue)
#223 

# Checklist

## All Submissions

[fix_cublas_sync.log](https://github.com/oneapi-src/oneMKL/files/9569839/fix_cublas_sync.log)

### DIFF
--- a/src/blas/backends/cublas/cublas_helper.hpp
+++ b/src/blas/backends/cublas/cublas_helper.hpp
@@ -189,7 +189,7 @@ public:
     CUBLAS_ERROR_FUNC(cublasGetStream, err, handle, &currentStreamId);    \
     cuStreamSynchronize(currentStreamId);
 
-#define CUBLAS_ERROR_FUNC_T(name, func, err, handle, ...)                \
+#define CUBLAS_ERROR_FUNC_T_SYNC(name, func, err, handle, ...)           \
     err = func(handle, __VA_ARGS__);                                     \
     if (err != CUBLAS_STATUS_SUCCESS) {                                  \
         throw cublas_error(std::string(name) + std::string(" : "), err); \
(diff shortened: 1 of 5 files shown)


mkrainiuk (approved):
Looks good to me, thank you!

src/blas/backends/cublas/cublas_helper.hpp
```diff
@@ -180,11 +180,23 @@ class cuda_error : virtual public std::runtime_error {
...
+#define CUBLAS_ERROR_FUNC_SYNC(name, err, handle, ...)                    \
+    err = name(handle, __VA_ARGS__);                                      \
+    if (err != CUBLAS_STATUS_SUCCESS) {                                   \
+        throw cublas_error(std::string(#name) + std::string(" : "), err); \
+    }                                                                     \
+    cudaStream_t currentStreamId;                                         \
+    CUBLAS_ERROR_FUNC(cublasGetStream, err, handle, &currentStreamId);    \
+    cuStreamSynchronize(currentStreamId);
+
+#define CUBLAS_ERROR_FUNC_T(name, func, err, handle, ...)                \
```
andrewtbarker: Not that important, but I suggest `CUBLAS_ERROR_FUNC_T_SYNC` here (since it's synchronizing), I think it's more consistent with what we had before and what we did in #215.

andrewtbarker (approved):
Thanks for the fix!

src/blas/backends/cublas/cublas_helper.hpp
mmeterel: Right, I thought about it and got lazy. I will update all the cublas backend files with new macro name.

mmeterel: macro name is updated with `_SYNC`

PR #235, opened 2022-10-10
Sync fft interfaces

anantsrivastava30:
# Description

Please include a summary of the change. Please also include relevant
motivation and context. See
[contribution guidelines](https://github.com/oneapi-src/oneMKL/blob/master/CONTRIBUTING.md)
for more details. If the change fixes an issue not documented in the project's
Github issue tracker, please document all steps necessary to reproduce it.

Fixes # (GitHub issue)

# Checklist

## All Submissions

## New interfaces

it was accepted? # (RFC)

## New features

## Bug fixes

      GitHub issue or in this PR)?


include/oneapi/mkl/dft/detail/descriptor_impl.hpp
```diff
@@ -0,0 +1,97 @@
...
+#include "oneapi/mkl/detail/export.hpp"
+#include "oneapi/mkl/detail/get_device_id.hpp"
+#include "oneapi/mkl/dft/types.hpp"
+
+namespace oneapi {
+namespace mkl {
+namespace dft {
+namespace detail {
+
+class descriptor_impl {
```
t4c1, line 22: Everything in this impl class is general (not backend specific). So, it does not need to be in descriptor_impl, but can be in the descriptor class. We only need pimpl for backend-specific parts (factorizations, twiddle tables etc) and these can only be initialized in the commit call, not in descriptor constructor.

Implementations will need to be backend specific (so one in every backend) and for now you can leave them empty. They will only be initialized in commit implementation for each backend. 

Maybe we dont even need pimpl, just a pointer that each backend can set to point to its internal data it needs to store in the descriptor (just thinking out loud).

include/oneapi/mkl/dft/detail/dft_loader.hpp
```diff
@@ -0,0 +1,32 @@
...
+namespace oneapi {
+namespace mkl {
+namespace dft {
+namespace detail {
+
+template <oneapi::mkl::dft::precision prec, oneapi::mkl::dft::domain dom>
+oneapi::mkl::dft::detail::descriptor_impl* create_descriptor(std::size_t length);
+
+template <oneapi::mkl::dft::precision prec, oneapi::mkl::dft::domain dom>
+oneapi::mkl::dft::detail::descriptor_impl* create_descriptor(std::vector<std::int64_t> dimensions);
```
t4c1: We won't need these after everything from descriptor_impl is moved to descriptor.

t4c1 (changes requested):

include/oneapi/mkl/dft/detail/descriptor_impl.hpp
anantsrivastava30: Hey Tadej, thank you for the review, I have a few options to move forward
1. impliment the constructor and the `set_value` +` get_value` in the `descriptor.hpp`, use pimpl only for commit
2. if we prefer to separate implimentation and declaration then I would have to explicitly specialize consturctors + pre-commit member functinos, which would bloat the implimentation code (e.g. `src/dft/backend/descriptor.cxx`)
3. could use `detail::create_descriptor` (pointer to class impl) for general implimentation, and detail::create_commit` (pointer to different commit class ?) along with `mklcpu::create_commit` (... `mklgpu::create_commit`), this would avoid some bloating as create_descriptor can be templated and passed on to the impl class.

for now I can work on option 1 as the prefered way, let me know your thoughts on this.

t4c1: Yeah, I like 1 the best as well. Other options sound like unnecessary complications.

anantsrivastava30: 710a4e157c8d3fbbfe2144708e4886066084c4c0, should enable you to work on gpu commit.

include/oneapi/mkl/dft/descriptor.hpp
```diff
@@ -27,21 +27,35 @@
...
 public:
     // Syntax for 1-dimensional DFT
-    descriptor(std::int64_t length);
+    descriptor(std::int64_t length)
+#ifdef ENABLE_MKLCPU_BACKEND
+        : pimpl_(mklcpu::create_descriptor(length)) {}
+#endif
+#ifdef ENABLE_MKLGPU_BACKEND
+        : pimpl_(mklgpu::create_descriptor(length)) {}
+#endif
```
t4c1: This will prevent us from compiling for multiple backends. `pimpl_` should be initialized in `commit`, where we know which backend is used (from the queue).

include/oneapi/mkl/dft/descriptor.hpp
```diff
@@ -27,21 +27,35 @@
...
+
     // Syntax for d-dimensional DFT
     descriptor(std::vector<std::int64_t> dimensions);
 
-    ~descriptor();
+    // ~descriptor();
```
t4c1: Why comment out destructor? I think this should be reverted.

include/oneapi/mkl/dft/detail/descriptor_impl.hpp
```diff
@@ -0,0 +1,45 @@
...
+    virtual descriptor_impl* copy_state() = 0;
+
+    virtual ~descriptor_impl() {}
+
+    sycl::queue& get_queue() {
+        return queue_;
+    }
+
+protected:
+    sycl::queue queue_;
```
t4c1, line 28: I think we don't need queue in the `descriptor_impl`, as it is part of the descriptor.

include/oneapi/mkl/dft/detail/dft_loader.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include "oneapi/mkl/dft/detail/descriptor_impl.hpp"
+
+namespace oneapi {
+namespace mkl {
+namespace dft {
+
+namespace mklcpu {
+
+ONEMKL_EXPORT oneapi::mkl::dft::detail::descriptor_impl* create_descriptor(std::size_t length);
```
t4c1: We do not need those anymore. In fact, this whole file can be removed.

t4c1 (changes requested):
I added multiple comments, but I think only the first one actually blocks any further implementations.

include/oneapi/mkl/dft/descriptor.hpp
t4c1: Also, destructor should free pimpl if commit was called before descriptor is destroyed.

anantsrivastava30: pimpl is casted as unique_ptr, it will free itself as it goes out of scope.

anantsrivastava30:
> I added multiple comments, but I think only the first one actually blocks any further implementations.

I think you reviewed old code, there have been a few commits after the one you reviewed.

include/oneapi/mkl/dft/descriptor.hpp
t4c1: Makes sense. This is fine as it is.

t4c1:
Right. Nothing in this should block us now, but the comment about not needing queue in `descriptor_impl` still stands.

KumudhaN:
@anantsrivastava30 Can you modify the PR to target the develop branch in your fork and then we should be able to work on top of that. 
Thanks!

anantsrivastava30:
> @anantsrivastava30 Can you modify the PR to target the develop branch in your fork and then we should be able to work on top of that. Thanks!

@KumudhaN thanks for pointing out opend a local PR! and merging it to local https://github.com/anantsrivastava30/oneMKL/pull/1

Rbiessy:
@anantsrivastava30 I think this can be closed right?

PR #236, opened 2022-10-10
[BLAS] Remove unused predicates (precondition / postcondition checks)

andrewtbarker:
# Description

The BLAS domain included a large amount of unnecessary boilerplate code to support precondition / postcondition checks for each routine. This PR removes all of this boilerplate but does not remove any actual active checks.

Adding checks is still possible, but we will no longer maintain thousands of lines of boilerplate placeholder code that is unlikely to ever be used.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Log: [test.txt](https://github.com/oneapi-src/oneMKL/files/9749509/test.txt), several tests fail as expected from #204


mkrainiuk (approved):
Thanks for PR!
The predicate stubs were added as potential feature that can allow users disabling internal checks in case if they want to skip unnecessary checks. E.g., Intel oneMKL provides extra direct call C/Fortran API for this purpose but for DPC++ API predicates can help with having this feature in the same API. If this feature is not yet required it makes sense to me to remove this code.

b-shi (approved):

vmalia (approved):

PR #238, opened 2022-10-13
Move `lapack_info_check` inside of `onemkl_cusolver_host_task`

JackAKirk:
# Description

In some cases cuSolver operations can return a successful error code while failing. The previous implementation of this check is done via SYCL and requires the CPU to wait until the cuSolver function completes. This is not great for performance as it prevents the user for issuing more work to the queue until it has received a response from oneMKL.

This PR moves the check inside the host task.

@AidanBeltonS I've updated the cusolver_batch.cpp cases too, can you check it is OK?


src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -571,7 +569,6 @@ inline sycl::event getrf_batch(const char *func_name, Func func, sycl::queue &qu
     // Allocate memory with 32-bit ints then copy over results
     std::uint64_t ipiv_size = stride_ipiv * batch_size;
     int *ipiv32 = (int *)malloc_device(sizeof(int) * ipiv_size, queue);
-    int *devInfo = (int *)malloc_device(sizeof(int) * batch_size, queue);
 
     auto done = queue.submit([&](sycl::handler &cgh) {
         int64_t num_events = dependencies.size();
```
AidanBeltonS: The dependencies loop can be simplified in the batch with `depends_on_events(cgh, dependencies);`

src/lapack/backends/cusolver/cusolver_batch.cpp
```diff
@@ -581,16 +578,17 @@ inline sycl::event getrf_batch(const char *func_name, Func func, sycl::queue &qu
         onemkl_cusolver_host_task(cgh, queue, [=](CusolverScopedContextHandler &sc) {
             auto handle = sc.get_handle(queue);
             auto a_ = reinterpret_cast<cuDataType *>(a);
-            auto devInfo_ = reinterpret_cast<int *>(devInfo);
             auto scratchpad_ = reinterpret_cast<cuDataType *>(scratchpad);
             auto ipiv_ = reinterpret_cast<int *>(ipiv32);
```
AidanBeltonS: You can also remove this unnecessary reinterpret cast as `ipiv32` is already an `int *`

src/lapack/backends/cusolver/cusolver_batch.cpp
JackAKirk: Nice one. I've made these changes.

AidanBeltonS:
LGTM!

JackAKirk:
@ericlars Could we get a review of this from someone? Thanks

PR #240, opened 2022-10-21
[doc] Fix stride_{a,b} descriptions in getrs_batch (LAPACK)

sknepper:
# Description

For getrs_batch functions with a stride_b input (getrs_batch strided for USM and BUFFER API, plus scratchpad_size), the stride_b description is missing, and the description of stride_a refers to the B matrices instead of the A matrices.


eischnei (approved):
This change appears to be in order.

sknepper:
LAPACK documentation was removed during renaming from oneMKL interfaces to oneMath. This PR is no longer relevant.

PR #247, opened 2022-11-02
[BLAS] fix linking issue for rotm() with oneMKL 2023.0

andrewtbarker:
# Description

This is a proposed fix for linking issues with the mklcpu backend seen when linking with upcoming oneMKL 2023.0 where the `rotm` routine has a different API in the newer version.

~The proposed fix works with the newer oneMKL but will not link with previous MKL versions, because the macro we are checking is not in fact defined. I am looking for a robust way to do this but I'm not sure it's easy.~

The proposed fix includes a relatively ugly hack to accommodate different versions of the MKL backend that have different function signatures for `rotm`.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Log: [test_2023.txt](https://github.com/oneapi-src/oneMKL/files/9924723/test_2023.txt)

### DIFF
--- a/cmake/FindMKL.cmake
+++ b/cmake/FindMKL.cmake
@@ -51,7 +51,9 @@ foreach(lib ${MKL_LIBRARIES})
   find_library(${lib}_file NAMES ${lib}
           HINTS $ENV{MKLROOT} ${MKL_ROOT}
           PATH_SUFFIXES lib/intel64)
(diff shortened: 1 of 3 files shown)


mkrainiuk (approved):
Looks good to me. Rotm tests on Windows also passed with Intel oneMKL 2022.2.

aelizaro (approved):

Issue #245, opened 2022-11-02
[cuBLAS] cuBLAS test sometimes segfaults with dpc++ and hipSYCL


nilsfriess:
# Summary
When trying to run the unit tests from the cuBLAS backend, one test case sometimes causes a segfault both when using hipSYCL and dpc++.

# Version
For the dpc++ tests, I used the current develop branch (b8ec0cf). For the tests with hipSYCL 7020ba92e72f27850d8812512a1e737c3e446c2e was used (the cuBLAS backend does not compile with hipSYCL when using the current oneMKL develop branch, the branch I used here has the changes to make it compile again; it is simply a library that oneMKL should be linked against when compiling with hipSYCL that is missing in one of the CMake files).

# Environment
* Hardware: NVIDIA GeForce GTX 1080
* CUDA 11.6
* Ubuntu 20.04
* dpc++: intel/llvm@8ca7d50456bde7f144eb5779bc6131bae841c1bb with CUDA support (see compile instructions below)
* hipSYCL: illuhad/hipSYCL@012e16d6d3d57330c176d7d536f657b0d8a9a197 compiled with CUDA backend enabled
* CMake Output:
    * [cmake_dpcpp.log](https://github.com/oneapi-src/oneMKL/files/9920662/cmake_dpcpp.log)
    * [cmake_hipSYCL.log](https://github.com/oneapi-src/oneMKL/files/9920663/cmake_hipSYCL.log)

# Steps to reproduce
Install dpc++ with CUDA support as described [here](https://intel.github.io/llvm-docs/GetStartedGuide.html), i.e., run (with the paths of course adapted)
```
CUDA_LIB_PATH=/usr/local/cuda-11.6/lib64/stubs/ CC=/usr/lib/llvm-14/bin/clang CXX=/usr/lib/llvm-14/bin/clang++ python ../buildbot/configure.py --cuda --cmake-opt="-DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-11.6"
CUDA_LIB_PATH=/usr/local/cuda-11.6/lib64/stubs/ CC=/usr/lib/llvm-14/bin/clang CXX=/usr/lib/llvm-14/bin/clang++ python ../buildbot/compile.py -j64
```
Then build oneMKL with the cuBLAS backend as described in the documentation, i.e., run 
```
[...]

PR #249, opened 2022-11-09
[BLAS, hipSYCL] Add CUDA_CUDA_LIBRARY to also link against libcuda.so

nilsfriess:
# Description

PR #228 added calls to `cuStreamSynchronize` in the cuBLAS backend. This function is located in `libcuda.so` but this library is not linked against when using hipSYCL to compile oneMKL. This PR adds `CUDA_CUDA_LIBRARY` in `FindcuBLAS.cmake` to fix this.

# Checklist

    The tests do not pass currently on my setup, see #245, but it seems very unlikely that this is related to the change in this PR.
    Does not apply here.


mkrainiuk (approved):
Thank you for the fix! The changes look good to me.

andrewtbarker (approved):
Looks good, thanks!

PR #251, opened 2022-11-18
[BLAS] Fix symbol conflicts between backends and reference

### LINKED ISSUE #204: BLAS: tests may not be executing the right backend

dnhsieh-intel:
# Description

When a backend respects CBLAS symbols, instead of comparing the backend against the reference libcblas library, the current testing structure would compare the reference against itself due to the conflicting symbols. This PR resolves this issue by loading reference functions at runtime into a local namespace.

Fixes #204

# Tests
Diagnostic messages were inserted in the Netlib `cblas_saxpy` and the oneMKL `axpy` unit test to check when the reference was called. More specifically:
- In [the Netlib cblas_saxpy.c](https://github.com/Reference-LAPACK/lapack/blob/master/CBLAS/src/cblas_saxpy.c)
  ``` c
  void cblas_saxpy( const CBLAS_INT N, const float alpha, const float *X,
                         const CBLAS_INT incX, float *Y, const CBLAS_INT incY)
  {
     printf("   Netlib cblas_saxpy called\n");
  #ifdef F77_INT
     F77_INT F77_N=N, F77_incX=incX, F77_incY=incY;
  #else
     #define F77_N N
     #define F77_incX incX
     #define F77_incY incY
  #endif
     F77_saxpy( &F77_N, &alpha, X, &F77_incX, Y, &F77_incY);
  }
  ```
- In [the oneMKL unit test axpy.cpp](https://github.com/oneapi-src/oneMKL/blob/develop/tests/unit_tests/blas/level1/axpy.cpp)
  ``` cpp
  template <typename fp>
  int test(device *dev, oneapi::mkl::layout layout, int N, int incx, int incy, fp alpha) {
      // Prepare data.
      vector<fp> x, y, y_ref;
  
      rand_vector(x, N, incx);
      rand_vector(y, N, incy);
      y_ref = y;
  
      // Call Reference AXPY.
      std::cout << "Call reference axpy" << std::endl;
      using fp_ref = typename ref_type_info<fp>::type;
      const int N_ref = N, incx_ref = incx, incy_ref = incy;
  
      ::axpy(&N_ref, (fp_ref *)&alpha, (fp_ref *)x.data(), &incx_ref, (fp_ref *)y_ref.data(),
             &incy_ref);
  
      // Call DPC++ AXPY.
      std::cout << "Call DPC++ axpy" << std::endl;
      ......
  ```

We use the _ct test as an example in the following. The results of the _rt test are the same.

Since the conflicts of symbols between the MKL backend and the reference have been fixed in PR #210, the reference function was called appropriately both with and without this PR:
```
$ ./bin/test_main_blas_ct --gtest_filter=AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major*CPU*
Note: Google Test filter = AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major*CPU*
[==========] Running 1 test from 1 test suite.
[----------] 1 test from AxpyTestSuite/AxpyTests
[ RUN      ] AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz
Call reference axpy
   Netlib cblas_saxpy called
Call DPC++ axpy
Call reference axpy
   Netlib cblas_saxpy called
Call DPC++ axpy
Call reference axpy
   Netlib cblas_saxpy called
Call DPC++ axpy
[       OK ] AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz (125 ms)
[----------] 1 test from AxpyTestSuite/AxpyTests (125 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (125 ms total)
[  PASSED  ] 1 test.
```

To simulate the situation of conflicting symbols, [the commit 634d7d2](https://github.com/oneapi-src/oneMKL/tree/634d7d25c7bed523d1157cb46177c4d7494862d2) right before PR #210 was used. Based on the commit, the diagnostic messages were:
```
$ ./bin/test_main_blas_ct --gtest_filter=AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major*CPU*
Note: Google Test filter = AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major*CPU*
[==========] Running 1 test from 1 test suite.
[----------] 1 test from AxpyTestSuite/AxpyTests
[ RUN      ] AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz
Call reference axpy
   Netlib cblas_saxpy called
Call DPC++ axpy
   Netlib cblas_saxpy called
Call reference axpy
   Netlib cblas_saxpy called
Call DPC++ axpy
   Netlib cblas_saxpy called
Call reference axpy
   Netlib cblas_saxpy called
Call DPC++ axpy
   Netlib cblas_saxpy called
[       OK ] AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz (104 ms)
[----------] 1 test from AxpyTestSuite/AxpyTests (104 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (104 ms total)
[  PASSED  ] 1 test.
```

We then applied the changes in this PR to the commit. We can see from the messages that the changes fixed the issue.
```
$ ./bin/test_main_blas_ct --gtest_filter=AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major*CPU*
Note: Google Test filter = AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major*CPU*
[==========] Running 1 test from 1 test suite.
[----------] 1 test from AxpyTestSuite/AxpyTests
[ RUN      ] AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz
Call reference axpy
   Netlib cblas_saxpy called
Call DPC++ axpy
Call reference axpy
   Netlib cblas_saxpy called
Call DPC++ axpy
Call reference axpy
   Netlib cblas_saxpy called
Call DPC++ axpy
[       OK ] AxpyTestSuite/AxpyTests.RealSinglePrecision/Column_Major_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz (116 ms)
[----------] 1 test from AxpyTestSuite/AxpyTests (116 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (117 ms total)
[  PASSED  ] 1 test.
```

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Attach a log: [blas_lnx_log.txt](https://github.com/oneapi-src/oneMKL/files/10036178/blas_lnx_log.txt)


tests/unit_tests/blas/include/reference_blas_templates.hpp
```diff
@@ -39,74 +39,2377 @@
...
+    int izamax_res = 0;
+    if (cblas_library() != NULL) {
+        if (cblas_izamax_p == NULL)
+            cblas_izamax_p = (int (*)(const int n, const void *x, const int incx))GET_FUNC(
+                h_libcblas, "cblas_izamax");
+        if (cblas_izamax_p != NULL)
+            izamax_res = cblas_izamax_p(n, x, incx);
+    }
+    return izamax_res;
+}
```
andrewtbarker: This file is already huge, I suggest putting everything above this point in a separate file (maybe named `reference_blas_wrappers.hpp` or something similar) and including it in `reference_blas_templates.hpp`. I actually wish there was a way to avoid so much wrapping and indirection but probably this is the best we can do.

andrewtbarker (commented):
Thanks for this pull request and the thorough investigation! I have a question about this:

> To simulate the situation of conflicting symbols, [the commit 634d7d2](https://github.com/oneapi-src/oneMKL/tree/634d7d25c7bed523d1157cb46177c4d7494862d2) right before PR https://github.com/oneapi-src/oneMKL/pull/210 was used.

Why this particular commit? The problem of using reference blas for both baseline and test persists even after #210.

tests/unit_tests/CMakeLists.txt
```diff
@@ -152,15 +152,15 @@ foreach(domain ${TARGET_DOMAINS})
   # Find individual tests within executable
     gtest_discover_tests(test_main_${domain}_rt
       PROPERTIES BUILD_RPATH ${CMAKE_BINARY_DIR}/lib
-      PROPERTIES ENVIRONMENT LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/lib:$ENV{LD_LIBRARY_PATH}
+      PROPERTIES ENVIRONMENT LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/lib:${CBLAS_LIB_DIR}:$ENV{LD_LIBRARY_PATH}
```
mkrainiuk: I don't think we need this library for any other domains except BLAS, could you please add a check that `domain` is BLAS and only in this case add the extra library?

tests/unit_tests/CMakeLists.txt
```diff
@@ -152,15 +152,15 @@ foreach(domain ${TARGET_DOMAINS})
...
       PROPERTIES TEST_PREFIX ${DOMAIN_PREFIX}/RT/
       DISCOVERY_TIMEOUT 30
     )
   endif()
 
   gtest_discover_tests(test_main_${domain}_ct
     PROPERTIES BUILD_RPATH ${CMAKE_BINARY_DIR}/lib
-    PROPERTIES ENVIRONMENT LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/lib:$ENV{LD_LIBRARY_PATH}
+    PROPERTIES ENVIRONMENT LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/lib:${CBLAS_LIB_DIR}:$ENV{LD_LIBRARY_PATH}
```
mkrainiuk: Please add `domain` check here if possible

tests/unit_tests/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ set(blas_TEST_LIST
         blas_batch
         blas_extensions)
 
-set(blas_TEST_LINK ${CBLAS_LINK})
+set(blas_TEST_LINK "")
```
mkrainiuk, line 47: Do we still need this line?

tests/unit_tests/blas/include/reference_blas_templates.hpp
```diff
@@ -39,74 +39,2377 @@
...
 
 extern "C" {
-static LIB_TYPE h = NULL;
-static void (*csrot_p)(const int *N, void *X, const int *incX, void *Y, const int *incY,
+static LIB_TYPE h_libblas = NULL;
+static LIB_TYPE blas_library() {
+    if (h_libblas == NULL) {
+        h_libblas = GET_LIB_HANDLE(REF_BLAS_LIBNAME);
+        if (h_libblas == NULL) {
+            throw oneapi::mkl::library_not_found("BLAS", "blas_library()",
```
mkrainiuk: Does it make sense to print REF_BLAS_LIBNAME here to provide more information about what exact library we tried to load?

mkrainiuk (commented):
Thanks for the PR! I have several questions to the changes.

tests/unit_tests/blas/include/reference_blas_templates.hpp
dnhsieh-intel: > The problem of using reference blas for both baseline and test persists even after https://github.com/oneapi-src/oneMKL/pull/210.

That's weird. I didn't see the problem in the develop using either gdb or output messages. Could you share what you observed please?

andrewtbarker: This open source product calls oneMKL CPU backend via DPC++ interfaces, which then call cblas symbols. I recall that those symbols were resolving to netlib even after #210. This is what I meant with [this comment](https://github.com/oneapi-src/oneMKL/issues/204#issuecomment-1230806775). But you have investigated this more closely than I did, I might be wrong.

dnhsieh-intel: I think we might have dodged the conflict. After PR #210, it looks to me that it is the _64 APIs that are called in the mklcpu backend, e.g., `cblas_saxpy_64`.

The wrappers were moved to `reference_blas_wrappers.hpp`. (commit [f17ce6c](https://github.com/oneapi-src/oneMKL/pull/251/commits/f17ce6c48dc27c299a4a3f8dc4842e8d2d5a604b))

tests/unit_tests/CMakeLists.txt
dnhsieh-intel: Good point! Addressed in the commit [f762e6f](https://github.com/oneapi-src/oneMKL/pull/251/commits/f762e6f871af3eb77133374852d1b0ccf37f97ed).

tests/unit_tests/CMakeLists.txt
dnhsieh-intel: Addressed in the commit [f762e6f](https://github.com/oneapi-src/oneMKL/pull/251/commits/f762e6f871af3eb77133374852d1b0ccf37f97ed).

tests/unit_tests/CMakeLists.txt
dnhsieh-intel: No, we don't, but it would improve readability as we don't need to worry if `blas_TEST_LINK` is defined somewhere else. In addition, it is consistent with the `rng` domain:
https://github.com/oneapi-src/oneMKL/blob/15946f83fc25ba130f73a65bcf3174639fb8c1a0/tests/unit_tests/CMakeLists.txt#L55-L60

tests/unit_tests/blas/include/reference_blas_templates.hpp
dnhsieh-intel: Addressed in the commit [f17ce6c](https://github.com/oneapi-src/oneMKL/pull/251/commits/f17ce6c48dc27c299a4a3f8dc4842e8d2d5a604b).

An example of the error message:
```
C++ exception with description "oneMKL: BLAS/cblas_library(): library not found: failed to load CBLAS library /nfs/site/proj/mkl/mirror/NN/tools/netlib/3.10.0_5d4180c/blas/lnx/lib/libcblas.so" thrown in the test body.
```

dnhsieh-intel:
Log of the commit [f17ce6c](https://github.com/oneapi-src/oneMKL/pull/251/commits/f17ce6c48dc27c299a4a3f8dc4842e8d2d5a604b): [blas_lnx_log_f17ce6c.txt](https://github.com/oneapi-src/oneMKL/files/10052739/blas_lnx_log_f17ce6c.txt)
100% tests passed, 0 tests failed out of 3530

tests/unit_tests/blas/include/reference_blas_templates.hpp
andrewtbarker: Oh, that makes sense, the _64 wrappers were a separate change. Thanks!

andrewtbarker (approved):

tests/unit_tests/CMakeLists.txt
mkrainiuk: Thank you! I have one more question regarding Windows build, we define LD_LIBRARY_PATH here for Netlib libraires but we don't define PATH on Windows, does it make sense to update PATH instead of LD_LIBRARY_PATH in case of Windows?

mkrainiuk (approved):
The changes look good to me. Thank you!

tests/unit_tests/CMakeLists.txt
dnhsieh-intel: I thought this is straightforward. I was wrong...

On Windows, if we try to set PATH, which is a semicolon-separated list, in `gtest_discover_tests`, for example:
``` cmake
gtest_discover_tests(test_main_${domain}_ct PROPERTIES ENVIRONMENT "PATH=C:\path1;C:\path2;C:\path3")
```
After CMake's internal processing, it turns out that the actual setting will only be `PATH=C:\path1`. This seems to be [a known CMake issue](https://gitlab.kitware.com/cmake/cmake/-/issues/21453).

Do you think we should open an issue to investigate a workaround? Right now, because users have to include Netlib in PATH to build on Windows, it's fine if they run tests directly after building. But they would encounter the "failed to load CBLAS library" error if they run tests in a separate session.

mkrainiuk: I think it's worth adding a note to our README about Netlib required in PATH on Windows. And we also can investigate potential W/A as separated issue. I don't think we need to mix the current issue and the Windows one in this PR.

dnhsieh-intel: Will be tracked in issue #252.

PR #260, opened 2022-12-29
[mklcpu][mklgpu] add OpenCL to required dependencies for Intel oneMKL SYCL library

### LINKED ISSUE #254: oneMKL CPU backend does not build with LLVM compiler
# Summary

The documentation claims the oneMKL CPU backend will work with LLVM compiler, but the build fails because `-lOpenCL` is missing.

# Version

Happens for git hash `f4866ab`.

# Environment

- CPU is Intel Core i9-7920X
- Backend library version is oneMKL 2022.2.
- Ubuntu 18.04.6 LTS
- LLVM (Intel project with LLVM technology) 20220319
- CMake build log attached: [build.txt](https://github.com/oneapi-src/oneMKL/files/10177883/build.txt)

# Steps to reproduce

Follow build instructions, using LLVM as compiler and setting `-DENABLE_MKLCPU_BACKEND=ON`

# Observed behavior

Build fails with the following error:
```
/usr/bin/ld: cannot find -lOpenCL
clang-15: error: linker command failed with exit code 1 (use -v to see invocation)
bin/blas/backends/mklcpu/CMakeFiles/onemkl_blas_mklcpu.dir/build.make:93: recipe for target 'lib/libonemkl_blas_mklcpu.so.0' failed
```
See attached build log for full log.

# Expected behavior

Build succeeds.

mkrainiuk:
# Description

Fixes #254 
The PR adds OpenCL library check to FindMKL.cmake, so that missed OpenCL can be caught at configuration step.

# Checklist

## All Submissions

    ```
    cmake .. -DENABLE_MKLGPU_BACKEND=False
    -- CMAKE_BUILD_TYPE: None, set to Release by default
    -- C compiler: icx was not found in PATH, using clang instead
    -- The CXX compiler identification is Clang 16.0.0
    ...
    -- SYCL_DEVICE_FILTER will be set to the following value(s): [cpu] for run-time dispatching examples
    -- Found OpenCL: /usr/lib/x86_64-linux-gnu/libOpenCL.so
    -- Configuring done
    -- Generating done
    ````


cmake/FindMKL.cmake
```diff
@@ -94,6 +94,7 @@ if (ENABLE_MKLCPU_BACKEND OR ENABLE_MKLGPU_BACKEND)
     list(APPEND MKL_LINK_C ${TBB_LINK})
   endif()
   if(USE_DPCPP_API)
+    find_package(OpenCL REQUIRED)
```
andrewtbarker, line 97: Does this also set the library name in some variable, so that line 81 and 85 above aren't necessary?

andrewtbarker (approved):
Looks good to me, thanks!

akabalov (approved):

cmake/FindMKL.cmake
mkrainiuk: It defines OpenCL::OpenCL target, but I'm not sure if it includes only -lOpenCL or any other extra flags and libraires that we don't want to include.
As the next step of improvement I think we need to integrate MKLConfig.cmake from Intel oneMKL instead of this local FindMKL.cmake.

andrewtbarker: Thanks, the approach in this PR is fine, I was just curious.

PR #275, opened 2023-02-02
[DFT] Multi-Dimension Many-Batch DFT tests

FMarno:
# Description

This builds on #261 to introduce tests for multi-dimensional DFTs and batched DFTs.

# Checklist

## All Submissions

### DIFF
--- a/tests/unit_tests/dft/include/compute_inplace.hpp
+++ b/tests/unit_tests/dft/include/compute_inplace.hpp
@@ -103,7 +103,7 @@ int DFT_Test<precision, domain>::test_in_place_buffer() {
             : size_total;
     const std::int64_t container_size_per_transform = container_size_total / batches;
     const std::int64_t backward_elements = domain == oneapi::mkl::dft::domain::REAL
-                                               ? container_size_per_transform / 2 + 1
+                                               ? container_size_per_transform / 2
                                                : container_size_per_transform;
 
     std::vector<FwdInputType> inout_host(container_size_total, 0);
@@ -212,7 +212,7 @@ int DFT_Test<precision, domain>::test_in_place_USM() {
             : size_total;
     const int64_t container_size_per_transform = container_size_total / batches;
     const std::int64_t backward_elements = domain == oneapi::mkl::dft::domain::REAL
-                                               ? container_size_per_transform / 2 + 1
+                                               ? container_size_per_transform / 2
                                                : container_size_per_transform;
 
     auto ua_input = usm_allocator_t<FwdInputType>(cxt, *dev);
--- a/tests/unit_tests/dft/include/compute_inplace.hpp
+++ b/tests/unit_tests/dft/include/compute_inplace.hpp
@@ -21,6 +21,67 @@
 #define ONEMKL_COMPUTE_INPLACE_HPP
 
 #include "compute_tester.hpp"
+#include <oneapi/mkl/exceptions.hpp>
+
+std::vector<std::int64_t> get_real_strides(const std::vector<std::int64_t>& sizes) {
+    switch (sizes.size()) {
+        case 1: return { 0, 1 };
+        case 2: return { 0, 2 * (sizes[1] / 2 + 1), 1 };
+        case 3: return { 0, 2 * sizes[1] * (sizes[2] / 2 + 1), 2 * (sizes[2] / 2 + 1), 1 };
+        default:
+            throw oneapi::mkl::unimplemented(
+                "compute_inplace", __FUNCTION__,
+                "not implemented for " + std::to_string(sizes.size()) + " dimensions");
+            return {};
+    }
+}
+
+std::vector<std::int64_t> get_complex_strides(const std::vector<std::int64_t>& sizes) {
+    switch (sizes.size()) {
+        case 1: return { 0, 1 };
+        case 2: return { 0, sizes[1] / 2 + 1, 1 };
+        case 3: return { 0, sizes[1] * (sizes[2] / 2 + 1), (sizes[2] / 2 + 1), 1 };
+        default:
+            throw oneapi::mkl::unimplemented(
+                "compute_inplace", __FUNCTION__,
+                "not implemented for " + std::to_string(sizes.size()) + " dimensions");
+            return {};
+    }
+}
+
+template <typename fp>
+std::vector<fp> get_conjugate_even_ref(const std::vector<std::int64_t>& sizes, std::int64_t batches,
+                                       std::vector<std::complex<fp>> output_ref) {
+    const std::int64_t conjugate_even_last_dim = sizes.back() + 2;
+    const std::int64_t rows =
+        std::accumulate(sizes.begin(), sizes.end() - 1, batches, std::multiplies<>{});
+    std::vector<fp> conjugate_even_ref(rows * conjugate_even_last_dim);
+    for (int j = 0; j < rows; j++) {
+        for (int i = 0; i < conjugate_even_last_dim; i += 2) {
+            conjugate_even_ref[j * conjugate_even_last_dim + i] =
+                output_ref[j * sizes.back() + i / 2].real();
+            conjugate_even_ref[j * conjugate_even_last_dim + i + 1] =
+                output_ref[j * sizes.back() + i / 2].imag();
+        }
+    }
+    return conjugate_even_ref;
+}
+
+template <typename T, typename al>
+void copy_strided(const std::vector<std::int64_t>& sizes, std::int64_t batches,
+                  const std::vector<T>& input, std::vector<T, al>& output) {
+    const std::int64_t rows =
+        std::accumulate(sizes.begin(), sizes.end() - 1, batches, std::multiplies<>{});
+
+    auto in_iter = input.cbegin();
+    auto out_iter = output.begin();
+    const auto row_len = sizes.back();
+    while (in_iter < input.cend()) {
+        std::copy(in_iter, in_iter + row_len, out_iter);
+        in_iter += row_len;
+        out_iter += row_len + 2;
+    }
+}
 
 template <oneapi::mkl::dft::precision precision, oneapi::mkl::dft::domain domain>
 int DFT_Test<precision, domain>::test_in_place_buffer() {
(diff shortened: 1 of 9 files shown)


tests/unit_tests/dft/include/reference_dft.hpp
```diff
@@ -23,44 +23,106 @@
...
+        reference_forward_dft_impl(in, out, sizes[0], 1);
+    }
+};
+
+template <typename TypeIn, typename TypeOut>
+struct reference<TypeIn, TypeOut, 2> {
+    static void forward_dft(const std::vector<std::int64_t> &sizes, TypeIn const *const in,
+                            TypeOut *out) {
+        const auto elements = std::accumulate(sizes.begin(), sizes.end(), 1, std::multiplies<>{});
+        std::vector<std::complex<double>> tmp(elements);
```
Rbiessy: Shouldn't we use `long double` for the temporary buffers here and below?

tests/unit_tests/dft/source/compute_tests.cpp
```diff
@@ -39,244 +39,74 @@ extern std::vector<sycl::device *> devices;
...
+    INSTANTIATE_TEST_DIMENSIONS_PRECISION_DOMAIN(out_of_place, , STORAGE)       \
+    INSTANTIATE_TEST_DIMENSIONS_PRECISION_DOMAIN(out_of_place, real_real_, STORAGE)
+
+INSTANTIATE_TEST_DIMENSIONS_PRECISION_DOMAIN_PLACE_LAYOUT(buffer)
+INSTANTIATE_TEST_DIMENSIONS_PRECISION_DOMAIN_PLACE_LAYOUT(USM)
+
+using shape = std::vector<int64_t>;
+using i64 = std::int64_t;
+// Parameter format - { shape of transform, number of transforms }
+// due to how we are testing, the last dimension must be even
```
Rbiessy: That's not ideal, what is the issue exactly with odd sizes?

tests/unit_tests/dft/include/reference_dft.hpp
FMarno: addressed in https://github.com/oneapi-src/oneMKL/pull/275/commits/41f72310ccc0e6e81661d12cf658d0b8f7c3243a

tests/unit_tests/dft/source/compute_tests.cpp
FMarno: addressed in https://github.com/oneapi-src/oneMKL/pull/275/commits/41f72310ccc0e6e81661d12cf658d0b8f7c3243a

Rbiessy (approved):
Very nice, LGTM

Rbiessy:
Tests log: [log_mklgpu.txt](https://github.com/oneapi-src/oneMKL/files/10853003/log_mklgpu.txt)

FMarno:
When testing #284 with these new tests I found an off by one error. This has been updated in https://github.com/oneapi-src/oneMKL/pull/275/commits/348867ba133cb964d5bb9ad9a1fbb79ab8f871a9

Rbiessy:
I can confirm the tests are still passing: [log_mklgpu.txt](https://github.com/oneapi-src/oneMKL/files/10871095/log_mklgpu.txt)

Rbiessy (approved):

Rbiessy:
test

tests/unit_tests/dft/include/reference_dft.hpp
```diff
@@ -23,44 +23,106 @@
...
 #include <vector>
+#include <numeric>
 
+#include <oneapi/mkl/exceptions.hpp>
 #include "test_common.hpp"
 
+namespace detail {
+using ref_t = long double; /* Do the calculations using long double */
+template <typename TypeIn, typename TypeOut>
+void reference_forward_dft_impl(TypeIn *in, TypeOut *out, size_t N, size_t stride) {
```
raphael-egan:
```suggestion
void reference_forward_dft_impl(const TypeIn *in, TypeOut *out, size_t N, size_t stride) {
```

tests/unit_tests/dft/include/reference_dft.hpp
```diff
@@ -23,44 +23,106 @@
...
+        out[k * stride] = static_cast<TypeOut>(out_temp);
+    }
+}
+
+template <typename TypeIn, typename TypeOut, int dims>
+struct reference {};
+
+template <typename TypeIn, typename TypeOut>
+struct reference<TypeIn, TypeOut, 1> {
+    static void forward_dft(const std::vector<std::int64_t> &sizes, TypeIn const *const in,
```
raphael-egan:
```suggestion
    static void forward_dft(const std::vector<std::int64_t> &sizes, const TypeIn *in,
```

tests/unit_tests/dft/include/reference_dft.hpp
```diff
@@ -23,44 +23,106 @@
...
+                            TypeOut *out) {
+        reference_forward_dft_impl(in, out, sizes[0], 1);
+    }
+};
+
+template <typename TypeIn, typename TypeOut>
+struct reference<TypeIn, TypeOut, 2> {
+    static void forward_dft(const std::vector<std::int64_t> &sizes, TypeIn const *const in,
```
raphael-egan:
```suggestion
    static void forward_dft(const std::vector<std::int64_t> &sizes, const TypeIn *in,
```

tests/unit_tests/dft/include/reference_dft.hpp
```diff
@@ -23,44 +23,106 @@
...
+        }
+        for (size_t i = 0; i < sizes[1]; i++) {
+            reference_forward_dft_impl(tmp.data() + i, out + i, sizes[0], sizes[1]);
+        }
+    }
+};
+
+template <typename TypeIn, typename TypeOut>
+struct reference<TypeIn, TypeOut, 3> {
+    static void forward_dft(const std::vector<std::int64_t> &sizes, TypeIn const *const in,
```
raphael-egan:
```suggestion
    static void forward_dft(const std::vector<std::int64_t> &sizes, const TypeIn *in,
```

tests/unit_tests/dft/include/reference_dft.hpp
```diff
@@ -23,44 +23,106 @@
...
+                            TypeOut *out) {
+        const auto elements = std::accumulate(sizes.begin(), sizes.end(), 1, std::multiplies<>{});
+        std::vector<std::complex<ref_t>> tmp1(elements);
+        std::vector<std::complex<ref_t>> tmp2(elements);
```
raphael-egan, line 83: maybe we can use `out` instead of `tmp1` to reduce the memory foot print?

tests/unit_tests/dft/include/reference_dft.hpp
```diff
@@ -23,44 +23,106 @@
...
+            for (size_t i = 0; i < sizes[2]; i++) {
+                reference_forward_dft_impl(tmp1.data() + i + j, tmp2.data() + i + j, sizes[1],
+                                           sizes[2]);
+            }
+        }
+        for (size_t i = 0; i < sizes[1] * sizes[2]; i++) {
+            reference_forward_dft_impl(tmp2.data() + i, out + i, sizes[0], sizes[1] * sizes[2]);
+        }
+    }
+};
```
raphael-egan, line 97: It looks like there is some (possibly) avoidable duplication between the implementation. Do you think we could address that? (e.g. generalizing over the last template argument with nested loops)

tests/unit_tests/dft/include/reference_dft.hpp
```diff
@@ -23,44 +23,106 @@
...
+
+template <typename TypeIn, typename TypeOut>
+struct reference<TypeIn, TypeOut, 3> {
+    static void forward_dft(const std::vector<std::int64_t> &sizes, TypeIn const *const in,
+                            TypeOut *out) {
+        const auto elements = std::accumulate(sizes.begin(), sizes.end(), 1, std::multiplies<>{});
+        std::vector<std::complex<ref_t>> tmp1(elements);
+        std::vector<std::complex<ref_t>> tmp2(elements);
+        for (size_t i = 0; i < sizes[0] * sizes[1] * sizes[2]; i += sizes[2]) {
```
raphael-egan: Is there a reason for not using `elements` here, beyond the possible type conflict with `i`?

tests/unit_tests/dft/include/test_common.hpp
```diff
@@ -150,17 +150,38 @@ void commit_descriptor(oneapi::mkl::dft::descriptor<precision, domain> &descript
...
+
+        assert(sizes.size() > 0);
+        info_name.append("sizes_");
+
+        // intersperse dimensions with "x"
+        std::for_each(sizes.begin(), sizes.end() - 1,
+                      [&info_name](auto s) { info_name.append(std::to_string(s)).append("x"); });
+        info_name.append(std::to_string(sizes.back()));
+
+        info_name.append("_batches_").append(std::to_string(batches));
```
raphael-egan, line 186: Are we exporting info about real/complex, in- or out-of-place, forward/backward? FWIW, our own convention is to use
`{r,c}{f,b}{i,o}(Ax)(Bx)C*M`
for row-batched problems, which captures those pieces of information (lmk if something is unclear in that labeling).
(Strides and distances are not the most relevant for now but we may want to export those too, in the end)

I do like the device info/name being added!

tests/unit_tests/dft/include/compute_out_of_place.hpp
```diff
@@ -22,32 +22,48 @@
 
 #include "compute_tester.hpp"
 
+template <oneapi::mkl::dft::domain domain>
+std::int64_t get_row_size(const std::vector<std::int64_t> &sizes) noexcept {
```
raphael-egan:
```suggestion
std::int64_t get_backward_row_size(const std::vector<std::int64_t> &sizes) noexcept {
```

tests/unit_tests/dft/include/compute_out_of_place.hpp
```diff
@@ -22,32 +22,48 @@
...
+    if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
+        return sizes.back() / 2 + 1;
+    }
+    else {
+        return sizes.back();
+    }
+}
+
 /* Note: There is no implementation for Domain Real */
```
raphael-egan: Is that comment still valid? If yes, could you give more details? It should be implemented, in my understanding.

tests/unit_tests/dft/include/compute_out_of_place.hpp
```diff
@@ -22,32 +22,48 @@
...
 template <oneapi::mkl::dft::precision precision, oneapi::mkl::dft::domain domain>
 int DFT_Test<precision, domain>::test_out_of_place_buffer() {
     if (!init(MemoryAccessModel::buffer)) {
         return test_skipped;
     }
 
-    const size_t bwd_size = domain == oneapi::mkl::dft::domain::REAL ? (size / 2) + 1 : size;
+    const auto backward_distance = forward_elements;
```
raphael-egan:
```suggestion
    const auto backward_distance = std::accumulate(sizes.begin(), sizes.end() - 1, 1, std::multiplies<>{})*get_row_size(const std::vector<std::int64_t> &sizes);
```
Not sure that the current implementation is what is intended for real transforms otherwise. Multi-dimensional real transforms on GPU would require that for closed-source MKL.
[same comment(s) for USM tests]

tests/unit_tests/dft/include/compute_inplace.hpp
```diff
@@ -21,71 +21,182 @@
...
+            conjugate_even_ref[j * conjugate_even_last_dim + i + 1] =
+                output_ref[j * sizes.back() + i / 2].imag();
+        }
+    }
+    return conjugate_even_ref;
+}
+
+template <typename T, typename al>
+void copy_strided(const std::vector<std::int64_t>& sizes, std::int64_t batches,
+                  const std::vector<T>& input, std::vector<T, al>& output) {
```
raphael-egan: `std::int64_t batches` argument is not used.

tests/unit_tests/dft/include/compute_out_of_place.hpp
```diff
@@ -60,9 +76,18 @@ int DFT_Test<precision, domain>::test_out_of_place_buffer() {
 
         {
             auto acc_bwd = bwd_buf.template get_host_access();
-            EXPECT_TRUE(check_equal_vector(acc_bwd.get_pointer(), out_host_ref.data(),
-                                           bwd_data.size(), abs_error_margin, rel_error_margin,
-                                           std::cout));
+            auto bwd_ptr = acc_bwd.get_pointer();
+            auto ref_iter = out_host_ref.begin();
+            const auto ref_row_stride = sizes.back();
+            const auto backward_row_stride = sizes.back();
```
raphael-egan:
```suggestion
            const auto backward_row_stride = get_row_size<domain>(sizes);
```
(if opting for the other suggested changes to enabled 2D/3D batched r2c, I believe this is required)

tests/unit_tests/dft/include/test_common.hpp
FMarno: So one of the tests will have a name like "ComputeTestSuite/ComputeTests_in_place.COMPLEX_SINGLE_in_place_buffer/sizes_4x3_batches_27_Intel_R__HD_Graphics_530__0x1912_". As more tests are added I can see the name getting a little bit long but it is at least clear.
Is this convention your mentioning documented somewhere? Maybe we can add it to the readme if we use it.
I would guess it stands for:
* {r,c} real/complex forward domain
* {f,b} forward/backward
* {i,o} in-place/out-of-place
* (Ax) not sure
* (Bx) batches?
* C*M not sure

What about precision?

tests/unit_tests/dft/include/compute_out_of_place.hpp
FMarno: The test appears to be running so I would say it is out of date
```
[ RUN      ] ComputeTestSuite/ComputeTests_out_of_place.REAL_SINGLE_out_of_place_buffer/sizes_8_batches_1_Intel_R__HD_Graphics_530__0x1912_
[       OK ] ComputeTestSuite/ComputeTests_out_of_place.REAL_SINGLE_out_of_place_buffer/sizes_8_batches_1_Intel_R__HD_Graphics_530__0x1912_ (264 ms)
```

raphael-egan: Is that log output alone a confirmation that the actual test was successfully run with results validated and not skipped due to catching an "unimplemented" exception (just asking for confirmation)? If yes, are real multi-dimensional tests executed as well?

tests/unit_tests/dft/include/test_common.hpp
raphael-egan: ok, so the name contains most of the information already except for fwd/bwd unless I'm mistaken. I leave it up to you to judge if that's a desirable addition or not in the current standing.

I don't think the convention that I was referring to is documented anywhere. Your understanding of {r,c}, {f,b} and {i,o} is correct. {Ax}{Bx}C represent the (forward domain's) size (resp. 3D size AxBxC, 2D size BxC or 1D size C) and M is the batch size (number of transforms).
If we want to report the precision as well, we precede that name with {s,d} (single/double, respectively).

tests/unit_tests/dft/include/compute_out_of_place.hpp
FMarno: Yes, that would appear as a skipped test. real multi-dimensional tests are executed, you can see the full list in the log here https://github.com/oneapi-src/oneMKL/pull/275#issuecomment-1451761782

tests/unit_tests/dft/include/compute_out_of_place.hpp
FMarno: I'm seeing some weird behaviour when doing this. Specifically If I make the changes show [here](https://github.com/FMarno/oneMKL/compare/multi_d_many_batch_pr...FMarno:oneMKL:multi_d_investigate) then the buffer tests pass but cause the usm tests to fail, although the usm tests are unchanged. Interestingly, if you run the USM tests alone then they still pass.
I'll have a look into this so might take some time to get to your other comments.

raphael-egan: What kind of failure is it for USM tests when they fail? (exception thrown, crash, incorrect results). Is there a reason we'd want to test different problem configurations between USM and buffers as the changes seem to do? Why not implement the changes for either case?

tests/unit_tests/dft/include/compute_out_of_place.hpp
raphael-egan: OK, I think I understand why and how the batched mD r2c oop tests were passing so far: I suspect it was using default strides resulting in roughly twice as large backward data allocations than standard usage would expect.

tests/unit_tests/dft/include/compute_out_of_place.hpp
FMarno: Incorrect results. Here are my test logs [real_single_oop.txt](https://github.com/oneapi-src/oneMKL/files/10999498/real_single_oop.txt) [real_single_oop_USM.txt](https://github.com/oneapi-src/oneMKL/files/10999499/real_single_oop_USM.txt), it is only the 2D single batch tests that are failing. Hopefully its just a problem with the test.
I don't think we want to test different configurations but it seems best to have a quick look into this since its unexpected behaviour.

tests/unit_tests/dft/include/reference_dft.hpp
FMarno: done in https://github.com/oneapi-src/oneMKL/pull/275/commits/ae4cc894901834d94f923d9d77588e93cb85e4b3

tests/unit_tests/dft/include/reference_dft.hpp
FMarno: done in https://github.com/oneapi-src/oneMKL/pull/275/commits/ae4cc894901834d94f923d9d77588e93cb85e4b3

tests/unit_tests/dft/include/reference_dft.hpp
FMarno: done in https://github.com/oneapi-src/oneMKL/pull/275/commits/ae4cc894901834d94f923d9d77588e93cb85e4b3

tests/unit_tests/dft/include/reference_dft.hpp
FMarno: done in https://github.com/oneapi-src/oneMKL/pull/275/commits/ae4cc894901834d94f923d9d77588e93cb85e4b3

tests/unit_tests/dft/include/reference_dft.hpp
FMarno: We are using `ref_t` (long double) to try to maintain as much accuracy as possible for the double precision tests.

tests/unit_tests/dft/include/reference_dft.hpp
FMarno: no reason, I guess maybe for consistency with the other loops but I've just changed it in https://github.com/oneapi-src/oneMKL/pull/275/commits/ae4cc894901834d94f923d9d77588e93cb85e4b3

tests/unit_tests/dft/include/compute_out_of_place.hpp
FMarno: changed in https://github.com/oneapi-src/oneMKL/pull/275/commits/ae4cc894901834d94f923d9d77588e93cb85e4b3

tests/unit_tests/dft/include/compute_inplace.hpp
FMarno: fixed in https://github.com/oneapi-src/oneMKL/pull/275/commits/ae4cc894901834d94f923d9d77588e93cb85e4b3

tests/unit_tests/dft/include/compute_out_of_place.hpp
FMarno: I've emailed you with a minimal reproducer for the problem.
The suggested change is added here https://github.com/oneapi-src/oneMKL/pull/275/commits/14bfce1c072f440e22efaf30a7ee9db5252073f5

tests/unit_tests/dft/include/compute_out_of_place.hpp
FMarno: absolutely, thanks for saving me from figuring that out! addressed in https://github.com/oneapi-src/oneMKL/pull/275/commits/14bfce1c072f440e22efaf30a7ee9db5252073f5

tests/unit_tests/dft/include/compute_out_of_place.hpp
FMarno: yes, I believe that is correct. Comment removed in https://github.com/oneapi-src/oneMKL/pull/275/commits/14bfce1c072f440e22efaf30a7ee9db5252073f5

tests/unit_tests/dft/include/test_common.hpp
FMarno: Thank you for clarifying. Each test runs both forward and backward transforms so I don't think we can add that information.
Currently I would rather keep the name as is, I like the readability, but I will be happy to revisit this when there are more parameters.

tests/unit_tests/dft/include/reference_dft.hpp
FMarno: I'm not seeing a simple way to achieve that. I would rather just leave it for now.

tests/unit_tests/dft/include/test_common.hpp
raphael-egan: Sounds good to me, no problem!

raphael-egan (approved):
Thanks for the work!

Rbiessy (approved):

lhuot (approved):

PR #278, opened 2023-02-10
[DFT] Introduce a cuFFT backend for the DFT interface

FMarno:
# Description

This builds on #259 to introduce a cuFFT backend. tests are included in #261.
Currently this only supports single precision, in-place, complex-to-complex dfts. A another PR will follow soon to implement as much of the interface as possible.

# Checklist

## All Submissions

- [x ] Have you formatted the code using clang-format?

## New features

- [x ] Have you provided motivation for adding a new feature?


examples/dft/run_time_dispatching/real_fwd_usm.cpp
```diff
@@ -107,7 +107,7 @@ int main(int argc, char** argv) {
     print_example_banner();
 
     try {
-        sycl::device my_dev((sycl::default_selector()));
+        sycl::device my_dev(sycl::default_selector{});
```
hjabird: Is this change intended?

include/oneapi/mkl/dft/detail/descriptor_impl.hpp
```diff
@@ -46,6 +46,8 @@ inline commit_impl* get_commit(descriptor<prec, dom>& desc);
 template <precision prec, domain dom>
 class descriptor {
 public:
+    constexpr static auto precision = prec;
```
hjabird: Doesn't seem to be in the [spec](https://spec.oneapi.io/versions/latest/elements/oneMKL/source/domains/dft/descriptor.html), so I don't think these should be here.

src/dft/backends/cufft/commit.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include <sycl/sycl.hpp>
+#else
+#include <CL/sycl.hpp>
+#endif
+
+#include "oneapi/mkl/types.hpp"
+#include "oneapi/mkl/exceptions.hpp"
+#include "oneapi/mkl/detail/backends.hpp"
+
+#include "oneapi/mkl/dft/detail/commit_impl.hpp"
```
hjabird, line 31: The MKLGPU backend needed to use the `detail` namespace to avoid name collisions with the closed-source. Here, that isn't needed. You shouldn't need eg. `dft::detail::precision`.

src/dft/backends/cufft/commit.cpp
```diff
@@ -0,0 +1,108 @@
...
+                throw mkl::exception("DFT", "commit", "Device does not support double precision.");
+            }
+        }
+        cufftCreate(&plan);
+
+        // The cudaStream for the plan is set a execution time so the interop handler can pick the stream.
+
+        const cufftType type = CUFFT_C2C;
+
+        auto n_copy = std::make_unique<int[]>(config_values.dimensions.size());
```
hjabird: Can we use `std::array<int, MaxSupportedDims>` since we're limited to (I think) 3D here?

src/dft/backends/cufft/forward.cpp
```diff
@@ -0,0 +1,179 @@
...
+                *result_ptr = cufftSetStream(plan, stream);
+                if (*result_ptr == CUFFT_SUCCESS) {
+                    *result_ptr = cufftExecC2C(plan, inout_native, inout_native, CUFFT_FORWARD);
+                }
+            });
+        }).wait();
+
+        if (result != CUFFT_SUCCESS) {
+            throw oneapi::mkl::exception("DFT", "compute_forward(desc, inout)",
+                                         "cuFFTResult value of " + std::to_string(result));
```
hjabird: What stops us from throwing this in the host task, avoiding the need for the `.wait()`? Same goes for `compute_backward`.

hjabird (commented):
Nice work!

include/oneapi/mkl/dft/detail/descriptor_impl.hpp
FMarno: I see what you are saying but it does help a lot here https://github.com/oneapi-src/oneMKL/pull/278/files#diff-7fb76b683c117a962220553f8ad016e187f45e5b5707936414d7662c1516ad30R68.

Do you have suggestions for an alternative? I could name it something weird to avoid collision

FMarno: Actually I guess I could just check the data-type. That should be easy.

t4c1: Alternatively you can make it private and make compute functions friends.

src/dft/backends/cufft/backward.cpp
```diff
@@ -0,0 +1,179 @@
...
+            cgh.host_task([=](sycl::interop_handle ih) {
+                auto inout_native = reinterpret_cast<cufftComplex *>(
+                    ih.get_native_mem<sycl::backend::ext_oneapi_cuda>(inout_acc));
+                auto stream = ih.get_native_queue<sycl::backend::ext_oneapi_cuda>();
+                *result_ptr = cufftSetStream(plan, stream);
+                if (*result_ptr == CUFFT_SUCCESS) {
+                    *result_ptr = cufftExecC2C(plan, inout_native, inout_native, CUFFT_INVERSE);
+                }
+            });
+        }).wait();
```
t4c1: This wait only waits on the host task to complete, BUT NOT on the device side work enqueued in the host task. 

From cuFFT documentation I do not find it clear if the execute calls are synchronous or asynchronous. The following paragraph assumes they are asynchronous.

Instead you should call cudaStreamSynchronize inside the host task. This will ensure the host task events and buffer synchronization actually match the device execution. 

If you move the following check on the result to the host task this wait becomes redundant.

src/dft/backends/cufft/commit.cpp
t4c1: Even if we can not a std::vector here would be nicer.

src/dft/backends/cufft/forward.cpp
```diff
@@ -0,0 +1,179 @@
...
+/// the expected value.
+template <dft::detail::config_param Param, dft::detail::config_value Expected, typename DescT>
+inline auto expect_config(DescT &desc, const char *message) {
+    dft::detail::config_value actual{ 0 };
+    desc.get_value(Param, &actual);
+    if (actual != Expected) {
+        throw mkl::invalid_argument("DFT", "compute_forward", message);
+    }
+}
+} // namespace detail
```
t4c1: Are these two functions copies of the ones in backward.hpp?

FMarno: They should have had different arguments for the throw, that is fixed in https://github.com/oneapi-src/oneMKL/pull/278/commits/548a3d933e1501cbcac7135fec1beeb26724261f.

src/dft/backends/cufft/backward.cpp
FMarno: That was the intension since I only wanted to wait for result_ptr to be written to, not for the FFT to be executed. Anyway, I've thrown directly from the host task in https://github.com/oneapi-src/oneMKL/pull/278/commits/548a3d933e1501cbcac7135fec1beeb26724261f so that won't be an issue anymore.

src/dft/backends/cufft/forward.cpp
FMarno: Nothing apparently so I've done that in https://github.com/oneapi-src/oneMKL/pull/278/commits/548a3d933e1501cbcac7135fec1beeb26724261f. I must have got myself confused into thinking that was awkward at some point.

src/dft/backends/cufft/commit.cpp
FMarno: The std::array can be used so it is in https://github.com/oneapi-src/oneMKL/pull/278/commits/548a3d933e1501cbcac7135fec1beeb26724261f

examples/dft/run_time_dispatching/real_fwd_usm.cpp
FMarno: Reverted in https://github.com/oneapi-src/oneMKL/pull/278/commits/548a3d933e1501cbcac7135fec1beeb26724261f

include/oneapi/mkl/dft/detail/descriptor_impl.hpp
FMarno: I've just gone with my suggestion in https://github.com/oneapi-src/oneMKL/pull/278/commits/548a3d933e1501cbcac7135fec1beeb26724261f since it lead to quite simple code. I can reinvestigate this if it becomes an issue when doing the full implementation.

src/dft/backends/cufft/commit.cpp
FMarno: I believe I've addressed that in all possible place in https://github.com/oneapi-src/oneMKL/pull/278/commits/9688071e96d1d3ad6a58f23a4839c85a2d0a3e1e

Rbiessy (approved):

hjabird (approved):

t4c1 (approved):

FMarno:
I believe it will be easier for the maintainers to only review #284, so I am closing this PR in favour of adding the whole backend at once.

Issue #270, opened 2023-01-26
Functional_test build errors with cuda backend 


JinlongLi2001:
### I'm building oneMKL for cuda backend using these cmake options:
````cmake 
cmake \
-DCMAKE_CXX_COMPILER=clang++ \
-DCMAKE_C_COMPILER=clang \
-DBUILD_SHARED_LIBS=ON \
-DENABLE_CUBLAS_BACKEND=ON \
-DENABLE_CUSOLVER_BACKEND=ON \
-DENABLE_CURAND_BACKEND=ON \
-DENABLE_MKLCPU_BACKEND=OFF \
-DENABLE_MKLGPU_BACKEND=OFF \
-DREF_BLAS_ROOT=/home/ljl/Apps/lapack \
-DREF_LAPACK_ROOT=/home/ljl/Apps/lapack \
-DBUILD_FUNCTIONAL_TESTS=ON \
-DCMAKE_BUILD_TYPE=Debug \
-DBUILD_EXAMPLES=ON \
..
````
### cmake -j failed with these bash output :
````bash
/home/ljl/Downloads/oneMKL/tests/unit_tests/blas/extensions/gemm_bias.cpp:138:48: error: no matching function for call to 'gemm_bias'
                TEST_RUN_CT_SELECT(main_queue, oneapi::mkl::blas::row_major::gemm_bias, transa,
                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
````
- Seems header has not be included correctly
- The same errors founded in(only tested,maybe not all of them) :
````bash
unit_tests/blas/level2/hpmv.cpp
unit_tests/lapack/source/gebrd.cpp
````
- A header included by these sources may be not correct

mmeterel:
@jiuxiaocloud Thanks for reporting the issue. I believe this error is coming because all CUDA backends AND functional tests are enabled. AFAIK, if you want to enable functional testing, we can only enable one backend at a time. 

Do you need to enable all backend and testing at the same time? If not, can you try single backend?

(ENABLE_CUBLAS_BACKEND=ON, -DENABLE_CUSOLVER_BACKEND=ON, -DENABLE_CURAND_BACKEND=ON, -DBUILD_FUNCTIONAL_TESTS=ON)

mmeterel:
[...]

PR #279, opened 2023-02-13
[doc] add note about multiple CUDA backends limitation

mkrainiuk:
# Description

As it was mentioned in #270 we have a limitation about building multiple CUDA backends cause problems in functional testing.
This PR adds a note about this limitation to build instructions.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Attach a log. N/A
- [x] Have you formatted the code using clang-format? N/A

### DIFF
--- a/docs/building_the_project.rst
+++ b/docs/building_the_project.rst
@@ -585,6 +585,10 @@ CMake.
   the targeted hardware. This backend has only been tested for the ``gfx90a``
   architecture (MI210) at the time of writing.
 
+.. note::
+  When building with ``BUILD_FUNCTIONAL_TESTS=yes`` (default option) only single CUDA backend can be built
+  (`#270 <https://github.com/oneapi-src/oneMKL/issues/270>`_).
+
 .. _project_cleanup:
 
 Project Cleanup


akabalov (approved):

docs/building_the_project.rst
```diff
@@ -585,6 +585,10 @@ CMake.
   the targeted hardware. This backend has only been tested for the ``gfx90a``
   architecture (MI210) at the time of writing.
 
+.. note::
+  Build with multiple CUDA backends enabled could cause testing problems
```
mmeterel: I think, the problem is, when functional test flag is enabled and we try to build multiple backends, there will be errors in the build. Can we rephrase the note something like:
`When BUILD_FUNCTIONAL_TESTS is enabled, only single CUDA backend can be built.`

mkrainiuk: Good point, let me update the note.

mkrainiuk: Updated.

mmeterel (approved):
Thanks!

PR #282, opened 2023-02-20
[DFT] Allow the descriptor to be modified and recommitted

FMarno:
# Description

Allows the descriptor class for DFTs to be modified and recommitted. Some backends will likely need to regenerate the descriptor completely, but this will allow for those that don't.

This builds on top of #259

# Checklist

## All Submissions

## New features

### DIFF
--- a/include/oneapi/mkl/dft/detail/commit_impl.hpp
+++ b/include/oneapi/mkl/dft/detail/commit_impl.hpp
@@ -57,7 +57,7 @@ public:
 
     virtual void* get_handle() noexcept = 0;
 
-    virtual void commit(sycl::queue& queue, const dft_values<prec, dom>&) = 0;
+    virtual void commit(const dft_values<prec, dom>&) = 0;
 
 private:
     mkl::backend backend_;
--- a/include/oneapi/mkl/dft/detail/commit_impl.hpp
+++ b/include/oneapi/mkl/dft/detail/commit_impl.hpp
@@ -32,6 +32,12 @@ enum class backend;
 
 namespace oneapi::mkl::dft::detail {
 
+enum class precision;
+enum class domain;
+template <precision prec, domain dom>
+class dft_values;
+
+template <precision prec, domain dom>
 class commit_impl {
 public:
     commit_impl(sycl::queue queue, mkl::backend backend) : queue_(queue), backend_(backend) {}
@@ -50,6 +56,8 @@ public:
     }
 
     virtual void* get_handle() noexcept = 0;
+
+    virtual void commit(sycl::queue& queue, const dft_values<prec, dom>&) = 0;
 
 private:
     mkl::backend backend_;
--- a/include/oneapi/mkl/dft/detail/descriptor_impl.hpp
+++ b/include/oneapi/mkl/dft/detail/descriptor_impl.hpp
@@ -41,7 +41,7 @@ template <precision prec, domain dom>
 class descriptor;
 
 template <precision prec, domain dom>
-inline commit_impl* get_commit(descriptor<prec, dom>& desc);
+inline commit_impl<prec, dom>* get_commit(descriptor<prec, dom>& desc);
 
 template <precision prec, domain dom>
 class descriptor {
@@ -74,16 +74,16 @@ public:
 
 private:
     // Has a value when the descriptor is committed.
-    std::unique_ptr<commit_impl> pimpl_;
+    std::unique_ptr<commit_impl<prec, dom>> pimpl_;
 
     // descriptor configuration values_ and structs
     dft_values<prec, dom> values_;
 
-    friend commit_impl* get_commit<prec, dom>(descriptor<prec, dom>&);
+    friend commit_impl<prec, dom>* get_commit<prec, dom>(descriptor<prec, dom>&);
 };
 
 template <precision prec, domain dom>
-inline commit_impl* get_commit(descriptor<prec, dom>& desc) {
+inline commit_impl<prec, dom>* get_commit(descriptor<prec, dom>& desc) {
     return desc.pimpl_.get();
 }
 
--- a/src/dft/backends/descriptor.cpp
+++ b/src/dft/backends/descriptor.cpp
@@ -28,7 +28,7 @@ namespace dft {
 
 template <precision prec, domain dom>
 void descriptor<prec, dom>::commit(sycl::queue &queue) {
-    if (!pimpl_) {
+    if (!pimpl_ || pimpl_->get_queue() != queue) {
         pimpl_.reset(detail::create_commit(*this, queue));
     }
     pimpl_->commit(values_);
--- a/src/dft/backends/descriptor.cpp
+++ b/src/dft/backends/descriptor.cpp
@@ -28,7 +28,10 @@ namespace dft {
 
 template <precision prec, domain dom>
 void descriptor<prec, dom>::commit(sycl::queue &queue) {
-    pimpl_.reset(detail::create_commit(*this, queue));
+    if (!pimpl_) {
+        pimpl_.reset(detail::create_commit(*this, queue));
+    }
+    pimpl_->commit(queue, values_);
 }
 template void descriptor<precision::SINGLE, domain::COMPLEX>::commit(sycl::queue &);
 template void descriptor<precision::SINGLE, domain::REAL>::commit(sycl::queue &);
--- a/src/dft/backends/descriptor.cpp
+++ b/src/dft/backends/descriptor.cpp
@@ -31,7 +31,7 @@ void descriptor<prec, dom>::commit(sycl::queue &queue) {
     if (!pimpl_) {
         pimpl_.reset(detail::create_commit(*this, queue));
     }
-    pimpl_->commit(queue, values_);
+    pimpl_->commit(values_);
 }
 template void descriptor<precision::SINGLE, domain::COMPLEX>::commit(sycl::queue &);
 template void descriptor<precision::SINGLE, domain::REAL>::commit(sycl::queue &);
--- a/src/dft/backends/descriptor.cpp
+++ b/src/dft/backends/descriptor.cpp
@@ -29,6 +29,9 @@ namespace dft {
 template <precision prec, domain dom>
 void descriptor<prec, dom>::commit(sycl::queue &queue) {
     if (!pimpl_ || pimpl_->get_queue() != queue) {
+        if (pimpl_) {
+            pimpl_->get_queue().wait();
+        }
         pimpl_.reset(detail::create_commit(*this, queue));
     }
     pimpl_->commit(values_);
--- a/src/dft/backends/mklcpu/commit.cpp
+++ b/src/dft/backends/mklcpu/commit.cpp
@@ -60,7 +60,7 @@ public:
         }
     }
 
-    void commit(sycl::queue&, const detail::dft_values<prec, dom>& config_values) override {
+    void commit(const detail::dft_values<prec, dom>& config_values) override {
         set_value(handle, config_values);
         if (auto status = DftiCommitDescriptor(handle); status != DFTI_NO_ERROR) {
             throw oneapi::mkl::exception("dft/backends/mklcpu", "commit",
--- a/src/dft/backends/mklcpu/commit.cpp
+++ b/src/dft/backends/mklcpu/commit.cpp
@@ -63,7 +63,8 @@ public:
 
     void commit(const detail::dft_values<prec, dom>& config_values) override {
         set_value(handle, config_values);
-        if (auto status = DftiCommitDescriptor(handle); status != DFTI_NO_ERROR) {
+        auto status = DftiCommitDescriptor(handle);
+        if (status != DFTI_NO_ERROR) {
             throw oneapi::mkl::exception(
                 "dft/backends/mklcpu", "commit",
                 "DftiCommitDescriptor failed with status: " + std::to_string(status));
--- a/src/dft/backends/mklcpu/commit.cpp
+++ b/src/dft/backends/mklcpu/commit.cpp
@@ -40,10 +40,10 @@ namespace dft {
 namespace mklcpu {
 
 template <precision prec, domain dom>
-class commit_derived_impl final : public detail::commit_impl {
+class commit_derived_impl final : public detail::commit_impl<prec, dom> {
 public:
     commit_derived_impl(sycl::queue queue, const detail::dft_values<prec, dom>& config_values)
-            : detail::commit_impl(queue, backend::mklcpu) {
+            : detail::commit_impl<prec, dom>(queue, backend::mklcpu) {
         DFT_ERROR status = DFT_NOTSET;
         if (config_values.dimensions.size() == 1) {
             status = DftiCreateDescriptor(&handle, get_precision(prec), get_domain(dom), 1,
@@ -58,11 +58,11 @@ public:
             throw oneapi::mkl::exception("dft/backends/mklcpu", "commit",
                                          "DftiCreateDescriptor failed");
         }
+    }
 
+    void commit(sycl::queue&, const detail::dft_values<prec, dom>& config_values) override {
         set_value(handle, config_values);
-
-        status = DftiCommitDescriptor(handle);
-        if (status != DFTI_NO_ERROR) {
+        if (auto status = DftiCommitDescriptor(handle); status != DFTI_NO_ERROR) {
             throw oneapi::mkl::exception("dft/backends/mklcpu", "commit",
                                          "DftiCommitDescriptor failed");
         }
@@ -122,18 +122,19 @@ private:
 };
 
 template <precision prec, domain dom>
-detail::commit_impl* create_commit(const descriptor<prec, dom>& desc, sycl::queue& sycl_queue) {
+detail::commit_impl<prec, dom>* create_commit(const descriptor<prec, dom>& desc,
+                                              sycl::queue& sycl_queue) {
     return new commit_derived_impl<prec, dom>(sycl_queue, desc.get_values());
 }
 
-template detail::commit_impl* create_commit(const descriptor<precision::SINGLE, domain::REAL>&,
-                                            sycl::queue&);
-template detail::commit_impl* create_commit(const descriptor<precision::SINGLE, domain::COMPLEX>&,
-                                            sycl::queue&);
-template detail::commit_impl* create_commit(const descriptor<precision::DOUBLE, domain::REAL>&,
-                                            sycl::queue&);
-template detail::commit_impl* create_commit(const descriptor<precision::DOUBLE, domain::COMPLEX>&,
-                                            sycl::queue&);
+template detail::commit_impl<precision::SINGLE, domain::REAL>* create_commit(
+    const descriptor<precision::SINGLE, domain::REAL>&, sycl::queue&);
+template detail::commit_impl<precision::SINGLE, domain::COMPLEX>* create_commit(
+    const descriptor<precision::SINGLE, domain::COMPLEX>&, sycl::queue&);
+template detail::commit_impl<precision::DOUBLE, domain::REAL>* create_commit(
+    const descriptor<precision::DOUBLE, domain::REAL>&, sycl::queue&);
+template detail::commit_impl<precision::DOUBLE, domain::COMPLEX>* create_commit(
+    const descriptor<precision::DOUBLE, domain::COMPLEX>&, sycl::queue&);
 
 } // namespace mklcpu
 } // namespace dft
--- a/src/dft/backends/mklcpu/commit.cpp
+++ b/src/dft/backends/mklcpu/commit.cpp
@@ -55,16 +55,18 @@ public:
                                           config_values.dimensions.data());
         }
         if (status != DFTI_NO_ERROR) {
-            throw oneapi::mkl::exception("dft/backends/mklcpu", "commit",
-                                         "DftiCreateDescriptor failed");
+            throw oneapi::mkl::exception(
+                "dft/backends/mklcpu", "commit",
+                "DftiCreateDescriptor failed with status: " + std::to_string(status));
         }
     }
 
     void commit(const detail::dft_values<prec, dom>& config_values) override {
         set_value(handle, config_values);
         if (auto status = DftiCommitDescriptor(handle); status != DFTI_NO_ERROR) {
-            throw oneapi::mkl::exception("dft/backends/mklcpu", "commit",
-                                         "DftiCommitDescriptor failed");
+            throw oneapi::mkl::exception(
+                "dft/backends/mklcpu", "commit",
+                "DftiCommitDescriptor failed with status: " + std::to_string(status));
         }
     }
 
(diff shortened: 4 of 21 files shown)


include/oneapi/mkl/dft/detail/commit_impl.hpp
```diff
@@ -28,35 +28,37 @@
...
 
 #include "oneapi/mkl/detail/backends.hpp"
+#include "types_impl.hpp"
 
 namespace oneapi {
 namespace mkl {
 namespace dft {
 namespace detail {
 
+template <precision prec, domain dom>
```
Rbiessy, line 40: Why does `commit_impl` need these templates now?

include/oneapi/mkl/dft/detail/descriptor_impl.hpp
```diff
@@ -68,22 +68,22 @@ class descriptor {
     void commit(backend_selector<backend::mklgpu> selector);
 #endif
 
-    dft_values<prec, dom> get_values() {
+    dft_values<prec, dom>& get_values() {
```
Rbiessy: Shouldn't this return a const reference to avoid bypassing any checks that would be in `set_value`? The method could be made `const` too.

include/oneapi/mkl/dft/detail/commit_impl.hpp
```diff
@@ -28,35 +28,37 @@
...
+    commit_impl& operator=(const commit_impl& other) = delete;
 
     virtual ~commit_impl() = default;
 
     sycl::queue& get_queue() {
         return queue_;
     }
 
-    mkl::backend& get_backend() {
+    mkl::backend get_backend() {
```
Rbiessy: That makes sense, shouldn't this change move to the main implementation MR? The method could be make `const` if we do this.

include/oneapi/mkl/dft/detail/types_impl.hpp
```diff
@@ -100,6 +101,8 @@ class dft_values {
     using real_t = std::conditional_t<prec == precision::SINGLE, float, double>;
 
 public:
+    std::set<config_param> changed;
```
Rbiessy: Do we really need to remember which param was changed? I thought we would assume anything can change between 2 commits.

src/dft/backends/mklcpu/commit.cpp
```diff
@@ -96,44 +97,137 @@ class commit_derived_impl : public detail::commit_impl {
...
-        set_value_item(descHandle, DFTI_OUTPUT_STRIDES, config.output_strides.data());
-        set_value_item(descHandle, DFTI_BACKWARD_SCALE, config.bwd_scale);
-        set_value_item(descHandle, DFTI_FORWARD_SCALE, config.fwd_scale);
-        set_value_item(descHandle, DFTI_NUMBER_OF_TRANSFORMS, config.number_of_transforms);
-        set_value_item(descHandle, DFTI_INPUT_DISTANCE, config.fwd_dist);
-        set_value_item(descHandle, DFTI_OUTPUT_DISTANCE, config.bwd_dist);
-        set_value_item(
-            descHandle, DFTI_PLACEMENT,
-            (config.placement == config_value::INPLACE) ? DFTI_INPLACE : DFTI_NOT_INPLACE);
+    void set_value(DFTI_DESCRIPTOR_HANDLE& descHandle, detail::dft_values<prec, dom>& config) {
```
Rbiessy: Can you give some explanations why is this change needed? I am wondering if this is a generic improvement (and if so if it needs to go in the main implementation PR) or if this is needed to allow recommitting?

src/dft/descriptor_config_helper.hpp
```diff
@@ -94,18 +94,9 @@ PARAM_TYPE_HELPER(config_param::COMMIT_STATUS, config_value)
...
-            throw mkl::invalid_argument("DFT", "set_value", "Given nullptr.");
-        }
-        for (int i{ 0 }; i < rank; ++i) {
-            if (set_val[i] <= 0) {
-                throw mkl::invalid_argument("DFT", "set_value",
-                                            "Invalid length value (negative or 0).");
-            }
-        }
-        std::copy(set_val, set_val + rank, vals.dimensions.begin());
+        throw mkl::invalid_argument("DFT", "set_value", "Read-only parameter.");
```
Rbiessy: From our discussions we agreed that length does not need to be read-only for now. We will change the spec if we decide it is needed.

include/oneapi/mkl/dft/detail/commit_impl.hpp
Rbiessy: I think we should make as many methods `const` as possible, in particular in the `include` folder. For instance `descriptor::get_value`. IMO it makes sense to make the change in the main MR, what do you think?

include/oneapi/mkl/dft/detail/commit_impl.hpp
FMarno: Because I introduced a pure virtual function `commit` that takes a `dft_values<prec,dom>&` argument, meaning the class needs to know the type of `prec` and `dom`.

include/oneapi/mkl/dft/detail/descriptor_impl.hpp
FMarno: With the current design, `dft_values` includes the `changed` set that needs to be updated. If we get rid of `changed` then this should be altered as you suggest.

src/dft/backends/mklcpu/commit.cpp
FMarno: This is needed so that `config.changed.clear()` is communicated back to the `dft_values`

src/dft/backends/mklgpu/commit.cpp
```diff
@@ -51,31 +51,35 @@ namespace detail {
...
         if constexpr (prec == dft::detail::precision::DOUBLE) {
             if (!queue.get_device().has(sycl::aspect::fp64)) {
-                throw mkl::exception("DFT", "commit", "Device does not support double precision.");
+                throw mkl::exception("dft/backends/mklgpu", "commit",
+                                     "Device does not support double precision.");
             }
         }
+    }
 
+    virtual void commit(sycl::queue& queue,
```
FMarno: commit should use the queue stored in the super class

src/dft/backends/mklcpu/commit.cpp
FMarno: unless you are referring to `set_value_item`, in that case it is general improvement since I noticed the return wasn't really used. I will add it to the main PR.

Rbiessy: That wasn't clear, I was referring to all the changes in `set_value` and in particular why we were adding the for loop and switch. It looks like this won't be needed if we remove `changed` so we're good.
I agree for changing the return type to `void` and adding the reference. By the way maybe `config` can be a const reference?

include/oneapi/mkl/dft/detail/descriptor_impl.hpp
FMarno: This is now const

include/oneapi/mkl/dft/detail/commit_impl.hpp
FMarno: done

include/oneapi/mkl/dft/detail/types_impl.hpp
FMarno: removed

src/dft/backends/mklcpu/commit.cpp
FMarno: Ah ok. I've removed the changes to `set_value` and made `config` a const reference.

src/dft/descriptor_config_helper.hpp
FMarno: This has been reverted

src/dft/descriptor.cxx
```diff
@@ -176,7 +172,9 @@ void descriptor<prec, dom>::get_value(config_param param, ...) const {
     va_start(vl, param);
     switch (param) {
         case config_param::FORWARD_DOMAIN: *va_arg(vl, dft::domain*) = dom; break;
-        case config_param::DIMENSION: *va_arg(vl, std::int64_t*) = values_.dimensions.size(); break;
+        case config_param::DIMENSION:
+            *va_arg(vl, std::int64_t*) = static_cast<int64_t>(values_.dimensions.size());
```
hjabird: Isn't this implicit?

src/dft/backends/descriptor.cpp
```diff
@@ -28,7 +28,10 @@ namespace dft {
 
 template <precision prec, domain dom>
 void descriptor<prec, dom>::commit(sycl::queue &queue) {
-    pimpl_.reset(detail::create_commit(*this, queue));
+    if (!pimpl_) {
+        pimpl_.reset(detail::create_commit(*this, queue));
+    }
+    pimpl_->commit(values_);
```
hjabird, line 37: We need to check that the same queue is being used here as for a previous call.

src/dft/backends/mklgpu/descriptor.cpp
```diff
@@ -28,7 +28,10 @@ namespace dft {
 
 template <precision prec, domain dom>
 void descriptor<prec, dom>::commit(backend_selector<backend::mklgpu> selector) {
-    pimpl_.reset(mklgpu::create_commit(*this, selector.get_queue()));
+    if (!pimpl_) {
+        pimpl_.reset(mklgpu::create_commit(*this, selector.get_queue()));
+    }
+    pimpl_->commit(values_);
```
hjabird, line 37: We need to make sure that this isn't called on a descriptor that was previously for a different backend or queue.

hjabird (commented):
I think the biggest challenge is what to do in situations that the spec doesn't cover - eg. changing queue or backend for the descriptor.

src/dft/descriptor.cxx
FMarno: I think so, I don't really know how va_arg works. I'm not a fan of implicit casts, but I can see how this doesn't really fit in this PR. It was supposed to be moved to the main PR in [this commit](https://github.com/oneapi-src/oneMKL/pull/282/commits/0eccafb0bd56914d2c5bc214b65777bd785058b8#diff-f9389e5ddc5e740a1ee4555ba07c25a5ec83815da56ba79f60836b9c0079309eR179) but I guess I missed it. I can add it to the other PR.

src/dft/backends/descriptor.cpp
FMarno: I had a think about it and decided that if the queue is changed at all then the whole plan can be reconstructed. If there are performance problem with this (opportunities where the plan could have been shared) we can deal with that later.

src/dft/backends/mklgpu/descriptor.cpp
FMarno: Same as the generic version.

src/dft/backends/descriptor.cpp
```diff
@@ -28,7 +28,7 @@ namespace dft {
 
 template <precision prec, domain dom>
 void descriptor<prec, dom>::commit(sycl::queue &queue) {
-    if (!pimpl_) {
+    if (!pimpl_ || pimpl_->get_queue() != queue) {
         pimpl_.reset(detail::create_commit(*this, queue));
```
hjabird, line 35: Next problem!

This frees the underlying commit. Doesn't that need to be kept alive for asynchronous compute calls that might still be in flight?

FMarno: That's true. You would hope any required underlying memory would be kept alive by the runtime, but I'm not sure there are actually guarantees for that.
What do you think about introducing a `pimpl_->get_queue().wait()`?

src/dft/backends/mklcpu/commit.cpp
```diff
@@ -57,11 +57,11 @@ class commit_derived_impl final : public detail::commit_impl {
...
                                          "DftiCreateDescriptor failed");
         }
+    }
 
+    void commit(const detail::dft_values<prec, dom>& config_values) override {
         set_value(handle, config_values);
-
-        status = DftiCommitDescriptor(handle);
-        if (status != DFTI_NO_ERROR) {
+        if (auto status = DftiCommitDescriptor(handle); status != DFTI_NO_ERROR) {
```
t4c1: What is the point of this change? I found previous code easier to read.

src/dft/backends/descriptor.cpp
hjabird: I see two ways of doing this:
* Your suggested way - waiting on the queue (`pimpl_->get_queue().wait()`)
  * Blocks on changing descriptor queue.
* Shared pointers to the commit impl that get copied into the backend implementations
  * Non-blocking, but definitely not KISS.
  * Hard to know what should happen when querying commit status with this.

My preference would be for your suggestion.

src/dft/backends/mklcpu/commit.cpp
FMarno: The status variable had to be declared anyway since this was the first use in this new function.
It was pretty silly though, because the status wasn't used so the return could have been directly compared.
I've used the status now, but I can split it into a definition and an if statement no problem.

src/dft/backends/descriptor.cpp
FMarno: I have gone with the first. I don't really understand what would "hold" the shared_pointer other than the descriptor, and how would it know it was done? it would be a mess!

src/dft/backends/mklcpu/commit.cpp
Rbiessy: I have a small preference for moving the definition and assignment outside of the if block too.

Rbiessy: This can also be replaced by `DftiCommitDescriptor(handle) != DFTI_NO_ERROR` if we're not using the status elsewhere. I can see this is a common pattern in https://github.com/oneapi-src/oneMKL/pull/278 too.

FMarno: Updated in https://github.com/oneapi-src/oneMKL/pull/282/commits/ece0cc49ae480a551ec2aeb965369b40146e1e2c

Rbiessy (approved):

t4c1 (approved):

src/dft/backends/mklgpu/descriptor.cpp
hjabird: I don't understand your reply. Right now, if I've commited with `backend_selector<mklcpu>` then `pimpl_` will have a value, so only the values will change. Likewise, I can change the queue for `mklgpu` and that won't get passed onto the backend commit impl.

hjabird: In my previous comment, I think I must have been looking at the uncorrected version of the code. Apologies for the confusion.

FMarno:
Log of the test runs. My machine only supports single precision and through opencl, but I hope this demonstrates the descriptor tests are passing. 
[multi_commit_result.txt](https://github.com/oneapi-src/oneMKL/files/10944386/multi_commit_result.txt)

Rbiessy:
Next time this needs to be rebased you could fix the typo `recommited` -> `recommitted` in the commit description.

tests/unit_tests/dft/source/descriptor_tests.cpp
```diff
@@ -405,6 +406,56 @@ inline void set_readonly_values(sycl::queue& sycl_queue) {
...
+        std::make_pair(config_param::BACKWARD_SCALE, PrecisionType{ 3.4 }),
+        std::make_pair(config_param::NUMBER_OF_TRANSFORMS, std::int64_t{ 5 }),
+        std::make_pair(config_param::COMPLEX_STORAGE, config_value::COMPLEX_COMPLEX),
+        std::make_pair(config_param::REAL_STORAGE, config_value::REAL_REAL),
+        std::make_pair(config_param::CONJUGATE_EVEN_STORAGE, config_value::COMPLEX_COMPLEX),
+        std::make_pair(config_param::PLACEMENT, config_value::INPLACE),
+        std::make_pair(config_param::INPUT_STRIDES, strides.data()),
+        std::make_pair(config_param::OUTPUT_STRIDES, strides.data()),
+        std::make_pair(config_param::FWD_DISTANCE, std::int64_t{ 6 }),
+        std::make_pair(config_param::BWD_DISTANCE, std::int64_t{ 7 }),
```
raphael-egan: This set of new parameters is not necessarily describing a consistent in-place configuration: is that intended? If the tests are currently passing, it is not a guarantee of future successful commit for mkl backend(s) (see [MKLD-13788](https://jira.devtools.intel.com/browse/MKLD-13788)).

src/dft/backends/descriptor.cpp
```diff
@@ -28,7 +28,13 @@ namespace dft {
...
 void descriptor<prec, dom>::commit(sycl::queue &queue) {
-    pimpl_.reset(detail::create_commit(*this, queue));
+    if (!pimpl_ || pimpl_->get_queue() != queue) {
+        if (pimpl_) {
+            pimpl_->get_queue().wait();
+        }
+        pimpl_.reset(detail::create_commit(*this, queue));
+    }
+    pimpl_->commit(values_);
 }
```
raphael-egan, line 38: Maybe it's worth adding a `recommit_values` test which would change the queue between commit calls?

tests/unit_tests/dft/source/descriptor_tests.cpp
FMarno: I've tried to address this by making the configuration out-of-place and batching changes.

src/dft/backends/descriptor.cpp
FMarno: I've added a tests to show that there is a wait on the old queue here https://github.com/oneapi-src/oneMKL/pull/282/commits/088fa08c85c0ebdd5baa81d22c5cb190689ac1a6
I was unable to prove that work is submitted to the new queue or not submitted to the old.
* I don't believe there is any direct way to query if a queue has not completed tasks
* I don't believe there is a way to query if an event is associated with a task.
* When an command group function object is submitted to the queue, I don't believe there are any guarantees that is will complete before slow work on other queues.

Because of all this I can't think of how to test that reliably.

tests/unit_tests/dft/source/descriptor_tests.cpp
```diff
@@ -405,6 +409,106 @@ inline void set_readonly_values(sycl::queue& sycl_queue) {
...
+
+    // add some work to the busy queue
+    auto e = busy_queue.submit([&](sycl::handler& cgh) {
+        cgh.host_task([&] {
+            std::unique_lock<std::mutex> lock(cv_m);
+            ASSERT_TRUE(cv.wait_for(lock, 5s, [&] { return signal; })); // returns false on timeout
+            std::this_thread::sleep_for(500ms);
+        });
+    });
+    std::this_thread::sleep_for(500ms);
```
Rbiessy, line 491: I don't follow how this sleep and the one above reduce the false positives?
I'm not really concerned by this anyway so I'm approving again.

Rbiessy (approved):

src/dft/backends/descriptor.cpp
raphael-egan: Thank you for that. I think you're right for every bullet point you mentioned and I'm sorry if my original suggestion generated complications/misunderstanding on the testing scope. Those changes look fine to me.
To be clearer, my original concern was about a use case like 
1. create a descriptor `desc`;
2. create a queue `q1` and commit `desc` to `q1`;
3. delete `q1`;
4. create a queue `q2` and commit `desc` to `q2`;
5. submit a compute task to `desc`.
I am not 100% sure this works as expected for the closed-source mkl. But that is something we can/should check on our side as that would be an issue for us to fix anyways. We may consider this thread resolved.

tests/unit_tests/dft/source/descriptor_tests.cpp
raphael-egan: Thank you. I believe this is fine as long as we stick to 1D configurations (we have extra restrictions for real mD cases).

raphael-egan (approved):
Thank you for your work!

tests/unit_tests/dft/source/descriptor_tests.cpp
FMarno: I am generally against using sleeps in tests, but I hope these make sense.
This sleep gives the above event a chance to complete, semi-proving that it will have at least started. I didn't want to assert that is was in-progress because that could technically fail.
The one above makes the kernel last longer because the conditional variable will quite likely get the mutex before the commit begins, so the kernel may finish before the commit even started.

Potentially making both sleeps the same length ruins the effect of the first one...

src/dft/backends/descriptor.cpp
FMarno: Ok, that's interesting. I think this should work because objects like `sycl::queue` are kept alive by the SYCL runtime with reference counting semantics. I guess I could have a test that follows the steps you've detailed, waits on `q2` and then shows that the event returned from the task is complete. It's not perfect but its something. I'm about to finish for the day but I'll do this sometime tomorrow.

tests/unit_tests/dft/source/descriptor_tests.cpp
Rbiessy: Thanks I got it now.

FMarno: Timing updated

tests/unit_tests/dft/source/descriptor_tests.cpp
```diff
@@ -405,6 +409,138 @@ inline void set_readonly_values(sycl::queue& sycl_queue) {
...
+
+template <oneapi::mkl::dft::precision precision, oneapi::mkl::dft::domain domain>
+inline void swap_out_dead_queue(sycl::queue& sycl_queue) {
+    // test that commit still works when the previously committed queue is no longer in scope
+    // the queue is not actually dead (due to reference counting)
+
+    // commit the descriptor on the "busy" queue
+    oneapi::mkl::dft::descriptor<precision, domain> descriptor{ default_1d_lengths };
+    {
+        sycl::queue transient_queue;
```
Rbiessy: I think we should make sure this queue is created with the same device as `sycl_queue` for this test to be relevant. `transient_queue` will use the default selector but `sycl_queue` is created with an explicit device. We might as well re-use the same `exception_handler`.
Same comment for `change_queue_causes_wait`.

FMarno: addressed in https://github.com/oneapi-src/oneMKL/pull/282/commits/d6a5430af2ed5de2e9586c8bffc6e0e2884db615

src/dft/backends/descriptor.cpp
FMarno: I've added a test for that in https://github.com/oneapi-src/oneMKL/pull/282/commits/8d0a608327e5784305e216dffdf57d4060981b2d

Rbiessy (approved):

lhuot (approved):

lhuot:
@FMarno conflicts need to be resolved before I can merge this. Thanks!

FMarno:
In the last rebase I made some changes to compute_inplace.hpp and compute_out_of_place.hpp which meant the same descriptor was used twice with a recommit in the middle to update the strides.

Here is evidence of the passing tests. I still only have access to single precision and the OpenCL backend - 
[recommit.txt](https://github.com/oneapi-src/oneMKL/files/11089049/recommit.txt)

Rbiessy (approved):

PR #284, opened 2023-03-01
[DFT] Introduce the cuFFT backend for the DFT interface.

FMarno:
# Description

Introduce a cuFFT backend for the dft interface. I believe this exposes as much of the cuFFT library as we can use.

# Checklist

## All Submissions


t4c1 (approved):

FMarno:
evidence of the tests running. real-real transforms aren't implemented. [cufft_backend_run.txt](https://github.com/oneapi-src/oneMKL/files/10945086/cufft_backend_run.txt)

Rbiessy (approved):

FMarno:
Multidimensional tests (#275) rebased onto these changes - 
[cufft_backend_run.txt](https://github.com/oneapi-src/oneMKL/files/10959230/cufft_backend_run.txt)

lhuot:
@anantsrivastava30 could you please take a look at this PR? Thanks!

FMarno:
@anantsrivastava30 The conflicts caused by the #282 are quite big so I will need some time to sort that out. I'll give you a ping when I'm ready.

CMakeLists.txt
```diff
@@ -46,15 +46,24 @@ option(ENABLE_MKLGPU_BACKEND "" ON)
...
 option(ENABLE_CURAND_BACKEND "" OFF)
 option(ENABLE_ROCRAND_BACKEND "" OFF)
+
+# lapack
+option(ENABLE_CUSOLVER_BACKEND "" OFF)
 option(ENABLE_ROCSOLVER_BACKEND "" OFF)
-option(ENABLE_NETLIB_BACKEND "" OFF)
+
+# dft
+option(ENABlE_CUFFT_BACKEND "enable the cuFFT backend for DFT" OFF)
```
Rbiessy: There's a sneaky lower case `l` here and below! `ENABlE_CUFFT_BACKEND` -> `ENABLE_CUFFT_BACKEND`
Also the description could start with an upper case IMO.

README.md
```diff
@@ -224,12 +228,18 @@ Supported domains: BLAS, LAPACK, RNG
             <td align="center">LLVM*, hipSYCL</td>
         </tr>
         <tr>
-            <td rowspan=1 align="center">DFT</td>
+            <td rowspan=2 align="center">DFT</td>
```
Rbiessy: We forgot to add "DFT" to the list above this table: "Supported domains". We can consider adding it as part of this change.

FMarno:
@anantsrivastava30 this is ready for review now. Sorry for the wait, I had a horrible bug with cufftDestroy changing the cuda context for no known reason. Should be fine now.

README.md
FMarno: Addressed in https://github.com/oneapi-src/oneMKL/pull/284/commits/f03050fb1e0c24b3d947949b379984c1439fabcb

CMakeLists.txt
FMarno: addressed in https://github.com/oneapi-src/oneMKL/pull/284/commits/f03050fb1e0c24b3d947949b379984c1439fabcb and https://github.com/oneapi-src/oneMKL/pull/284/commits/e6bd53e78bbc209029e252e470b7d40712115515

FMarno:
@anantsrivastava30 I've fixed that merge conflict now so please review when you're ready

CMakeLists.txt
```diff
@@ -46,15 +46,24 @@ option(ENABLE_MKLGPU_BACKEND "" ON)
...
 option(ENABLE_CURAND_BACKEND "" OFF)
 option(ENABLE_ROCRAND_BACKEND "" OFF)
+
+# lapack
+option(ENABLE_CUSOLVER_BACKEND "" OFF)
 option(ENABLE_ROCSOLVER_BACKEND "" OFF)
-option(ENABLE_NETLIB_BACKEND "" OFF)
+
+# dft
+option(ENABLE_CUFFT_BACKEND "Enable the cuFFT backend for DFT" OFF)
```
lhuot: Why the string is non-empty for DFT only?
When is this string used?

examples/README.md
```diff
@@ -380,29 +380,54 @@ Running with single precision real data type on:
...
 
-Runtime dispatching example with both mklgpu backend
+Runtime dispatching example with both MKLGPU and cuFFT backend
 
 ```none
 SYCL_DEVICE_FILTER=gpu ./bin/example_dft_complex_fwd_buffer_mklgpu
 
 ########################################################################
-# Complex out-of-place forward transform for Buffer API's example:
+# DFTI complex in-place forward transform with USM API example:
```
lhuot, line 389: Description says USM API but executable above is fwd_buffer. Is that expected?

src/dft/backends/cufft/backward.cpp
```diff
@@ -0,0 +1,176 @@
+/*******************************************************************************
+* Copyright Codeplay Software Ltd.
```
lhuot, line 2: Is that the proper copyright? Sometimes copyright for new file are Intel, sometimes CodePlay. Can you clarify please?

tests/unit_tests/dft/include/compute_inplace.hpp
```diff
@@ -165,6 +163,8 @@ int DFT_Test<precision, domain>::test_in_place_buffer() {
         }
     }
 
+    // account for scaling that occurs during DFT
+    std::for_each(input.begin(), input.end(), [this](auto& x) { x *= forward_elements; });
```
lhuot: Why did you need to do that instead of using the backward scale as it was before?

tests/unit_tests/dft/include/compute_tester.hpp
```diff
@@ -131,7 +131,7 @@ struct DFT_Test {
         // Heuristic for the average-case error margins
         abs_error_margin =
             std::abs(max_norm_ref) * std::log2(static_cast<double>(forward_elements));
-        rel_error_margin = 5.0 * std::log2(static_cast<double>(forward_elements));
+        rel_error_margin = 200.0 * std::log2(static_cast<double>(forward_elements));
```
lhuot, line 134: Why did you need to increase the error tolerance?

lhuot:
> evidence of the tests running. real-real transforms aren't implemented. [cufft_backend_run.txt](https://github.com/oneapi-src/oneMKL/files/10945086/cufft_backend_run.txt)

Thanks for the log! Could you also please check that all oneMKL GPU backend tests are passing after these changes and attached the log to the PR?

CMakeLists.txt
FMarno: That is help text. It will appear in CMakeCache.txt (in the build folder) or when using a configuration tool like `ccmake`. e.g. from my CMakeCache.txt
```
ENABLE_CUBLAS_BACKEND:BOOL=OFF

//enable the cuFFT backend for DFT
ENABLE_CUFFT_BACKEND:BOOL=ON

ENABLE_CURAND_BACKEND:BOOL=OFF

ENABLE_CUSOLVER_BACKEND:BOOL=OFF
```
I will put a similar thing for the others.

tests/unit_tests/dft/include/compute_inplace.hpp
FMarno: cuFFT doesn't support forward or backward scaling. We consider adding a kernel that would do a simple multiplication, but decided that users could easily integrate this into their own kernels and get much better performance (since they could perform the scaling at the same time as other work, avoiding the cost of loading and storing all the data just to perform a single divide). Testing for forward and backward scale is included in our testing roadmap, a which point we will test many values, not just 1/N.

lhuot: So if a user of the oneMKL Interface library is using the oneMKL SYCL APIs with cuFFT backend and setting the bwd_scale then the scale will not be applied to the result of the FFT?

FMarno: Yes, that does seem bad.
I could add a check to the commit step for cuFFT so that an exception is thrown when `BACKWARD|FORWARD_SCALE != 1`

src/dft/backends/cufft/backward.cpp
FMarno: @Rbiessy advised me that all copyright headers for new files should say Codeplay. This was likely just a copy-paste error by myself. This is resolved in https://github.com/oneapi-src/oneMKL/pull/284/commits/a3ca80cf5f81b6551c067f913980282ae585e5bf and https://github.com/oneapi-src/oneMKL/pull/284/commits/bcebaf23b14094d60a7cf729f33462c54d461779

CMakeLists.txt
FMarno: Updated in https://github.com/oneapi-src/oneMKL/pull/284/commits/42d77b3ec0e8937f3bb7b711ce85b0f9cd6c04d3

examples/README.md
FMarno: That was the wrong binary name, I just reran the runtime example to check that the same output was given and it was. Fixed in https://github.com/oneapi-src/oneMKL/pull/284/commits/3292262947a725ea4fe604c9dd28d102f97eb480

tests/unit_tests/dft/include/compute_inplace.hpp
FMarno: I've added a fix in https://github.com/oneapi-src/oneMKL/pull/284/commits/fb06c4621a0555f0a2facad396e7603ba3a0466e. This does have a problem in that I had to rearrange the test since the invalid scale was stored, and then all subsequent attempts to commit became invalid.

anantsrivastava30:
can you add the output for the exmaple with NVIDIA backend.

tests/unit_tests/dft/include/compute_inplace.hpp
lhuot: I think we should find a way to make the backward, forward scale works for the cuFFT backend.
The goal of this project is to demonstrate that we can implement the oneMKL SYCL APIs as defined in the oneMKL specification for many platforms.
So as is we have a gap for Nvidia platform.
We could add the scale for the Nvidia backend in a different PR though and as a first step maybe throw an "unimplemented" exception for those scale for the cuFFT backend and update the test only for the cuFFT backend to either apply the scale on the test side if the exception is catch or just simply skip scaling in the cuFFT backend tests.

FMarno: I think the team is open to adding an extra kernel that will do the scaling when `FORWARD/BACKWARD_SCALE != 1` if it comes with some documentation that for cuFFT it will likely add overhead. We would like to add this in a later PR though.
When we get to adding the FORWARD/BACKWARD_SCALE tests, a range of values will be tested and the test will skip when a backend throws an "unimplemented" exception.

tests/unit_tests/dft/include/compute_tester.hpp
FMarno: cuFFT has failing tests at lower relative error margin. For example, if I use `100.*` instead of `200.0 *` I get one failing test
```
[ RUN      ] ComputeTestSuite/ComputeTests_in_place.REAL_DOUBLE_in_place_USM/sizes_128_batches_1_NVIDIA_A100_PCIE_40GB
Mismatching results: actual = 0.0533297 vs. reference = 0.0533297
 relative error = 2.66472e-13 absolute error = 1.42109e-14 relative bound = 1.55431e-13 absolute bound = 1.21182e-14
 at index i =113
/home/finlay/oneMKL/tests/unit_tests/dft/source/../include/compute_inplace.hpp:275: Failure
Value of: check_equal_vector( inout.data() + j * row_elements_to_conjugate_even_components(sizes.back()), input.data() + j * sizes.back(), sizes.back(), abs_error_margin, rel_error_margin, std::cout)
  Actual: false
Expected: true
/home/finlay/oneMKL/tests/unit_tests/dft/source/compute_tests.cpp:74: Failure
Expected equality of these values:
  res
    Which is: 0
  1
[  FAILED  ] ComputeTestSuite/ComputeTests_in_place.REAL_DOUBLE_in_place_USM/sizes_128_batches_1_NVIDIA_A100_PCIE_40GB, where GetParam() = (0x11ce720, 32-byte object <B0-CB 17-01 00-00 00-00 B8-CB 17-01 00-00 00-00 B8-CB 17-01 00-00 00-00 01-00 00-00 00-00 00-00>) (4 ms)
```
at the original value of `5.0*` I get 24 failing tests. I just thought 200 was a nice number that let all the tests pass, I admit not much more thought went into it.

FMarno:
Log from running after the latest changes.

[cufft_run.log](https://github.com/oneapi-src/oneMKL/files/11288858/cufft_run.log)
[mklgpu_run.log](https://github.com/oneapi-src/oneMKL/files/11288859/mklgpu_run.log)

Output of the example @anantsrivastava30 
```

########################################################################
# DFTI complex in-place forward transform with USM API example:
#
# Using APIs:
#   USM forward complex in-place
#   Run-time dispatch
#
# Using single precision (float) data type
#
# Device will be selected during runtime.
# The environment variable SYCL_DEVICE_FILTER can be used to specify
# SYCL device
#
########################################################################

Running DFT complex forward example on GPU device
Device name is: NVIDIA A100-PCIE-40GB
Running with single precision real data type:
DFT example run_time dispatch
DFT example ran OK

```

src/dft/backends/cufft/commit.cpp
```diff
@@ -0,0 +1,231 @@
...
+                            ostride, // ostride
+                            fwd_dist, // odist
+                            bwd_type, // type
+                            batch // batch
+        );
+        if (res != CUFFT_SUCCESS) {
+            throw mkl::exception("dft/backends/cufft", __FUNCTION__,
+                                 "Failed to create backward cuFFT plan.");
+        }
+        plans = { fwd_plan, bwd_plan };
```
lhuot: I believe you will have the same problems than for the CPU backend.
For real FFT there should be some cases where the input and output strides will be "valid" strides only for one of the descriptor (forward or backward descriptor) and here since you are forcing both plans to be created successfully then you might always fail to commit for some valid real FFT case.

FMarno: Thank you for pointing this out. cuFFT has ignored this so far, but I suppose a future update may change that and it would be awkward if it broke oneMKL! I took a bit of a different approach and checked the validity of the strides to decide if the plan should be created. This is done in https://github.com/oneapi-src/oneMKL/pull/284/commits/e34aae2d6f0ddaa07a8ac1fd18877b8d0c5b0782.

FMarno:
Test run after the strides validity check changes
[cufft_run.log](https://github.com/oneapi-src/oneMKL/files/11298065/cufft_run.log)

tests/unit_tests/dft/include/compute_inplace.hpp
lhuot: Could you please open an issue to track this?

lhuot (approved):
Looks good to me as long as the unimplemented scale is tracked with an issue. Thanks!

FMarno:
Thanks @lhuot. I've created an issue here https://github.com/oneapi-src/oneMKL/issues/313.
I have fixed the merge conflict but github seems to not be picking up the changes I've pushed???
When I get a chance I will run the tests once more

FMarno:
Test log [cufft_run.log](https://github.com/oneapi-src/oneMKL/files/11433642/cufft_run.log)

PR #290, opened 2023-03-07
[DFT] Add more warning flags 

Rbiessy:
# Description

This PR builds on https://github.com/oneapi-src/oneMKL/pull/259.

Fix warnings when compiling the DFT domain with some warning flags:
`-Wall -Wextra -Wshadow -Wconversion -Wpedantic`.
The same warnings can be added to more domains if desired.

As part of fixes the warnings this also splits the DFT descriptor tests
as the device was not needed for some of them.

# Checklist

## All Submissions

## New interfaces

it was accepted? # (RFC)

## New features

## Bug fixes

      GitHub issue or in this PR)?

### DIFF
--- a/cmake/WarningsUtils.cmake
+++ b/cmake/WarningsUtils.cmake
@@ -1,5 +1,5 @@
 #===============================================================================
-# Copyright 2023 Codeplay Software Ltd.
+# Copyright Codeplay Software Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/cmake/WarningsUtils.cmake
+++ b/cmake/WarningsUtils.cmake
@@ -1,0 +1,49 @@
+#===============================================================================
+# Copyright 2023 Codeplay Software Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+#
+# SPDX-License-Identifier: Apache-2.0
+#===============================================================================
+
+include_guard(GLOBAL)
+
+add_library(onemkl_warnings INTERFACE)
+
+set(ONEMKL_WARNINGS "")
+
+include(CheckCXXCompilerFlag)
+macro(add_warning flag)
+  check_cxx_compiler_flag(${flag} IS_SUPPORTED)
+  message(STATUS "DBG '${flag}': ${IS_SUPPORTED}")
+  if(${IS_SUPPORTED})
+    list(APPEND ONEMKL_WARNINGS ${flag})
+  else()
+    message(WARNING "Compiler does not support ${flag}")
+  endif()
+endmacro()
+
+add_warning("-Wall")
+add_warning("-Wextra")
+add_warning("-Wshadow")
+add_warning("-Wconversion")
+add_warning("-Wpedantic")
+
+message(STATUS "Using warnings: ${ONEMKL_WARNINGS}")
+
+# The onemkl_warnings target can be linked to any other target to enable warnings.
+target_compile_options(onemkl_warnings INTERFACE ${ONEMKL_WARNINGS})
+
+# Add the library to install package
+install(TARGETS onemkl_warnings EXPORT oneMKLTargets)
--- a/examples/dft/compile_time_dispatching/complex_fwd_buffer_mklgpu.cpp
+++ b/examples/dft/compile_time_dispatching/complex_fwd_buffer_mklgpu.cpp
@@ -29,7 +29,7 @@
 #include "oneapi/mkl.hpp"
 
 void run_example(const sycl::device& gpu_device) {
-    constexpr int N = 10;
+    constexpr std::size_t N = 10;
 
     // Catch asynchronous exceptions for cpu
     auto gpu_error_handler = [&](sycl::exception_list exceptions) {
@@ -55,7 +55,7 @@ void run_example(const sycl::device& gpu_device) {
     // 1. create descriptors
     oneapi::mkl::dft::descriptor<oneapi::mkl::dft::precision::SINGLE,
                                  oneapi::mkl::dft::domain::COMPLEX>
-        desc(N);
+        desc(static_cast<std::int64_t>(N));
 
     // 2. variadic set_value
     desc.set_value(oneapi::mkl::dft::config_param::PLACEMENT,
@@ -100,7 +100,7 @@ void print_example_banner() {
 //
 // Main entry point for example.
 //
-int main(int argc, char** argv) {
+int main(int /*argc*/, char** /*argv*/) {
     print_example_banner();
 
     try {
--- a/src/dft/backends/mklcpu/backward.cpp
+++ b/src/dft/backends/mklcpu/backward.cpp
@@ -39,30 +39,33 @@ namespace mklcpu {
 
 //In-place transform
 template <typename descriptor_type, typename data_type>
-ONEMKL_EXPORT void compute_backward(descriptor_type &desc, sycl::buffer<data_type, 1> &inout) {
+ONEMKL_EXPORT void compute_backward(descriptor_type& /*desc*/,
+                                    sycl::buffer<data_type, 1>& /*inout*/) {
     throw mkl::unimplemented("DFT", "compute_backward", "Not implemented for MKLCPU");
 }
 
 //In-place transform, using config_param::COMPLEX_STORAGE=config_value::REAL_REAL data format
 template <typename descriptor_type, typename data_type>
-ONEMKL_EXPORT void compute_backward(descriptor_type &desc, sycl::buffer<data_type, 1> &inout_re,
-                                    sycl::buffer<data_type, 1> &inout_im) {
+ONEMKL_EXPORT void compute_backward(descriptor_type& /*desc*/,
+                                    sycl::buffer<data_type, 1>& /*inout_re*/,
+                                    sycl::buffer<data_type, 1>& /*inout_im*/) {
     throw mkl::unimplemented("DFT", "compute_backward", "Not implemented for MKLCPU");
 }
 
 //Out-of-place transform
 template <typename descriptor_type, typename input_type, typename output_type>
-ONEMKL_EXPORT void compute_backward(descriptor_type &desc, sycl::buffer<input_type, 1> &in,
-                                    sycl::buffer<output_type, 1> &out) {
+ONEMKL_EXPORT void compute_backward(descriptor_type& /*desc*/, sycl::buffer<input_type, 1>& /*in*/,
+                                    sycl::buffer<output_type, 1>& /*out*/) {
     throw mkl::unimplemented("DFT", "compute_backward", "Not implemented for MKLCPU");
 }
 
 //Out-of-place transform, using config_param::COMPLEX_STORAGE=config_value::REAL_REAL data format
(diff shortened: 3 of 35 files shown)


Rbiessy:
Tests log: [tests_log.txt](https://github.com/oneapi-src/oneMKL/files/10911296/tests_log.txt)
Specific DFT test logs with the descriptor tests split: [test_dft_log.txt](https://github.com/oneapi-src/oneMKL/files/10911299/test_dft_log.txt)
Note the lapack failures are not related to these changes and the DFT example is expected to fail with the CPU backend with https://github.com/oneapi-src/oneMKL/pull/259 as it is not supported yet.

tests/unit_tests/dft/include/compute_tester.hpp
```diff
@@ -49,8 +49,8 @@ struct DFT_Test {
 
     enum class MemoryAccessModel { buffer, usm };
 
-    const std::int64_t size;
-    const std::int64_t conjugate_even_size;
+    const std::size_t size;
+    const std::size_t conjugate_even_size;
```
FMarno: I think I would prefer these to stay as `std::int64_t`. The logic for that is that they are likely to be used with `descriptor.set_value` and mismatching types has caused us problems there in the past.

tests/unit_tests/dft/source/descriptor_tests.cpp
```diff
@@ -401,53 +395,89 @@ inline void set_readonly_values(sycl::queue& sycl_queue) {
...
 
 template <oneapi::mkl::dft::precision precision, oneapi::mkl::dft::domain domain>
-int test(sycl::device* dev) {
-    sycl::queue sycl_queue(*dev, exception_handler);
+static int test_getter_setter() {
+    set_and_get_lengths<precision, domain>();
+    set_and_get_strides<precision, domain>();
+    set_and_get_values<precision, domain>();
+    get_readonly_values<precision, domain>();
+    set_readonly_values<precision, domain>();
```
FMarno: test changes should probably be in a separate PR. Did your other changes show any issues that necessitated these test changes?

FMarno (changes requested):
Looks pretty good.

tests/unit_tests/dft/include/compute_tester.hpp
Rbiessy: I was wondering about this. We often use these sizes to initialize a vector or a buffer which expect a `size_t`. This would be an issue if we try to test negative sizes so using `size_t` looks safer to me.

tests/unit_tests/dft/source/descriptor_tests.cpp
Rbiessy: That's fair, I can split it. This came up simply because some tests had an unused queue argument.

tests/unit_tests/dft/include/compute_tester.hpp
FMarno: Sorry, I'm not sure I follow the logic there. How would we test negative sizes if we use an unsized type or is the point that you want to avoid that? (maybe just creating new tests/adding to descriptor tests would be easier)

Rbiessy: Sorry I meant that I think we should avoid testing negative sizes with this descriptor so I have a preference for using `size_t` here.

FMarno: Ok then, I can get with that. please resolve this comment.

tests/unit_tests/dft/source/descriptor_tests.cpp
Rbiessy: Done with https://github.com/oneapi-src/oneMKL/pull/290/commits/2a89e405d2741d5a3319206393aa59436ab8bb04

Rbiessy:
I have reworded the first commit description.
Logs after re-running the descriptor tests: [test_dft_descriptor_log.txt](https://github.com/oneapi-src/oneMKL/files/10919938/test_dft_descriptor_log.txt)

FMarno (approved):

examples/dft/compile_time_dispatching/complex_fwd_buffer_mklgpu.cpp
```diff
@@ -55,7 +55,7 @@ void run_example(const sycl::device& gpu_device) {
     // 1. create descriptors
     oneapi::mkl::dft::descriptor<oneapi::mkl::dft::precision::SINGLE,
                                  oneapi::mkl::dft::domain::COMPLEX>
-        desc(N);
+        desc(static_cast<std::int64_t>(N));
```
hjabird, line 58: Why can't we just start with N as a `std::int64_t`?

tests/unit_tests/include/test_helper.hpp
```diff
@@ -217,6 +217,7 @@ static inline void aligned_free(void *p) {
 
 /* Support for Unified Shared Memory allocations for different backends */
 static inline void *malloc_shared(size_t align, size_t size, sycl::device dev, sycl::context ctx) {
+    (void)align;
```
hjabird, line 220:
```suggestion
static inline void *malloc_shared(std::size_t align, std::size_t size, sycl::device dev, sycl::context ctx) 
  (void)align;
```

cmake/WarningsUtils.cmake
```diff
@@ -0,0 +1,49 @@
...
+add_warning("-Wconversion")
+add_warning("-Wpedantic")
+
+message(STATUS "Using warnings: ${ONEMKL_WARNINGS}")
+
+# The onemkl_warnings target can be linked to any other target to enable warnings.
+target_compile_options(onemkl_warnings INTERFACE ${ONEMKL_WARNINGS})
+
+# Add the library to install package
+install(TARGETS onemkl_warnings EXPORT oneMKLTargets)
```
hjabird, line 49: Why does this need to be installed?

cmake/WarningsUtils.cmake
```diff
@@ -0,0 +1,49 @@
+#===============================================================================
+# Copyright 2023 Codeplay Software Ltd.
```
hjabird:
```suggestion
# Copyright Codeplay Software Ltd.
```

examples/dft/compile_time_dispatching/complex_fwd_buffer_mklgpu.cpp
Rbiessy: `N` is used twice to create `std::vector` which gives a warning if the size isn't unsigned. I figured it's best to avoid the risk of having a negative `N` which would result in a huge allocation on the host (well probably fail without allocating anything)

tests/unit_tests/include/test_helper.hpp
Rbiessy: I'm not sure about this, most of the common oneMKL code uses `size_t`. I have a preference for `std::size_t` which we use mostly in the DFT domain. I think it's more consistent if we don't change the common code.

cmake/WarningsUtils.cmake
Rbiessy: CMake is failing otherwise:
```
CMake Error: install(EXPORT "oneMKLTargets" ...) includes target "onemkl_dft_mklgpu_obj" which requires target "onemkl_warnings" that is not in any export set.
CMake Error: install(EXPORT "oneMKLTargets" ...) includes target "onemkl_dft" which requires target "onemkl_warnings" that is not in any export set.
CMake Error in CMakeLists.txt:
  export called with target "onemkl_dft_mklgpu_obj" which requires target
  "onemkl_warnings" that is not in any export set.


CMake Error in CMakeLists.txt:
  export called with target "onemkl_dft" which requires target
  "onemkl_warnings" that is not in any export set.
```

I'm not entirely sure why. I've made sure to link this target with `PRIVATE` so it does not affect users linking with oneMKL. Maybe CMake is not smart enough to detect that `onemkl_warnings` is not really needed?

hjabird (approved):

lhuot:
> Tests log: [tests_log.txt](https://github.com/oneapi-src/oneMKL/files/10911296/tests_log.txt) Specific DFT test logs with the descriptor tests split: [test_dft_log.txt](https://github.com/oneapi-src/oneMKL/files/10911299/test_dft_log.txt) Note the lapack failures are not related to these changes and the DFT example is expected to fail with the CPU backend with #259 as it is not supported yet.

I see 	5952 - dft/EXAMPLE/RT/real_fwd_usm/cpu (Failed)
in the test log. I thought we disabled the DFT test on CPU. Why is this still running?

examples/dft/compile_time_dispatching/complex_fwd_buffer_mklgpu.cpp
```diff
@@ -29,7 +29,7 @@
 #include "oneapi/mkl.hpp"
 
 void run_example(const sycl::device& gpu_device) {
-    constexpr int N = 10;
+    constexpr std::size_t N = 10;
```
lhuot, line 32: Is there a reason why not using std::int64_t here and remove the cast below?

lhuot:
I think it would be good to be consistent across all domains for the warning flag used. It can be done in separate PR though.
@mkrainiuk @aelizaro @mmeterel do you agree with the general idea of using those warning flag?

Rbiessy:
> > Tests log: [tests_log.txt](https://github.com/oneapi-src/oneMKL/files/10911296/tests_log.txt) Specific DFT test logs with the descriptor tests split: [test_dft_log.txt](https://github.com/oneapi-src/oneMKL/files/10911299/test_dft_log.txt) Note the lapack failures are not related to these changes and the DFT example is expected to fail with the CPU backend with #259 as it is not supported yet.
> 
> I see 5952 - dft/EXAMPLE/RT/real_fwd_usm/cpu (Failed) in the test log. I thought we disabled the DFT test on CPU. Why is this still running?

The log is older, we only recently disabled the DFT tests for the CPU backend. It's not a risk here anyway.

examples/dft/compile_time_dispatching/complex_fwd_buffer_mklgpu.cpp
Rbiessy: `N` is also used to initialize 2 `std::vector` which requires an unsigned size. I think it's safer to use unsigned here so that we avoid using a negative size and try to allocate a huge size.

mkrainiuk:
> I think it would be good to be consistent across all domains for the warning flag used. It can be done in separate PR though. @mkrainiuk @aelizaro @mmeterel do you agree with the general idea of using those warning flag?

I think it's a good practice to build without any warnings, so I agree with this idea. And we can add this flags to other domains after fixing all problems if we have any.

mkrainiuk (approved):

lhuot (approved):

lhuot:
@Rbiessy please let me know if you'd still like to make some changes here or if I can merge this PR. Thanks!

src/dft/backends/mklcpu/backward.cpp
```diff
@@ -39,66 +39,70 @@ namespace mklcpu {
...
+ONEMKL_EXPORT void compute_backward(descriptor_type& /*desc*/,
+                                    sycl::buffer<data_type, 1>& /*inout*/) {
     throw mkl::unimplemented("DFT", "compute_backward", "Not implemented for MKLCPU");
 }
 
 //In-place transform, using config_param::COMPLEX_STORAGE=config_value::REAL_REAL data format
 template <typename descriptor_type, typename data_type>
-ONEMKL_EXPORT void compute_backward(descriptor_type &desc, sycl::buffer<data_type, 1> &inout_re,
-                                    sycl::buffer<data_type, 1> &inout_im) {
+ONEMKL_EXPORT void compute_backward(descriptor_type& /*desc*/,
```
anantsrivastava30, line 49: whats the need for `/*desc*/` ?

Rbiessy: This is one way to fix unused variable warnings. The advantage is that it still gives the developer the original name of the variable if we decide to use it later.

Rbiessy:
> @Rbiessy please let me know if you'd still like to make some changes here or if I can merge this PR. Thanks!

Yes I am happy to see this merged once all the conversations are resolved, thanks!

Note that we should make sure future PRs in the DFT domain don't create warnings.

PR #292, opened 2023-03-10
[BLAS::SYCL-BLAS backend] Enable syclblas spr level 2 operator

muhammad-tanvir-1211:
# Description

This PR enables the `spr` Level 2 operator in SYCL-BLAS backend for oneMKL. 

FYI: this PR is blocked by #262  and #264.
 
# Checklist

## All Submissions
- [syclblas_ct.txt](https://github.com/oneapi-src/oneMKL/files/10943146/syclblas_ct.txt)
- [syclblas_rt.txt](https://github.com/oneapi-src/oneMKL/files/10943147/syclblas_rt.txt)
- The failures for double-precision functions on an iGPU that does not support double precision. In these cases, the backend throws `mkl::unsupported_device`, which results in a test failure.
- The tests were run with `--gtest_filter=-ImatcopyBatchStrideTestSuite*` because this test suite segfaults. This function is not implemented in SYCL-BLAS.

### DIFF
--- a/src/blas/backends/syclblas/syclblas_level2.cxx
+++ b/src/blas/backends/syclblas/syclblas_level2.cxx
@@ -147,7 +147,7 @@ void spmv(sycl::queue &queue, oneapi::mkl::uplo upper_lower, std::int64_t n, rea
 
 void spr(sycl::queue &queue, oneapi::mkl::uplo upper_lower, std::int64_t n, real_t alpha,
          sycl::buffer<real_t, 1> &x, std::int64_t incx, sycl::buffer<real_t, 1> &a) {
-    throw unimplemented("blas", "spr", "");
+    CALL_SYCLBLAS_FN(::blas::_spr, queue, upper_lower, n, alpha, x, incx, a);
 }
 
 void spr2(sycl::queue &queue, oneapi::mkl::uplo upper_lower, std::int64_t n, real_t alpha,
--- a/src/blas/backends/syclblas/CMakeLists.txt
+++ b/src/blas/backends/syclblas/CMakeLists.txt
@@ -140,7 +140,7 @@ if (NOT sycl_blas_FOUND)
   FetchContent_Declare(
     sycl-blas
     GIT_REPOSITORY https://github.com/codeplaysoftware/sycl-blas
-    GIT_TAG        20bfe691be07c9a2a316715094a2441f30e497d7
+    GIT_TAG        2c6c6a8a909e5a666d3426692bd33e55ba2be529
   )
   FetchContent_MakeAvailable(sycl-blas)
   message(STATUS "Looking for sycl_blas - downloaded")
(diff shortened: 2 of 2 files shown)


src/blas/backends/syclblas/syclblas_level2.cxx
```diff
@@ -147,7 +147,11 @@ void spmv(sycl::queue &queue, oneapi::mkl::uplo upper_lower, std::int64_t n, rea
 
 void spr(sycl::queue &queue, oneapi::mkl::uplo upper_lower, std::int64_t n, real_t alpha,
          sycl::buffer<real_t, 1> &x, std::int64_t incx, sycl::buffer<real_t, 1> &a) {
-    throw unimplemented("blas", "spr", "");
+    oneapi::mkl::uplo Uplo = (is_column_major() && upper_lower == oneapi::mkl::uplo::upper) ||
+                                     (!is_column_major() && upper_lower == oneapi::mkl::uplo::lower)
+                                 ? oneapi::mkl::uplo::upper
+                                 : oneapi::mkl::uplo::lower;
+    CALL_SYCLBLAS_FN(::blas::_spr, queue, Uplo, n, alpha, x, incx, a);
```
hjabird: In SYCL-BLAS common [here](https://github.com/oneapi-src/oneMKL/pull/292/files#diff-fdb63b676d60e1e1df94b116ce1092e17b547a896edfbe33e86494f61ab52b68) , it is assumed that only row-major is implemented. Consequently, I think this will still throw `mkl::unimplemented`. Modification is needed to syclblas_common.hpp.

mehdi-goli: @hjabird  Thanks for mentioning it, currently SYCL-BLAS has no Row-major implementation for any of the functions we have. SYCL-BLAS is column major only for all cases. Once the row-major integration is added, it should add be integrated for oneMKL Row-major. Hence, all the SYCL-BLAS integration for oneMKL should only include the column-major for now and throw un-implemented for the row-major.

muhammad-tanvir-1211: Hi @hjabird, Thanks for pointing this out. Like Mehdi said, SYCL-BLAS only supports `col-major` layout at the moment. However, `spr` operator has this unique property that it can work the same for both `col-major` and `row-major` layouts by carefully selecting the `upper` and `lower` triangular matrix layouts. That's is the reason I put this check here to select the appropriate value for `Uplo` variable based on whether the layout is `col-major` or `row-major`.

muhammad-tanvir-1211: Hi @hjabird, your explanation above was correct, I was hoping that the `RowMajor` tests would run with the changes I made for `spr` but it seems like it requires some other changes as well. I have removed the conditional statement for now and will hopefully address this when we add `row-major` support for SYCL-BLAS backend.

andrewtbarker (approved):

mkrainiuk (approved):

PR #288, opened 2023-03-06
[DFT] Synced mklcpu backend + Real_Real tests

anantsrivastava30:
# Description

This PR enables the mkl_cpu backend for the FFT domain to be introduced in https://github.com/oneapi-src/oneMKL/pull/259. This enables the DFT domain to support CPU device conforming to the [Spec OneAPI](https://spec.oneapi.io/versions/latest/elements/oneMKL/source/domains/dft/).
Supports `COMPLEX_STORAGE=COMPLEX_COMPLEX, REAL_REAL` for COMPLEX transforms; 

REAL STORAGE is not used anymore and instead CONJUGATE_EVEN_STORAGE is used for REAL transforms,

Tested with:
- [x] 1D {complex/real} {sp/dp} {split complex} https://github.com/oneapi-src/oneMKL/pull/261
- [x] mD, many-batched (**WIP**) https://github.com/oneapi-src/oneMKL/pull/275

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Attach a log.
- [x] Have you formatted the code using clang-format?

## New interfaces

- [ ] Have you provided motivation for adding a new feature as part of RFC and
it was accepted? # (RFC)
- [ ] What version of oneAPI spec the interface is targeted?
- [ ] Complete [New features](pull_request_template.md#new-features) checklist

## New features

- [ ] Have you provided motivation for adding a new feature?
- [ ] Have you added relevant tests?

## Bug fixes

- [ ] Have you added relevant regression tests?
- [ ] Have you included information on how to reproduce the issue (either in a
      GitHub issue or in this PR)?

PR #301, opened 2023-03-31
[DFT] Disable the MKLCPU DFT backend

### LINKED ISSUE #294: ctest fails at dft/EXAMPLE/RT/real_fwd_usm/cpu
I tried to build the latest oneMKL develop branch with following cmake command:
cmake -DCMAKE_CXX_COMPILER=clang++ \
-DCMAKE_C_COMPILER=clang \
-DENABLE_MKLCPU_BACKEND=TRUE \
-DENABLE_MKLGPU_BACKEND=TRUE \
-DENABLE_MKLCPU_THREAD_TBB=TRUE \
-DBUILD_FUNCTIONAL_TESTS=FALSE \
-DCMAKE_BUILD_TYPE=Release \
-BUILD_DOC=TRUE \
-DCMAKE_INSTALL_PREFIX=/DIR_TO_MKL/oneMKL/build/install/ \
..

Where clang++ & clang are the llvm components of oneAPI.
Then when I ran the ctest, I get one error only for the dft/EXAMPLE/RT/real_fwd_usm/cpu test case.
Does anyone have the same problem or any idea about this issue?
image

FMarno:
# Description

The MKLCPU backend is not yet supported (and currently causing the build to break due to a [copy-paste mistake](https://github.com/oneapi-src/oneMKL/blame/develop/src/dft/backends/mklcpu/descriptor.cpp#L35) by me), so we should disable it from building until #288 is complete.

The build error can be reproduce using this cmake command: `cmake -GNinja ../ -DCMAKE_CXX_COMPILER=$CXX -DCMAKE_C_COMPILER=$CC -DMKL_ROOT=/opt/intel/oneapi/mkl/2023.0.0 -DBUILD_FUNCTIONAL_TESTS=OFF`
This is because `ENABLE_MKLCPU_BACKEND` is defaulted to ON and when no `TARGET_DOMAIN` is specified it will add `dft` automatically.

Fixes #294

# Checklist

## All Submissions

- [N/A] Do all unit tests pass locally? Attach a log.

### DIFF
--- a/examples/dft/run_time_dispatching/CMakeLists.txt
+++ b/examples/dft/run_time_dispatching/CMakeLists.txt
@@ -29,9 +29,6 @@ endif()
 # If users build more than one backend (i.e. mklcpu and mklgpu, or mklcpu and CUDA), they may need to
 # overwrite SYCL_DEVICE_FILTER in their environment to run on the desired backend
 set(DEVICE_FILTERS "")
-if(ENABLE_MKLCPU_BACKEND)
-  list(APPEND DEVICE_FILTERS "cpu")
-endif()
 if(ENABLE_MKLGPU_BACKEND)
   list(APPEND DEVICE_FILTERS "gpu")
 endif()
--- a/src/dft/backends/CMakeLists.txt
+++ b/src/dft/backends/CMakeLists.txt
@@ -20,7 +20,3 @@
 if(ENABLE_MKLGPU_BACKEND)
   add_subdirectory(mklgpu)
 endif()
-
-if(ENABLE_MKLCPU_BACKEND)
-  add_subdirectory(mklcpu)
-endif()


Rbiessy (approved):

FMarno:
@lhuot @raphael-egan could you please take a look at this ASAP since it is causing builds to fail.

Rbiessy (approved):

hjabird (approved):

lhuot:
I am not sure to understand why there is a call to mklgpu::create_commit in the mklcpu backend implementation.
Can't you fix this problem https://github.com/oneapi-src/oneMKL/blame/develop/src/dft/backends/mklcpu/descriptor.cpp#L35 rather than disabling the build for mklcpu?

FMarno:
> I am not sure to understand why there is a call to mklgpu::create_commit in the mklcpu backend implementation. Can't you fix this problem https://github.com/oneapi-src/oneMKL/blame/develop/src/dft/backends/mklcpu/descriptor.cpp#L35 rather than disabling the build for mklcpu?

That is a indeed incorrect code, but there is probably a number of other less obvious instances of incorrect code so I think it's wise to disable this for now. I can imagine someone trying to use the MKLCPU DFT backend just because its there. If you like I can also fix that mistake.

mmeterel:
@FMarno Did you also update README so that it is in sync with what is being supported and what is not?

FMarno:
> @FMarno Did you also update README so that it is in sync with what is being supported and what is not?

MKLCPU was never listed as a supported backend for the DFT interface. Only MKLGPU is currently supported.

lhuot:
> > I am not sure to understand why there is a call to mklgpu::create_commit in the mklcpu backend implementation. Can't you fix this problem https://github.com/oneapi-src/oneMKL/blame/develop/src/dft/backends/mklcpu/descriptor.cpp#L35 rather than disabling the build for mklcpu?
> 
> That is a indeed incorrect code, but there is probably a number of other less obvious instances of incorrect code so I think it's wise to disable this for now. I can imagine someone trying to use the MKLCPU DFT backend just because its there. If you like I can also fix that mistake.

I agree that most of the mklcpu backend is not yet implemented and won't be used. But having it built allows to catch issues (such as this one). My worry is that if we disable it from the build, we may add even more issues in that backend without knowing and then it will just delay proper enabling of this backend and increase scope of work for https://github.com/oneapi-src/oneMKL/pull/288.

FMarno:
@lhuot that's fair. I have opened a new PR to fix the build instead https://github.com/oneapi-src/oneMKL/pull/303.

FMarno:
Reopening due to issues with #303

FMarno:
@lhuot Upon further consideration, I don't believe it is reasonable to expect contributors to maintain an incomplete implementation of the DFT interface. I understand this could potentially lead to more work to implement the MKLCPU backend, but we don't currently have plans for any more functionality changes. With this in mind I believe it would be best to disable this backend so users don't have to deal with this build error.

lhuot:
> Reopening due to issues with #303

Could you please give more details on what are the issues with #303? Tagging @anantsrivastava30 for awareness
Could you also please clarify what is currently broken? Is it the build of the library, the build of the examples or tests? Or are the examples or tests failing?
Which commit caused this failure?

FMarno:
Currently the MKLCPU backend to the DFT interface does not build due to a mistake in [this commit](https://github.com/oneapi-src/oneMKL/commit/f58b93f5cfa29a92b9b1e8605942d596f057374c), particularly [this line](https://github.com/oneapi-src/oneMKL/blame/develop/src/dft/backends/mklcpu/descriptor.cpp#L35). When `mklgpu` is changed to `mklcpu` on that line the code will build but then the descriptor tests will fail. Potentially those tests should skip in this situation, but that would require work on the tests and the mklcpu backend.
The issue with #303 is that we will now have failing tests, which I considered bad. Maybe that is not an issue in the context of the release cycle?
If we decide that we should fix the build instead of disabling it, then it becomes unclear if any changes are correct since the rest of the backend hasn't been implemented properly yet. Any further work would also have to guess at a patch for the MKLCPU backend, and it would also be unclear if this was correct. Alls these untested patches to the MKLCPU backend could add awkward debugging to the development time of #288.

lhuot:
> Currently the MKLCPU backend to the DFT interface does not build due to a mistake in [this commit](https://github.com/oneapi-src/oneMKL/commit/f58b93f5cfa29a92b9b1e8605942d596f057374c), particularly [this line](https://github.com/oneapi-src/oneMKL/blame/develop/src/dft/backends/mklcpu/descriptor.cpp#L35). When `mklgpu` is changed to `mklcpu` on that line the code will build but then the descriptor tests will fail. Potentially those tests should skip in this situation, but that would require work on the tests and the mklcpu backend. The issue with #303 is that we will now have failing tests, which I considered bad. Maybe that is not an issue in the context of the release cycle? If we decide that we should fix the build instead of disabling it, then it becomes unclear if any changes are correct since the rest of the backend hasn't been implemented properly yet. Any further work would also have to guess at a patch for the MKLCPU backend, and it would also be unclear if this was correct. Alls these untested patches to the MKLCPU backend could add awkward debugging to the development time of #288.

I agree that the incomplete implementation of the mklcpu backend creates a strange situation and it would be best to complete this backend asap. I think it makes sense to me, as the implementation is incomplete, to fix the build and disable the tests for the mklcpu backend. As long as this backend implementation is incomplete then the responsibility of all contributor is to maintain a passing build and the responsibility of the mklcpu backend contributor (here @anantsrivastava30) would be to make sure all tests are passing when completing the implementation (and I agree the more PR#288 is delayed the more this might be complicated).
@FMarno could you please modify this patch to fix the build and disable only the examples and tests but not the build?

FMarno:
@lhuot we've hit some problems doing that.
The example can be disable from ctest, but it would still be possible to manually run the exe with the env variable `SYCL_DEVICE_FILTER=cpu` set, causing the mklcpu code to be run (probably not a huge issue because the user is going out of their way to cause problems at that point). This is because all the runtime v-tables are still initialised.

The tests are harder to disable. The devices are enumerated in [main_test](https://github.com/oneapi-src/oneMKL/blob/develop/tests/unit_tests/main_test.cpp#L114) and I don't think it's possible to disable MKLCPU for one interface but not the others.

We could have missed something, but we're not seeing a simple fix here.

lhuot:
> @lhuot we've hit some problems doing that. The example can be disable from ctest, but it would still be possible to manually run the exe with the env variable `SYCL_DEVICE_FILTER=cpu` set, causing the mklcpu code to be run (probably not a huge issue because the user is going out of their way to cause problems at that point). This is because all the runtime v-tables are still initialised.
> 
> The tests are harder to disable. The devices are enumerated in [main_test](https://github.com/oneapi-src/oneMKL/blob/develop/tests/unit_tests/main_test.cpp#L114) and I don't think it's possible to disable MKLCPU for one interface but not the others.
> 
> We could have missed something, but we're not seeing a simple fix here.

Can't we add a check in the dft test on the device being used and skip the test if it is a CPU? For example extending the skip_test function here https://github.com/oneapi-src/oneMKL/blob/develop/tests/unit_tests/dft/include/compute_tester.hpp#LL104C1-L105C1

FMarno:
@lhuot please see my implementation at #303, following your suggestion. If you're happy with that then I am too.

FMarno:
fixed in #303

PR #291, opened 2023-03-08
[DFT] Split descriptor tests that do not need a device

Rbiessy:
# Description

This builds on https://github.com/oneapi-src/oneMKL/pull/259.

Some descriptor tests do not need a device as we do not need to commit
them. This follows from the discussion in https://github.com/oneapi-src/oneMKL/pull/261#discussion_r1101780387.
The descriptor is now only committed to check that the commit status is
updated.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Attach a log.
- [x] Have you formatted the code using clang-format?

## New interfaces

- [ ] Have you provided motivation for adding a new feature as part of RFC and
it was accepted? # (RFC)
- [ ] What version of oneAPI spec the interface is targeted?
- [ ] Complete [New features](pull_request_template.md#new-features) checklist

## New features

- [ ] Have you provided motivation for adding a new feature?
- [ ] Have you added relevant tests?

## Bug fixes

- [ ] Have you added relevant regression tests?
- [ ] Have you included information on how to reproduce the issue (either in a
      GitHub issue or in this PR)?

PR #303, opened 2023-04-03
[DFT] Fix MKLCPU DFT backend build issue

### LINKED ISSUE #294: ctest fails at dft/EXAMPLE/RT/real_fwd_usm/cpu

FMarno:
# Description

Fix a build issue introduced to the MKLCPU backend. This backend is not currently supported so the tests haven't been run.
This can be reproduced with the cmake setup 
```
cmake -GNinja -Bbuild -S. -DCMAKE_CXX_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin/icpx -DCMAKE_C_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin/icx -DMKL_ROOT=/opt/intel/oneapi/mkl/2023.0.0
```
 since ENABLE_MKLCPU_BACKEND is defaulted to ON and when no TARGET_DOMAIN is specified it will add dft automatically.

I've also updated the README because DFT was not listed as a supported domain.

Fixes #294 

# Checklist

## All Submissions

## Bug fixes

- [x] Have you added relevant regression tests? - **NA**
      GitHub issue or in this PR)?

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ $> clang++ -fsycl app.o –L$ONEMKL/lib –lonemkl_blas_mklcpu –lonemkl_blas_c
 
 ### Supported Configurations:
 
-Supported domains: BLAS, LAPACK, RNG
+Supported domains: BLAS, LAPACK, RNG, DFT
 
 #### Linux*
 
--- a/examples/dft/run_time_dispatching/CMakeLists.txt
+++ b/examples/dft/run_time_dispatching/CMakeLists.txt
@@ -29,9 +29,6 @@ endif()
 # If users build more than one backend (i.e. mklcpu and mklgpu, or mklcpu and CUDA), they may need to
 # overwrite SYCL_DEVICE_FILTER in their environment to run on the desired backend
 set(DEVICE_FILTERS "")
-if(ENABLE_MKLCPU_BACKEND)
-  list(APPEND DEVICE_FILTERS "cpu")
-endif()
 if(ENABLE_MKLGPU_BACKEND)
   list(APPEND DEVICE_FILTERS "gpu")
 endif()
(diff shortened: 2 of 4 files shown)


FMarno:
@Rbiessy realised the issue with this solution is that we now have failing tests for the MLKCPU backend enabled. Here is a log of the test run: [mklcpu_run.txt](https://github.com/oneapi-src/oneMKL/files/11139318/mklcpu_run-1680528997468.txt).
I decoded the return status of 3 with `DftiErrorMessage` and it said `Intel MKL DFTI ERROR: Inconsistent configuration parameters`.
I still personally believe #301 is the correct solution since the MKLCPU backend was never in a stable position to begin with.

Rbiessy:
https://github.com/oneapi-src/oneMKL/pull/291 may fix this issue with the MKLCPU backend since it is not committing the descriptor when testing that we can get and set the values.
I agree that https://github.com/oneapi-src/oneMKL/pull/301 is the safest approach until the CPU backend is properly enabled.

FMarno:
Test run - [test_run.txt](https://github.com/oneapi-src/oneMKL/files/11169685/test_run.txt)

lhuot (approved):
Thanks for the fix!

dbollapr (approved):

raphael-egan (approved):

PR #304, opened 2023-04-18
[BLAS] allow device pointers in cublas output scalars

### LINKED ISSUE #280: iamax and asum bug related to USM
# Summary
While working with an NVIDIA GPU, I attempted to use the iamax function, but encountered a segmentation fault that only occurred when using USM (Unified Shared Memory) and obtaining memory through the "malloc_device" function. After investigating possible causes, I found that obtaining memory through the "malloc_shared" function for the "result" variable resolved the issue. Further digging revealed that the variable is reinterpreted from int to int64_t in [this line](https://github.com/oneapi-src/oneMKL/blob/develop/src/blas/backends/cublas/cublas_level1.cpp#L1021) to ensure compatibility between oneMKL and cuBLAS. However, since the operation is performed on the host side, a segmentation fault occurs when memory for that variable is obtained with "malloc_device". It's worth noting that the same issue may be occurring with rocBlas.

On the other side, I have also found a related bug but with asum function. Getting memory for the "result" variable with "malloc_device" ends with a segmentation fault, but everything works well with "malloc_shared". In this case, I could not find any evidence of why it is crashing.

# Version
oneMKL githash: 45c43edcfaea9773a01fcd32b1354984d6780b8e

# Environment
* HW: NVIDIA 1030GT
* OS: Ubuntu 22.04

andrewtbarker:
# Description

For Level-1 BLAS functions that return a scalar (asum, iamax, iamin, dot, nrm2, rotg, rotmg), we previously assumed in the USM interfaces that the pointers were host-accessible. This PR allows such pointers to be device USM pointers.

For iamax, iamin, and rotmg the changes are somewhat ugly because of the need to copy data back and forth to satisfy the appropriate interface, but the result is similar to what is already done in the buffer APIs.

Fixes #280

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? [test.txt](https://github.com/oneapi-src/oneMKL/files/11266186/test.txt)

## Bug fixes

      GitHub issue or in this PR)?

### DIFF
--- a/src/blas/backends/cublas/cublas_level1.cpp
+++ b/src/blas/backends/cublas/cublas_level1.cpp
@@ -864,6 +864,8 @@ inline sycl::event dot(const char *func_name, Func func, sycl::queue &queue, int
                        const std::vector<sycl::event> &dependencies) {
     using cuDataType = typename CudaEquivalentType<T>::Type;
     overflow_check(n, incx, incy);
+    bool result_on_device =
+        sycl::get_pointer_type(result, queue.get_context()) == sycl::usm::alloc::device;
     auto done = queue.submit([&](sycl::handler &cgh) {
         int64_t num_events = dependencies.size();
         for (int64_t i = 0; i < num_events; i++) {
@@ -874,8 +876,14 @@ inline sycl::event dot(const char *func_name, Func func, sycl::queue &queue, int
             auto x_ = reinterpret_cast<const cuDataType *>(x);
             auto y_ = reinterpret_cast<const cuDataType *>(y);
             auto res_ = reinterpret_cast<cuDataType *>(result);
+            if (result_on_device) {
+                cublasSetPointerMode(handle, CUBLAS_POINTER_MODE_DEVICE);
+            }
             cublasStatus_t err;
             CUBLAS_ERROR_FUNC_T_SYNC(func_name, func, err, handle, n, x_, incx, y_, incy, res_);
+            if (result_on_device) {
+                cublasSetPointerMode(handle, CUBLAS_POINTER_MODE_DEVICE);
+            }
         });
     });
     return done;
@@ -939,7 +947,9 @@ sycl::event sdsdot(sycl::queue &queue, int64_t n, float sb, const float *x, int6
                    const float *y, int64_t incy, float *result,
                    const std::vector<sycl::event> &dependencies) {
     overflow_check(n, incx, incy);
-    // cuBLAS does not support sdot so we need to mimic sdot.
+    bool result_on_device =
+        sycl::get_pointer_type(result, queue.get_context()) == sycl::usm::alloc::device;
+    // cuBLAS does not support sdsdot so we need to mimic sdot.
     auto done = queue.submit([&](sycl::handler &cgh) {
         int64_t num_events = dependencies.size();
         for (int64_t i = 0; i < num_events; i++) {
@@ -950,13 +960,32 @@ sycl::event sdsdot(sycl::queue &queue, int64_t n, float sb, const float *x, int6
             auto x_ = reinterpret_cast<const float *>(x);
             auto y_ = reinterpret_cast<const float *>(y);
             auto res_ = reinterpret_cast<float *>(result);
+            if (result_on_device) {
+                cublasSetPointerMode(handle, CUBLAS_POINTER_MODE_DEVICE);
+            }
             cublasStatus_t err;
             CUBLAS_ERROR_FUNC_SYNC(cublasSdot, err, handle, n, x_, incx, y_, incy, res_);
+            if (result_on_device) {
+                cublasSetPointerMode(handle, CUBLAS_POINTER_MODE_HOST);
+            }
         });
     });
     done.wait();
-    result[0] = result[0] + sb;
-    return done;
+    if (result_on_device) {
+        // The following does copy device to host and then host to device
+        // just to adjust with sb constant. This is pretty inefficient, and
+        // should maybe be replaced with a sycl GPU kernel, but it duplicated what
+        // is done in the buffer API
+        float host_result;
+        queue.memcpy(&host_result, result, sizeof(float)).wait();
+        host_result += sb;
+        auto last_ev = queue.memcpy(result, &host_result, sizeof(float));
+        return last_ev;
+    }
+    else {
+        result[0] = result[0] + sb;
+        return done;
+    }
 }
 
 sycl::event dot(sycl::queue &queue, int64_t n, const float *x, int64_t incx, const float *y,
@@ -1038,27 +1067,22 @@ inline sycl::event iamax(const char *func_name, Func func, sycl::queue &queue, i
             }
         });
     });
+    done.wait();
     if (result_on_device) {
-        done.wait();
-        // the following is terrible, should be a sycl single_task kernel
-        // but with the cuda interop this always gives a PI_ERROR_INVALID_BINARY
-        // this duplicates what done in the buffer interface iamax
-        // ---
-        // 1. copy device int to host int
+        // The following does copy device to host and then host to device
+        // just to adjust to 0-base indexing. This is pretty inefficient, and
+        // should maybe be replaced with a sycl GPU kernel, but it duplicated what
+        // is done in the buffer API
         int host_int;
         int64_t host_int64;
         queue.memcpy(&host_int, int_res_p, sizeof(int)).wait();
-        // 2. adjust host int because of 0-base indexing
-        // 3. convert host int to host int64
         host_int64 = std::max((int64_t) host_int - 1, int64_t{ 0 });
-        // 4. copy host int64 to device int64 for output
         auto last_ev = queue.memcpy(result, &host_int64, sizeof(int64_t));
         last_ev.wait();
         sycl::free(int_res_p, queue);
         return last_ev;
     }
     else {
-        done.wait();
         result[0] = std::max((int64_t)(*int_res_p - 1), int64_t{ 0 });
         return done;
     }
@@ -1138,7 +1162,6 @@ inline sycl::event iamin(const char *func_name, Func func, sycl::queue &queue, i
         onemkl_cublas_host_task(cgh, queue, [=](CublasScopedContextHandler &sc) {
             auto handle = sc.get_handle(queue);
             auto x_ = reinterpret_cast<const cuDataType *>(x);
-            // auto int_res_p_ = reinterpret_cast<int *>(int_res_p);
             if (result_on_device) {
                 cublasSetPointerMode(handle, CUBLAS_POINTER_MODE_DEVICE);
             }
@@ -1151,27 +1174,22 @@ inline sycl::event iamin(const char *func_name, Func func, sycl::queue &queue, i
             }
         });
     });
+    done.wait();
     if (result_on_device) {
-        done.wait();
-        // the following is terrible, should be a sycl single_task kernel
-        // but with the cuda interop this always gives a PI_ERROR_INVALID_BINARY
-        // this duplicates what done in the buffer interface iamax
-        // ---
-        // 1. copy device int to host int
+        // The following does copy device to host and then host to device
+        // just to adjust to 0-base indexing. This is pretty inefficient, and
+        // should maybe be replaced with a sycl GPU kernel, but it duplicated what
+        // is done in the buffer API
         int host_int;
         int64_t host_int64;
         queue.memcpy(&host_int, int_res_p, sizeof(int)).wait();
-        // 2. adjust host int because of 0-base indexing
-        // 3. convert host int to host int64
-        host_int64 = std::max((int64_t) (host_int - 1), int64_t{ 0 });
-        // 4. copy host int64 to device int64 for output
+        host_int64 = std::max((int64_t) host_int - 1, int64_t{ 0 });
         auto last_ev = queue.memcpy(result, &host_int64, sizeof(int64_t));
         last_ev.wait();
         sycl::free(int_res_p, queue);
         return last_ev;
     }
     else {
-        done.wait();
         result[0] = std::max((int64_t)(*int_res_p - 1), int64_t{ 0 });
         return done;
     }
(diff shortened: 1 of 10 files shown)


src/blas/backends/cublas/cublas_level1.cpp
```diff
@@ -752,6 +760,16 @@ inline sycl::event rotg(const char *func_name, Func func, sycl::queue &queue, T1
                         T1 *s, const std::vector<sycl::event> &dependencies) {
     using cuDataType1 = typename CudaEquivalentType<T1>::Type;
     using cuDataType2 = typename CudaEquivalentType<T2>::Type;
+    auto ctx = queue.get_context();
+    bool results_on_device = sycl::get_pointer_type(a, ctx) == sycl::usm::alloc::device;
+    if (results_on_device) {
+        if (sycl::get_pointer_type(b, ctx) == sycl::usm::alloc::unknown ||
```
mmeterel: What is the motivation for this check and why only check `sycl::usm::alloc::unknown`?

src/blas/backends/cublas/cublas_level1.cpp
```diff
@@ -633,9 +635,15 @@ inline sycl::event asum(const char *func_name, Func func, sycl::queue &queue, in
...
             auto x_ = reinterpret_cast<const cuDataType1 *>(x);
             auto res_ = reinterpret_cast<cuDataType2 *>(result);
+            if (result_on_device) {
+                cublasSetPointerMode(handle, CUBLAS_POINTER_MODE_DEVICE);
+            }
             cublasStatus_t err;
             // ASUM does not support negative index
             CUBLAS_ERROR_FUNC_T_SYNC(func_name, func, err, handle, n, x_, std::abs(incx), res_);
+            if (result_on_device) {
+                cublasSetPointerMode(handle, CUBLAS_POINTER_MODE_HOST);
```
mmeterel, line 645: Did not understand why we are switching to `..MODE_HOST` here.

src/blas/backends/cublas/cublas_level1.cpp
```diff
@@ -942,13 +976,32 @@ sycl::event sdsdot(sycl::queue &queue, int64_t n, float sb, const float *x, int6
...
+            }
         });
     });
     done.wait();
-    result[0] = result[0] + sb;
-    return done;
+    if (result_on_device) {
+        // The following does copy device to host and then host to device
+        // just to adjust with sb constant. This is pretty inefficient, and
+        // should maybe be replaced with a sycl GPU kernel, but it duplicated what
```
mmeterel, line 999: Agreed with sycl GPU kernel. Any reason why it is not done in this PR?

src/blas/backends/cublas/cublas_level1.cpp
```diff
@@ -960,6 +1013,20 @@ template <typename Func, typename T>
...
+    if (results_on_device) {
+        if (sycl::get_pointer_type(d2, ctx) == sycl::usm::alloc::unknown ||
+            sycl::get_pointer_type(x1, ctx) == sycl::usm::alloc::unknown) {
+            throw oneapi::mkl::exception(
+                "blas", "rotmg", "If any pointer is device-only, all must be device accessible");
+        }
+    }
+    cuDataType *y1_;
+    if (results_on_device) {
+        y1_ = sycl::malloc_device<cuDataType>(1, queue);
```
mmeterel, line 1038: y1 is scalar and not an output. Why do we need to copy over to device?

src/blas/backends/cublas/cublas_level1.cpp
```diff
@@ -1001,7 +1080,15 @@ inline sycl::event iamax(const char *func_name, Func func, sycl::queue &queue, i
     // This change may cause failure as the result of integer overflow
     // based on the size.
     int int_res = 0;
-    int *int_res_p = &int_res;
+    int *int_res_p = nullptr;
+    bool result_on_device =
+        sycl::get_pointer_type(result, queue.get_context()) == sycl::usm::alloc::device;
+    if (result_on_device) {
+        int_res_p = sycl::malloc_device<int>(1, queue);
```
mmeterel, line 1098: Why do we need this temporary memory? Can't we use `result` provided by user?

src/blas/backends/cublas/cublas_level1.cpp
```diff
@@ -1010,16 +1097,37 @@ inline sycl::event iamax(const char *func_name, Func func, sycl::queue &queue, i
...
+            }
         });
     });
     done.wait();
-    result[0] = std::max((int64_t)(*int_res_p - 1), int64_t{ 0 });
-    return done;
+    if (result_on_device) {
+        // The following does copy device to host and then host to device
+        // just to adjust to 0-base indexing. This is pretty inefficient, and
+        // should maybe be replaced with a sycl GPU kernel, but it duplicated what
```
mmeterel, line 1127: would prefer avoiding D2H+H2D copies.

tests/unit_tests/blas/include/test_common.hpp
```diff
@@ -460,6 +460,13 @@ typename std::enable_if<std::is_integral<fp>::value, bool>::type check_equal(fp
     return (x == x_ref);
 }
 
+template <typename fp>
+bool check_equal_ptr(sycl::queue queue, fp *x, fp x_ref, int error_mag) {
```
mmeterel, line 464: can we give a more descriptive name to indicate `x` is actually on device?

tests/unit_tests/blas/level1/rotg_usm.cpp
```diff
@@ -86,15 +86,29 @@ int test(device *dev, oneapi::mkl::layout layout) {
...
+    }
 
-    a_p[0] = a;
-    b_p[0] = b;
-    s_p[0] = s;
-    c_p[0] = c;
+    main_queue.memcpy(a_p, &a, sizeof(fp)).wait();
+    main_queue.memcpy(b_p, &b, sizeof(fp)).wait();
+    main_queue.memcpy(s_p, &s, sizeof(fp)).wait();
+    main_queue.memcpy(c_p, &c, sizeof(fp_scalar)).wait();
```
mmeterel: Can we avoid these if `malloc_shared` is used? Also no need for multiple `wait()`. One at the end is sufficient.

src/blas/backends/cublas/cublas_level1.cpp
andrewtbarker: `unknown` is the only type that is not device-accessible. We could remove the check entirely, users will get a segfault instead of this error message.

src/blas/backends/cublas/cublas_level1.cpp
andrewtbarker: `...MODE_HOST` is the default for our use of cublas in this project, so we switch back to this after the cublas call. I am trying to be consistent with what is done for the buffer APIs above.

src/blas/backends/cublas/cublas_level1.cpp
andrewtbarker: I tried it and could not get it to work. Any SYCL kernel I launched got a `PI_ERROR_INVALID_BINARY` error.

src/blas/backends/cublas/cublas_level1.cpp
andrewtbarker: cublas expects a pointer here, for usm::shared we can do `y1_ = reinterpret_cast<...>(&y1);` but once we set `CUBLAS_POINTER_MODE_DEVICE` cuda will want a device-accessible pointer for every scalar argument (whether input or output). Taking `&y1` when `y1` is on host leads to a segfault.

src/blas/backends/cublas/cublas_level1.cpp
andrewtbarker: `result` is `int64_t`, `int_res_p` is `int`. This problem (and workaround) was present before this PR.

src/blas/backends/cublas/cublas_level1.cpp
andrewtbarker: Me too! Note that the buffer APIs do this too, and I did try to avoid it. If you know of a way to avoid it I will be happy to put it in.

tests/unit_tests/blas/include/test_common.hpp
andrewtbarker: I'm happy to change the name, but this function will work fine whether x is on host or device or shared.

tests/unit_tests/blas/level1/rotg_usm.cpp
andrewtbarker: Agreed on the `.wait()` (fixed). We can certainly avoid these if `malloc_shared` is used, but the point of the new tests is to test what happens with device memory, which caused a segfault before this PR.

src/blas/backends/cublas/cublas_level1.cpp
andrewtbarker: See eg line 57 [here](https://github.com/oneapi-src/oneMKL/blob/develop/src/blas/backends/cublas/cublas_level1.cpp#L57)

andrewtbarker:
@mmeterel Is there anything else you would like to see changed in this PR?

src/blas/backends/cublas/cublas_level1.cpp
mmeterel: I understand switching to `...MODE_HOST` especially when we run all BLAS functions back to back with ctest. MY confusion was, what does the example user code calling `asum` require/expect for the pointer mode when the results are on device?

src/blas/backends/cublas/cublas_level1.cpp
```diff
@@ -752,6 +760,16 @@ inline sycl::event rotg(const char *func_name, Func func, sycl::queue &queue, T1
...
     using cuDataType1 = typename CudaEquivalentType<T1>::Type;
     using cuDataType2 = typename CudaEquivalentType<T2>::Type;
+    auto ctx = queue.get_context();
+    bool results_on_device = sycl::get_pointer_type(a, ctx) == sycl::usm::alloc::device;
+    if (results_on_device) {
+        if (sycl::get_pointer_type(b, ctx) == sycl::usm::alloc::unknown ||
+            sycl::get_pointer_type(c, ctx) == sycl::usm::alloc::unknown ||
+            sycl::get_pointer_type(s, ctx) == sycl::usm::alloc::unknown) {
+            throw oneapi::mkl::exception(
+                "blas", "rotg", "If any pointer is device-only, all must be device accessible");
```
mmeterel: The check does not match with the error message. We are only checking for the case where `a` is device-only. Is this on purpose to avoid all possible combinations?
Also, can we change "device-only" to "only device accessible" or something similar?

src/blas/backends/cublas/cublas_level1.cpp
mmeterel: Should we open a ticket to LLVM compiler?

src/blas/backends/cublas/cublas_level1.cpp
mmeterel: Thanks. I did not realize the APIs are different in terms of y1 (pointer vs scalar)

src/blas/backends/cublas/cublas_level1.cpp
mmeterel: Buffer API does D2H only I think. If it is a compiler bug, we could open a ticket.

mmeterel:
> @mmeterel Is there anything else you would like to see changed in this PR?

@andrewtbarker added my comments.

src/blas/backends/cublas/cublas_level1.cpp
andrewtbarker: I think we are not expecting a user of the oneMKL interfaces to know or care about this mode, we only use it internally when we interface with cuBLAS.

My understanding (based on pre-existing MODE choice in buffer API before this PR) is that the entire project expects `CUBLAS_POINTER_MODE_HOST` to be set in general. When we see a device pointer for a scalar parameter, we switch temporarily to `...MODE_DEVICE`, make a cuBLAS call, and switch back. This was the case for the buffer API before this PR and now it is the case for USM as well.

I am not sure if I am actually understanding your question, let me know if I'm missing the point.

src/blas/backends/cublas/cublas_level1.cpp
andrewtbarker: From the comment on `iamax` in the buffer API [here](https://github.com/oneapi-src/oneMKL/blob/develop/src/blas/backends/cublas/cublas_level1.cpp#L476):

> This requires to bring the data to host, copy it, and return it back to the device

I am not sure it is a compiler bug. A standalone SYCL kernel runs fine on nvidia hardware, there is some issue with the combination of the SYCL kernels and the cuBLAS calls which may be my mistake. Figuring out exactly what it is may be a significant effort.

In my opinion fixing the bug in #280 and reaching parity with the buffer interface is a sufficient improvement to justify this PR even if performance is not ideal in a case that was not even functional before. If the performance or memory use becomes a problem we can deal with that at a later stage.

src/blas/backends/cublas/cublas_level1.cpp
andrewtbarker: Fixed, thanks.

src/blas/backends/cublas/cublas_level1.cpp
mmeterel: I was probably interpreting the `CUBLAS_POINTER_MODE_HOST` and `CUBLAS_POINTER_MODE_DEVICE` macros wrong. I was thinking, the application needs to set `CUBLAS_POINTER_MODE_DEVICE` if the result is on device, but after reading the cuBLAS documentation, looks like these modes are only indicating whether the CPU host is blocked by cuBLAS call or not. Sorry for the confusion.

src/blas/backends/cublas/cublas_level1.cpp
mmeterel: Thanks

src/blas/backends/cublas/cublas_level1.cpp
mmeterel: Agreed. Could you open an issue to this repo to share your findings on adding a SYCL kernel? If/when we work on performance parity with cuBLAS native calls for these APIs, your findings could be useful.

mmeterel (approved):
Thanks for working on this Andrew!

src/blas/backends/cublas/cublas_level1.cpp
andrewtbarker: Good idea, opened #312

andrewtbarker:
@mkrainiuk Can you take a quick look at this?

mkrainiuk (approved):
Thanks for the PR. Looks good to me. Do we have any performance implications?

andrewtbarker:
@mkrainiuk Performance implications for existing use cases should be minimal, just the overhead of checking whether a pointer is a device pointer. For the newly supported use case the performance is not optimal (opened #312 to track) but it was not functional before.

PR #305, opened 2023-04-20
Get rid of warning when we #include "cuda.hpp" from the LLVM SYCL backends for RNG

andreyfe1:
# Description

Recent compilers issue warnings when we `#include "cuda.hpp"` from the LLVM SYCL backends. This PR removes those warnings.

# Checklist

## All Submissions


aelizaro (approved):
Looks good to me, thank you!

mkrainiuk (approved):

PR #311, opened 2023-05-01
[DFT] Rearrange DFT compute tests so unimplemented always skips

FMarno:
# Description

Since #288, MKLGPU now fails the REAL_REAL tests because the descriptor commit throws (due to not supporting REAL_REAL for COMPLEX_STORAGE). I've rearranged the tests so they always skip when something is not implemented.

To reproduce run the tests with the MKLGPU backend. The complex-to-complex `*real_real*` test will fail.

Fixes # (GitHub issue)

# Checklist

## All Submissions

## Bug fixes

      GitHub issue or in this PR)?

### DIFF
--- a/tests/unit_tests/dft/include/compute_inplace.hpp
+++ b/tests/unit_tests/dft/include/compute_inplace.hpp
@@ -124,13 +124,7 @@ int DFT_Test<precision, domain>::test_in_place_buffer() {
     {
         sycl::buffer<FwdInputType, 1> inout_buf{ inout_host };
 
-        try {
-            oneapi::mkl::dft::compute_forward<descriptor_t, FwdInputType>(descriptor, inout_buf);
-        }
-        catch (oneapi::mkl::unimplemented& e) {
-            std::cout << "Skipping test because: \"" << e.what() << "\"" << std::endl;
-            return test_skipped;
-        }
+        oneapi::mkl::dft::compute_forward<descriptor_t, FwdInputType>(descriptor, inout_buf);
 
         {
             auto acc_host = inout_buf.template get_host_access();
@@ -158,14 +152,8 @@ int DFT_Test<precision, domain>::test_in_place_buffer() {
             commit_descriptor(descriptor, sycl_queue);
         }
 
-        try {
-            oneapi::mkl::dft::compute_backward<std::remove_reference_t<decltype(descriptor)>,
-                                               FwdInputType>(descriptor, inout_buf);
-        }
-        catch (oneapi::mkl::unimplemented& e) {
-            std::cout << "Skipping test because: \"" << e.what() << "\"" << std::endl;
-            return test_skipped;
-        }
+        oneapi::mkl::dft::compute_backward<std::remove_reference_t<decltype(descriptor)>,
+                                           FwdInputType>(descriptor, inout_buf);
     }
 
     if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
@@ -231,16 +219,10 @@ int DFT_Test<precision, domain>::test_in_place_USM() {
         std::copy(input.begin(), input.end(), inout.begin());
     }
 
-    try {
-        std::vector<sycl::event> dependencies;
-        oneapi::mkl::dft::compute_forward<descriptor_t, FwdInputType>(descriptor, inout.data(),
-                                                                      dependencies)
-            .wait();
-    }
-    catch (oneapi::mkl::unimplemented& e) {
-        std::cout << "Skipping test because: \"" << e.what() << "\"" << std::endl;
-        return test_skipped;
-    }
+    std::vector<sycl::event> no_dependencies;
+    oneapi::mkl::dft::compute_forward<descriptor_t, FwdInputType>(descriptor, inout.data(),
+                                                                  no_dependencies)
+        .wait();
 
     if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
         std::vector<FwdInputType> conjugate_even_ref =
@@ -262,17 +244,10 @@ int DFT_Test<precision, domain>::test_in_place_USM() {
         commit_descriptor(descriptor, sycl_queue);
     }
 
-    try {
-        std::vector<sycl::event> dependencies;
-        sycl::event done =
-            oneapi::mkl::dft::compute_backward<std::remove_reference_t<decltype(descriptor)>,
-                                               FwdInputType>(descriptor, inout.data());
-        done.wait();
-    }
-    catch (oneapi::mkl::unimplemented& e) {
-        std::cout << "Skipping test because: \"" << e.what() << "\"" << std::endl;
-        return test_skipped;
-    }
+    sycl::event done =
+        oneapi::mkl::dft::compute_backward<std::remove_reference_t<decltype(descriptor)>,
+                                           FwdInputType>(descriptor, inout.data(), no_dependencies);
+    done.wait();
 
     if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
         for (std::size_t j = 0; j < real_first_dims; j++) {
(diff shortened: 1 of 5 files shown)


hjabird (commented):
The code itself looks good, but I don't understand the need to move from `return test_skipped` to `GTEST_SKIP`. 

The original use of `test_skipped` seems more consistent with the tests from the other backends (eg. [here](https://github.com/oneapi-src/oneMKL/blob/4c7e7eb7227bf131d6326d3dd84f2bfc57ecf3b9/tests/unit_tests/blas/level1/axpby.cpp#L117)). Can you comment on why `GTEST_SKIP` is needed in something that seems otherwise functionally equivalent?

FMarno:
[mklgpu_run.log](https://github.com/oneapi-src/oneMKL/files/11365336/mklgpu_run.log)

FMarno:
> The code itself looks good, but I don't understand the need to move from `return test_skipped` to `GTEST_SKIP`.
> 
> The original use of `test_skipped` seems more consistent with the tests from the other backends (eg. [here](https://github.com/oneapi-src/oneMKL/blob/4c7e7eb7227bf131d6326d3dd84f2bfc57ecf3b9/tests/unit_tests/blas/level1/axpby.cpp#L117)). Can you comment on why `GTEST_SKIP` is needed in something that seems otherwise functionally equivalent?

`test_skipped` and `EXPECT_TRUEORSKIP` are oneMKL specific and will internally call `GTEST_SKIP`. Since the `catch` occurs at the top level of the test I would have to do `EXPECT_TRUEORSKIP(test_skipped)` but I just went straight to the `GTEST_SKIP`.

hjabird (approved):

tests/unit_tests/dft/include/compute_inplace.hpp
```diff
@@ -231,16 +219,10 @@ int DFT_Test<precision, domain>::test_in_place_USM() {
...
-            .wait();
-    }
-    catch (oneapi::mkl::unimplemented& e) {
-        std::cout << "Skipping test because: \"" << e.what() << "\"" << std::endl;
-        return test_skipped;
-    }
+    std::vector<sycl::event> no_dependencies;
+    oneapi::mkl::dft::compute_forward<descriptor_t, FwdInputType>(descriptor, inout.data(),
+                                                                  no_dependencies)
+        .wait();
```
Rbiessy: Now could be a good time to switch to `wait_and_throw()` in all of our tests. This allows us to catch asynchronous errors so this is best practice in SYCL. What do you think?

tests/unit_tests/dft/source/compute_tests.cpp
```diff
@@ -50,12 +50,18 @@ class ComputeTests_real_real_out_of_place
...
-                std::get<1>(GetParam()).batches                                                  \
-            };                                                                                   \
-        EXPECT_TRUEORSKIP(test.test_##PLACE##_##LAYOUT##STORAGE());                              \
+        try {                                                                                    \
+            auto test =                                                                          \
+                DFT_Test<oneapi::mkl::dft::precision::PRECISION,                                 \
+                         oneapi::mkl::dft::domain::DOMAIN>{ std::get<0>(GetParam()),             \
+                                                            std::get<1>(GetParam()).sizes,       \
+                                                            std::get<1>(GetParam()).batches };   \
+            EXPECT_TRUE(test.test_##PLACE##_##LAYOUT##STORAGE());                                \
```
Rbiessy: We are still returning `test_skipped` in the cases USM or `double` are not supported so this should still be `EXPECT_TRUEORSKIP`.

tests/unit_tests/dft/include/compute_inplace.hpp
FMarno: Sounds like a good idea! done in https://github.com/oneapi-src/oneMKL/pull/311/commits/0f0f28745a69f38dfa057fb159a45f98ef2ea6b3

tests/unit_tests/dft/source/compute_tests.cpp
FMarno: That's true. I think I missed this because the tests were passing instead. Fixed in https://github.com/oneapi-src/oneMKL/pull/311/commits/0f0f28745a69f38dfa057fb159a45f98ef2ea6b3.

FMarno:
[mklgpu_run.log](https://github.com/oneapi-src/oneMKL/files/11374701/mklgpu_run.log)
Updated log

Rbiessy (approved):

anantsrivastava30:
Do you also think you can move 
``` c++
    if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
        std::cout << "skipping real split tests as they are not supported" << std::endl;
        return test_skipped;
    }
```
this bit from the real_real test and skip the REAL domain from the `compute_tests.cpp` as a part of this PR

anantsrivastava30:
Please attach the CPU logs as well

FMarno:
> Do you also think you can move
> 
> ```c++
>     if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
>         std::cout << "skipping real split tests as they are not supported" << std::endl;
>         return test_skipped;
>     }
> ```
> 
> this bit from the real_real test and skip the REAL domain from the `compute_tests.cpp` as a part of this PR

I would rather save that for a separate PR, I need to convince myself that its impossible for that to be implemented. Also I see it as a separate issue. 
I did have the same thought though so most of the work is done here https://github.com/oneapi-src/oneMKL/compare/develop...FMarno:oneMKL:no_real_real_real2complex?expand=1

FMarno:
> Please attach the CPU logs as well

[mklcpu_run.log](https://github.com/oneapi-src/oneMKL/files/11375891/mklcpu_run.log)
mklcpu log

anantsrivastava30:
this time around the logs have this : 
```
977/977 Test #977: dft/EXAMPLE/CT/complex_fwd_buffer_mklcpu .........................................................................................................................................   Passed    0.30 sec

100% tests passed, 0 tests failed out of 977

Total Test time (real) = 260.39 sec

The following tests did not run:
	140 - DFT/RT/ComputeTestSuite/ComputeTests_real_real_in_place.REAL_SINGLE_in_place_real_real_buffer/sizes_8_batches_1_11th_Gen_Intel_R__Core_TM__i9_11900K___3_50GHz (Skipped)
...
...
```
but previously the behavior was to show these as passed but in the full logs (sample is from running `ctest` as well 
```
1937 968/968 Test #1943: DFT/CT/DescriptorCommitTestSuite/DescriptorCommitTests.DescriptorCommitTestsComplexDouble/Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz .........................................   Passed    0.81 sec
1938
1939 100% tests passed, 0 tests failed out of 968
<EOF>
```
old LastTest.log:
```
 3660 [ RUN      ] ComputeTestSuite/ComputeTests_real_real_in_place.REAL_SINGLE_in_place_real_real_buffer/sizes_8_batches_1_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz
 3661 skipping real split tests as they are not supported
 3662 [  SKIPPED ] ComputeTestSuite/ComputeTests_real_real_in_place.REAL_SINGLE_in_place_real_real_buffer/sizes_8_batches_1_Intel_R__Core_TM__i7_6770HQ_CPU___2_60GHz (78 ms)
 3663 [----------] 1 test from ComputeTestSuite/ComputeTests_real_real_in_place (78 ms total)
 3664
 3665 [----------] Global test environment tear-down
 3666 [==========] 1 test from 1 test suite ran. (78 ms total)
 3667 [  PASSED  ] 0 tests.
``` 
what attributes this change in behavior ?

FMarno:
> 

The test log is from a run of ctest. the example you gave will have come from directly running `test_main_dft_ct` or `test_main_dft_rt`

FMarno:
@anantsrivastava30 are you satisfied with this answer? Is there anything you would like clarified/changed for this to be approved?

anantsrivastava30:
> @anantsrivastava30 are you satisfied with this answer? Is there anything you would like clarified/changed for this to be approved?

I updated my initial question to indicate that I am running tests using `ctest` as well, and I see this huge tail at the end of every ctest run telling me that : these 100 odd tests did not run, I would have preferred that information to stay in the logs.

I however do not see that as a huge concern and every thing else looks good.

anantsrivastava30 (approved):
Approved

FMarno:
@anantsrivastava30 Thanks for clarifying. Unfortunately I don't see the same behaviour on develop (sha 4c7e7eb7227bf131d6326d3dd84f2bfc57ecf3b9). I always see 977 tests running with the real to complex "real_real" tests being skipped.

PR #320, opened 2023-05-16
Fix clang-format

mkrainiuk:
# Description

PR fixes some clang-format problems introduced by the last several commits.

# Checklist

## All Submissions

### DIFF
--- a/src/blas/backends/syclblas/syclblas_common.hpp
+++ b/src/blas/backends/syclblas/syclblas_common.hpp
@@ -141,7 +141,7 @@ inline auto convert_to_syclblas_type(ArgT... args) {
 **/
 template <typename CheckT, sycl::aspect AspectVal>
 struct throw_if_unsupported_by_device {
-/** Operator to throw if unsupported.
+    /** Operator to throw if unsupported.
  * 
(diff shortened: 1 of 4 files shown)


src/blas/backends/syclblas/syclblas_common.hpp
```diff
@@ -141,7 +141,7 @@ inline auto convert_to_syclblas_type(ArgT... args) {
 **/
 template <typename CheckT, sycl::aspect AspectVal>
 struct throw_if_unsupported_by_device {
-/** Operator to throw if unsupported.
+    /** Operator to throw if unsupported.
```
andrewtbarker, line 144: I don't really like this format change, maybe it's worth it for clang-format not to complain.

andrewtbarker (approved):
Thanks for cleaning this up - some of the mess is my fault.

akabalov (approved):

src/blas/backends/syclblas/syclblas_common.hpp
mkrainiuk: Agree, this one looks not correct, I think we can merge it to stop clang-format complaining about it and fix the clang-format at some point.

PR #329, opened 2023-05-25
[DFT] add backend_selector to compile-time dispatch compute_xxxward

### LINKED ISSUE #309: [DFT] Building multiple DFT backends at once causes linking conflicts
# Summary

# Version

# Steps to reproduce

# Observed behavior
```
...
```
# Expected behavior

FMarno:
# Description
Add a `backend_selector` argument to all the backend specific `compute_xxxward` functions. This is only there to give the function a unique signature, to avoid linking conflicts.

Fixes #309

# Checklist

## All Submissions

## New interfaces

it was accepted? # (RFC)

## New features

## Bug fixes

      GitHub issue or in this PR)?

### DIFF
--- a/examples/dft/compile_time_dispatching/CMakeLists.txt
+++ b/examples/dft/compile_time_dispatching/CMakeLists.txt
@@ -45,3 +45,4 @@ foreach(dfti_backend ${DFTI_CT_BACKENDS})
   # Register example as ctest
   add_test(NAME dft/EXAMPLE/CT/complex_fwd_buffer_${dfti_backend} COMMAND ${EXAMPLE_NAME})
 endforeach(dfti_backend)
+
--- a/examples/dft/compile_time_dispatching/CMakeLists.txt
+++ b/examples/dft/compile_time_dispatching/CMakeLists.txt
@@ -18,46 +18,30 @@
 #===============================================================================
 
 #Build object from all sources
-set(DFTI_CT_SOURCES "")
+set(DFTI_CT_BACKENDS "")
 
 if(ENABLE_MKLGPU_BACKEND)
-  list(APPEND DFTI_CT_SOURCES "complex_fwd_buffer_mklgpu")
+  list(APPEND DFTI_CT_BACKENDS "mklgpu")
 endif()
 
 if(ENABLE_MKLCPU_BACKEND)
-  list(APPEND DFTI_CT_SOURCES "complex_fwd_buffer_mklcpu")
+  list(APPEND DFTI_CT_BACKENDS "mklcpu")
 endif()
 
 include(WarningsUtils)
 
-foreach(dfti_ct_sources ${DFTI_CT_SOURCES})
-  # add executable and define include directories
-  # add dependencies and link libraries
-  # register example as ctest
- add_executable(example_${domain}_${dfti_ct_sources} ${dfti_ct_sources}.cpp)
-  target_include_directories(example_${domain}_${dfti_ct_sources}
+foreach(dfti_backend ${DFTI_CT_BACKENDS})
+  set(EXAMPLE_NAME example_dft_complex_fwd_buffer_${dfti_backend})
+  add_executable(${EXAMPLE_NAME} complex_fwd_buffer_${dfti_backend}.cpp)
+  target_include_directories(${EXAMPLE_NAME}
       PUBLIC ${PROJECT_SOURCE_DIR}/examples/include
       PUBLIC ${PROJECT_SOURCE_DIR}/include
       PUBLIC ${CMAKE_BINARY_DIR}/bin
   )
 
-  set(ONEMKL_LIBRARIES_${domain} "")
-  if(domain STREQUAL "dft")
-    if(dfti_ct_sources MATCHES "_mklcpu$")
-      add_dependencies(example_${domain}_${dfti_ct_sources} onemkl_${domain}_mklcpu)
-      list(APPEND ONEMKL_LIBRARIES_${domain} onemkl_${domain}_mklcpu)
-    endif()
-    if(dfti_ct_sources MATCHES "_mklgpu$")
-      add_dependencies(example_${domain}_${dfti_ct_sources} onemkl_${domain}_mklgpu)
-      list(APPEND ONEMKL_LIBRARIES_${domain} onemkl_${domain}_mklgpu)
-    endif() 
-  endif()
+  add_dependencies(${EXAMPLE_NAME} onemkl_dft_${dfti_backend})
+  target_link_libraries(${EXAMPLE_NAME} PRIVATE ONEMKL::SYCL::SYCL onemkl_dft_${dfti_backend})
 
-  target_link_libraries(example_${domain}_${dfti_ct_sources}
-      PUBLIC ${ONEMKL_LIBRARIES_${domain}}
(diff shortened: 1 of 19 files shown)


examples/dft/compile_time_dispatching/CMakeLists.txt
```diff
@@ -18,46 +18,30 @@
...
-      PUBLIC ${ONEMKL_LIBRARIES_${domain}}
-      ${ONEMKL_LIBRARIES_${domain}}
-      ONEMKL::SYCL::SYCL
-  )
   # Register example as ctest
- add_test(NAME ${domain}/EXAMPLE/CT/${dfti_ct_sources} COMMAND example_${domain}_${dfti_ct_sources})
-endforeach(dfti_ct_sources)
\ No newline at end of file
+  add_test(NAME dft/EXAMPLE/CT/complex_fwd_buffer_${dfti_backend} COMMAND ${EXAMPLE_NAME})
+endforeach(dfti_backend)
```
Rbiessy: Nitpicking, could add a new line.

FMarno:
test run - [mklcpu_mklgpu_test_run.log](https://github.com/oneapi-src/oneMKL/files/11569452/mklcpu_mklgpu_test_run.log)

tests/unit_tests/dft/source/descriptor_tests.cpp
```diff
@@ -431,8 +431,7 @@ inline void recommit_values(sycl::queue& sycl_queue) {
     std::vector<test_params> argument_groups{
         // not changeable
         // FORWARD_DOMAIN, PRECISION, DIMENSION, COMMIT_STATUS
-        { std::make_pair(config_param::NUMBER_OF_TRANSFORMS, std::int64_t{ 5 }),
```
Rbiessy, line 434: Why is this needed?

tests/unit_tests/dft/include/test_common.hpp
```diff
@@ -162,6 +162,17 @@ void commit_descriptor(oneapi::mkl::dft::descriptor<precision, domain> &descript
...
 }
 
+template <typename Func, typename... Args>
+void dispatch(sycl::queue queue, Func func, Args &&... args) {
+#ifdef CALL_RT_API
+    (void)queue;
+    func(std::forward<Args>(args)...);
+#else
+    // can't get return from this, so func must capture return values by reference in a lambda
+    TEST_RUN_CT_SELECT(queue, func, std::forward<Args>(args)...);
```
Rbiessy, line 172: Is this working for the dft domain? We've had to introduce `TEST_RUN_CT_SELECT_NO_ARGS` to use cufft for instance.

src/dft/backends/cufft/CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ target_link_libraries(${LIB_OBJ} PUBLIC ONEMKL::SYCL::SYCL ${MKL_LINK_SYCL})
 set_target_properties(${LIB_OBJ} PROPERTIES
   POSITION_INDEPENDENT_CODE ON
 )
-target_link_libraries(${LIB_NAME} PUBLIC ${LIB_OBJ})
+target_link_libraries(${LIB_NAME} PUBLIC ${LIB_OBJ} onemkl_dft)
```
Rbiessy, line 51: It makes sense to me that we add `onemkl_dft` here but it does not seem to match with other domains. I'm not sure where else `onemkl_dft` was linked before.

tests/unit_tests/CMakeLists.txt
```diff
@@ -182,6 +182,9 @@ foreach(domain ${TARGET_DOMAINS})
       PROPERTIES TEST_PREFIX ${DOMAIN_PREFIX}/RT/
       DISCOVERY_TIMEOUT 30
     )
+
+    # ensure rt tests have up-to-date libs
+    add_dependencies(test_main_${domain}_rt ${ONEMKL_LIBRARIES_${domain}})
```
Rbiessy, line 187: In most cases `add_dependencies` shouldn't be needed if we use `target_link_libraries`. I can see the compile tests above use `add_dependencies` but I'm not sure why, do you know if there is a good reason for it?

src/dft/backends/cufft/CMakeLists.txt
FMarno: `onemkl_dft` couldn't be linked before because linking with both `onemkl_dft` and with `onemkl_dft_xxxxx` would cause multiple definitions of the `compute_xxxward` functions.
WRT the other domains, I'm not sure how that works. I'm adding `onemkl_dft` here so users get everything by just linking `onemkl_dft_xxxxx`, maintaining the current functionality. Personally I think users should specify both libraries when linking since there will potentially be multiple definitions for some functions e.g. if you tried to link both `onemkl_dft_mklcpu` and `onemkl_dft_mklgpu`.

tests/unit_tests/dft/source/descriptor_tests.cpp
FMarno: mklcpu considers it an invalid configuration if `NUMBER_OF_TRANSFORMS != 1 && XWD_DISTANCE < N`

tests/unit_tests/dft/include/test_common.hpp
FMarno: I added the `TEST_RUN_CT_SELECT` previously, so it works fine https://github.com/oneapi-src/oneMKL/blob/055b3701c0b1eeb7a3c33866cd9239f0ddb1015c/tests/unit_tests/include/test_helper.hpp#L175
I can post a cufft log as well

tests/unit_tests/CMakeLists.txt
FMarno: I'm not sure why they have done that for the compile-time tests. 
For this change I had a few instances were I would update the library and then try to build only the test and I would be confused why nothing changed! Is there a better way I should be doing this?

examples/dft/compile_time_dispatching/CMakeLists.txt
FMarno: addressed in https://github.com/oneapi-src/oneMKL/pull/329/commits/80e5213fe058f8ef84710d1467996f2a58459ff9

tests/unit_tests/dft/include/test_common.hpp
Rbiessy: Thanks, my bad I missed it.

FMarno:
[cufft_run.log](https://github.com/oneapi-src/oneMKL/files/11626502/cufft_run.log)

tests/unit_tests/CMakeLists.txt
Rbiessy: I am surprised that `add_dependencies` is needed as I would expect `test_main_${domain}_rt` is already linked with `${ONEMKL_LIBRARIES_${domain}}` somehow so I am wondering if we are missing some dependencies elsewhere for the other domains.

Rbiessy:
@mkrainiuk could you or someone else help with the review? I have left 2 comments open as I think there are some things we don't understand in the CMake today. Hopefully someone can help with that.

FMarno:
I've converted this to a draft since we aren't sure that this is the direction we would like to take.

FMarno:
problem was fixed with #342 instead

PR #331, opened 2023-06-13
[DFT] Remove comptime execution from DFT domain

### LINKED ISSUE #309: [DFT] Building multiple DFT backends at once causes linking conflicts

FMarno:
# Description

Due to the current design of the compile-time execution in the DFT domain, there can be redefinitions of the functions `compute_forward` and `compute_backward`, leading to unexpected behaviour when linking. This change removes the compile-time dispatch for `compute_forward` and `compute_backward` and also aligns the design more with the RNG domain. `commit` is not removed since it does not suffer from the same issues (due to it having a `backend_selector` argument).
The issue is discussed further in #309, and this patch is an alternative to #329.

Fixes #309

# Checklist

## All Submissions

## Bug fixes

      GitHub issue or in this PR)?


FMarno:
MKLGPU and MKLCPU run - 
[mklgpu_mklcpu_run.txt](https://github.com/oneapi-src/oneMKL/files/11733008/mklgpu_mklcpu_run.txt)

tests/unit_tests/include/test_helper.hpp
```diff
@@ -172,6 +172,7 @@
                 TEST_RUN_NVIDIAGPU_CUBLAS_SELECT(q, func, __VA_ARGS__);    \
```
hjabird: Why are the changes in `test_helper.hpp` needed as part of this PR?

tests/unit_tests/dft/source/descriptor_tests.cpp
```diff
@@ -431,8 +431,7 @@ inline void recommit_values(sycl::queue& sycl_queue) {
     std::vector<test_params> argument_groups{
         // not changeable
         // FORWARD_DOMAIN, PRECISION, DIMENSION, COMMIT_STATUS
-        { std::make_pair(config_param::NUMBER_OF_TRANSFORMS, std::int64_t{ 5 }),
```
hjabird, line 434: Why is this being deleted?

tests/unit_tests/dft/source/descriptor_tests.cpp
```diff
@@ -538,8 +537,9 @@ inline void swap_out_dead_queue(sycl::queue& sycl_queue) {
     auto inout = sycl::malloc_device<forward_type>(default_1d_lengths + 2, sycl_queue);
     sycl_queue.wait();
 
-    auto transform_event = oneapi::mkl::dft::compute_forward<decltype(descriptor), forward_type>(
-        descriptor, inout, std::vector<sycl::event>{});
```
hjabird: Why did this need to change?

tests/unit_tests/dft/source/descriptor_tests.cpp
FMarno: When number of transforms is > 1, mklcpu expects the FWD_DISTANCE and BWD_DISTANCE to be set appropriately, so this test was throwing an error saying "inconsistent configuration".

tests/unit_tests/include/test_helper.hpp
FMarno: One of the changes wasn't helpful so I've removed it. The other changes removes a warning. Updated in https://github.com/oneapi-src/oneMKL/pull/331/commits/cccea27fa808f6d506ba3ab8b42072b587b7eb5d

tests/unit_tests/dft/source/descriptor_tests.cpp
FMarno: reverted in https://github.com/oneapi-src/oneMKL/pull/331/commits/cccea27fa808f6d506ba3ab8b42072b587b7eb5d

src/dft/backends/mklcpu/CMakeLists.txt
```diff
@@ -60,7 +59,7 @@ endif()
 set_target_properties(${LIB_OBJ} PROPERTIES
   POSITION_INDEPENDENT_CODE ON
 )
-target_link_libraries(${LIB_NAME} PUBLIC ${LIB_OBJ})
+target_link_libraries(${LIB_NAME} PUBLIC ${LIB_OBJ} onemkl_dft)
```
Rbiessy, line 62: From my current understanding of the CMake `onemkl_dft` should be linked to the tests and examples only to match with the other domains.

Rbiessy (commented):
Thank you for this it looks good to me overall! Also could you clarify does this remove or affect any exported symbols? From what I can see these are only internal changes.

FMarno:
> Thank you for this it looks good to me overall! Also could you clarify does this remove or affect any exported symbols? From what I can see these are only internal changes.

If I've got this all correct then the "function table" object in the dynamic library should be a lot smaller (no more compute_forward/backward), the `onemkl_dft_*****` libraries should no long have unique definitions of oneapi::mkl::dft::compute_forward/backward, but should export the ones from `onemkl_dft` (but I see you are suggesting removing that), and the `commit_impl` "interface" looks different but that shouldn't be exposed to the user.

src/dft/backends/mklcpu/CMakeLists.txt
FMarno: Sorry, I'm having trouble seeing where this happens in the project. Saying that, I also don't see how the other domains are linking the runtime stuff at all.

FMarno: ~~I think that RNG might be doing their runtime dispatch in the headers~~ That's probably not completely true, but a lot of the work is happening in the headers.

Rbiessy: You're right, my bad. That looks good to me so let's keep it here!

Rbiessy (approved):

src/dft/backends/mklcpu/CMakeLists.txt
Rbiessy: I think the `onemkl_${domain}` libraries may not be linked at all but opened in `dlopen` instead.

FMarno: I see! I think I know what I need to do but this is a little awkward with templates :D

FMarno:
Closing in favour of #342

PR #342, opened 2023-07-04
[DFT] remove compile-time compute dispatch from dft domain

### LINKED ISSUE #309: [DFT] Building multiple DFT backends at once causes linking conflicts

FMarno:
# Description

The current design of compile-time dispatch for the DFT domain causes multiple definitions for the `compute_forward/backward` functions, leading to linking the wrong symbols and undesirable behaviour. This PR removes the compile-time dispatch functions so this is no longer an issue.

This PR also attempts to simplify the library in a couple of ways:
* removing many of the internal implementations of `compute_forward/backward`, moving the complexity to a templated header which will cast the types.
* removing `compute_forward/backward` from the function table. These are called via a virtual function on `commit_impl` now.

Fixes #309

# Checklist

## All Submissions

## Bug fixes

- [x] Have you added relevant regression tests? - N/A
      GitHub issue or in this PR)?


FMarno:
test logs:
[mklcpu_mklgpu_run.txt](https://github.com/oneapi-src/oneMKL/files/11948487/mklcpu_mklgpu_run.txt)
[cufft_run.log](https://github.com/oneapi-src/oneMKL/files/11948495/cufft_run.log)
[rocfft_run.log](https://github.com/oneapi-src/oneMKL/files/11948496/rocfft_run.log)

Rbiessy (approved):
That looks fine to me. I just want to clarify in this PR that this does not change the interface so the `compile_time_dispatching` examples still run like before but the backends are not truly dispatched at compile time, just like the RNG domain.

FMarno:
> That looks fine to me. I just want to clarify in this PR that this does not change the interface so the `compile_time_dispatching` examples still run like before but the backends are not truly dispatched at compile time, just like the RNG domain.

That's correct. Only the commit is compile-time in the compile-time tests.

include/oneapi/mkl/dft/backward.hpp
```diff
@@ -26,51 +26,105 @@
...
+
 namespace oneapi::mkl::dft {
 //Buffer version
 
 //In-place transform
 template <typename descriptor_type, typename data_type>
-void compute_backward(descriptor_type &desc, sycl::buffer<data_type, 1> &inout);
+void compute_backward(descriptor_type &desc, sycl::buffer<data_type, 1> &inout) {
+    using fwd_type = typename detail::descriptor_info<descriptor_type>::forward_type;
+    auto type_corrected_inout = inout.template reinterpret<fwd_type, 1>(
```
hjabird, line 41: In quite a few places, buffers are reinterpreted to a "correct" type without checking the validity of the input types. I think more type checking is needed here. Eg.
* A user might be lead to expect that eg. passing single-precision data to a double precision transform is a supported operation.
* A user might want errors on passing in eg. integers.

include/oneapi/mkl/dft/detail/commit_impl.hpp
```diff
@@ -26,42 +26,88 @@
...
 
     // rule of three
-    commit_impl(const commit_impl& other) = delete;
-    commit_impl& operator=(const commit_impl& other) = delete;
+    commit_impl(const commit_impl &other) = delete;
+    commit_impl &operator=(const commit_impl &other) = delete;
     virtual ~commit_impl() = default;
 
-    sycl::queue& get_queue() noexcept {
+    sycl::queue &get_queue() noexcept {
```
hjabird, line 58: There are formatting changes - was the original incorrectly formatted?

lhuot:
@anantsrivastava30 could you please take a look at this PR? Thanks!

include/oneapi/mkl/dft/detail/dft_ct.hxx
```diff
@@ -25,96 +25,114 @@ ONEMKL_EXPORT dft::detail::commit_impl<prec, dom> *create_commit(
...
+using scalar = typename detail::descriptor_info<descriptor_type>::scalar_type;
+template <typename descriptor_type>
+using fwd = typename detail::descriptor_info<descriptor_type>::forward_type;
+template <typename descriptor_type>
+using bwd = typename detail::descriptor_info<descriptor_type>::backward_type;
+
 //In-place transform
-template <typename descriptor_type, typename data_type>
-ONEMKL_EXPORT void compute_forward(descriptor_type &desc, sycl::buffer<data_type, 1> &inout);
+template <typename descriptor_type>
```
hjabird, line 36: Are all the function declarations in this file still needed - aren't they replaced by the virtual methods in the commit impl?

include/oneapi/mkl/dft/detail/types_impl.hpp
```diff
@@ -39,6 +39,60 @@ struct precision_t {
...
+
+template <precision prec, domain dom>
+class descriptor;
+
+template <class... T>
+constexpr bool always_false = false;
+
+template <typename descriptor_type>
+struct descriptor_info {
+    static_assert(always_false<descriptor_type>, "Not a valid descriptor type");
```
hjabird, line 62: Yes, but won't there be an error anyway when `scalar_type`, `forward_type` and `backward_type` aren't defined? I think we should this and the `always_false` since `descriptor_info` will only ever be oneMKL developer-facing.

include/oneapi/mkl/dft/detail/types_impl.hpp
```diff
@@ -39,6 +39,60 @@ struct precision_t {
...
+template <>
+struct descriptor_info<descriptor<precision::DOUBLE, domain::COMPLEX>> {
+    using scalar_type = double;
+    using forward_type = std::complex<double>;
+    using backward_type = std::complex<double>;
+};
+
+// compute the range of a reinterpreted buffer
+template <typename In, typename Out>
+std::size_t reinterpret_range(std::size_t size) {
```
hjabird, line 102: Throw on invalid size argument?
 
Eg `In` is `float`, `Out` is `complex<float>`, `size` is odd.

include/oneapi/mkl/dft/detail/commit_impl.hpp
```diff
@@ -26,42 +26,88 @@
...
-    virtual void* get_handle() noexcept = 0;
+    virtual void *get_handle() noexcept = 0;
 
-    virtual void commit(const dft_values<prec, dom>&) = 0;
+    virtual void commit(const dft_values<prec, dom> &) = 0;
 
-private:
-    sycl::queue queue_;
-    mkl::backend backend_;
+    virtual void forward_ip_cc(descriptor_type &desc, sycl::buffer<fwd_type, 1> &inout) = 0;
```
hjabird, line 70: I don't see where these new functions are overrided for the derived commit impls. What am I missing?

include/oneapi/mkl/dft/detail/commit_impl.hpp
FMarno: I think so. I've run clang-format-10 and `clang-format-11` over it and that how it wanted to be.

include/oneapi/mkl/dft/detail/dft_ct.hxx
FMarno: These are the ones in the `oneapi::mkl::dft::BACKEND` namespace and the virtual methods all just call the corresponding method (in `src/dft/backends/backend_compute_signature.cxx`), so yes I could just move the implementation to the `commit_impl` classes.
Maintaining the backend specific free functions does have the benefit of easily adding the compile-time dispatch in the future, but maybe things would be simpler without it. (and YAGNI principle, but the work is already done, and this is a smaller change(maybe)).
At the very least I can look at removing the `compute_forward/compute_backward` declarations from this file, since I don't think that needs to be user facing any more.

include/oneapi/mkl/dft/detail/types_impl.hpp
FMarno: There will be, but this is a good error in my opinion. e.g. I updated on of the tests to use `descriptor_t&` as the first template argument and part of the output was 
```
finlay@devpc-c310:~/perf/oneMKL/build$ ninja | head
[1/4] Building CXX object tests/unit_tests/dft/source/CMakeFiles/dft_source_rt.dir/compute_tests.cpp.o
FAILED: tests/unit_tests/dft/source/CMakeFiles/dft_source_rt.dir/compute_tests.cpp.o 
/home/finlay/intel/oneapi/compiler/2023.1.0/linux/bin/icpx  -I/home/finlay/perf/oneMKL/tests/unit_tests/dft/source/../include -I/home/finlay/perf/oneMKL/tests/unit_tests/dft/source/../../include -I/home/finlay/perf/oneMKL/include -I/home/finlay/perf/oneMKL/deps/googletest/include -I/home/finlay/perf/oneMKL/build/bin -O3 -DNDEBUG -std=gnu++17 -DCALL_RT_API -DNOMINMAX -fsycl -Wall -Wextra -Wshadow -Wconversion -Wpedantic -MD -MT tests/unit_tests/dft/source/CMakeFiles/dft_source_rt.dir/compute_tests.cpp.o -MF tests/unit_tests/dft/source/CMakeFiles/dft_source_rt.dir/compute_tests.cpp.o.d -o tests/unit_tests/dft/source/CMakeFiles/dft_source_rt.dir/compute_tests.cpp.o -c /home/finlay/perf/oneMKL/tests/unit_tests/dft/source/compute_tests.cpp
In file included from /home/finlay/perf/oneMKL/tests/unit_tests/dft/source/compute_tests.cpp:29:
In file included from /home/finlay/perf/oneMKL/tests/unit_tests/dft/source/../../include/test_helper.hpp:33:
In file included from /home/finlay/perf/oneMKL/include/oneapi/mkl.hpp:26:
In file included from /home/finlay/perf/oneMKL/include/oneapi/mkl/dft.hpp:33:
In file included from /home/finlay/perf/oneMKL/include/oneapi/mkl/dft/detail/dft_loader.hpp:30:
/home/finlay/perf/oneMKL/include/oneapi/mkl/dft/detail/types_impl.hpp:55:5: error: static assertion failed due to requirement 'always_false<oneapi::mkl::dft::detail::descriptor<oneapi::mkl::dft::detail::precision::SINGLE, oneapi::mkl::dft::detail::domain::COMPLEX> &>': Not a valid descriptor type
    static_assert(always_false<descriptor_type>, "Not a valid descriptor type");
```
potentially I could replace this with a static assert at the user-facing call site though.

include/oneapi/mkl/dft/detail/commit_impl.hpp
FMarno: src/dft/backends/backend_compute_signature.cxx

include/oneapi/mkl/dft/backward.hpp
FMarno: added in https://github.com/oneapi-src/oneMKL/pull/342/commits/a8a004c57696f512cc9f9e54aab3aaea0bedd536

include/oneapi/mkl/dft/detail/types_impl.hpp
FMarno: I'd like to keep this. When the user specifies the wrong template argument for the descriptor, it will give them a more plain error.

include/oneapi/mkl/dft/detail/types_impl.hpp
FMarno: added in https://github.com/oneapi-src/oneMKL/pull/342/commits/a8a004c57696f512cc9f9e54aab3aaea0bedd536

hjabird (approved):

Rbiessy (approved):

include/oneapi/mkl/dft/backward.hpp
```diff
@@ -26,51 +26,137 @@
...
-                      sycl::buffer<output_type, 1> &out);
+                      sycl::buffer<output_type, 1> &out) {
+    static_assert(detail::valid_compute_arg<descriptor_type, input_type>::value,
+                  "unexpected type for input_type");
+    static_assert(detail::valid_compute_arg<descriptor_type, output_type>::value,
+                  "unexpected type for output_type");
+
+    using fwd_type = typename detail::descriptor_info<descriptor_type>::forward_type;
+    using bwd_type = typename detail::descriptor_info<descriptor_type>::backward_type;
+    auto type_corrected_in = in.template reinterpret<bwd_type, 1>(
```
anantsrivastava30, line 72: whats the benifit of converting the `input_type` to 'scalar_type' which is strictly only a 'float' or 'double'.Is it because you want to supprot not only scalar and complex in/out_type but sycl::float2/4 double2/4, or some nuance of the rocFFT/cuFFT ?

anantsrivastava30 (approved):
I approve of this change, pending a answers to one question I had. Otherwise looks like a good solution to me. Thanks for fixing this.

include/oneapi/mkl/dft/backward.hpp
FMarno: I'm not completely sure I understand your question but I'll give you some info that might clarify things.
The user facing functions `compute_forward/compute_backward` currently have no restrictions on the types of `input_type` and `output_type` (though I have cheekily added some sensible ones in a static assert!). Where the functions are not user facing, I've made the parameter types purely dependent on the descriptor type, with (hopefully) sensible types chosen in the [`descriptor_info`](https://github.com/oneapi-src/oneMKL/blob/4905064a608aa89e47e152e0d0480239b5f06251/include/oneapi/mkl/dft/detail/types_impl.hpp#L61) struct.
The casts that happen in `include/oneapi/mkl/dft/backward.hpp` and `include/oneapi/mkl/dft/forward.hpp` are there to reinterpret the type that the user has chosen to the internal types.

FMarno:
Here is an up-to-date test log.
[mklcpu_mklgpu_run.log](https://github.com/oneapi-src/oneMKL/files/12145026/mklcpu_mklgpu_run.log)
@anantsrivastava30 I will wait for you to resolve your comment before I merge

include/oneapi/mkl/dft/backward.hpp
anantsrivastava30: Well what I mean is this reinterpret_type converts any valid type that passes the assertions just above to its fundamental type, i.e float or double, so my question is do we really need that ? Because in the sycl API we support float/double/complex<float/double> input, and in the C_API also we accept any of the former. So what exactly are we getting out of this reinterpretation?

FMarno: The type is not just reinterpreted to the scalar type, it is sometimes the `forward_type` or `backward_type` from `descriptor_info`.

I think it has a few benefits.
1. greater clarity about the expected data layout
2. reduced number of functions we need to expose from the library
3. reduced argument ambiguity (REAL_REAL in-place and COMPLEX_COMPLEX out-of-place have the same number of arguments, which can be confusing. With this system they at least take different arguments.)

Most importantly, the functions in `backward.hpp` call virtual functions on `commit_impl`, so those can't have a templated argument, so we either reinterpret or have ~100 virtual functions on the `commit_impl` class.

anantsrivastava30:
@FMarno I have resolved my comment, please feel free to merge this.

FMarno:
new test run after rebase [mklcpu_mklgpu_run.log](https://github.com/oneapi-src/oneMKL/files/12182384/mklcpu_mklgpu_run.log)

PR #347, opened 2023-07-11
[DFT] read only accessors return const reference with get_pointer()

anantsrivastava30:
# Description
sycl 2020 spec states:
Member functions of accessor which return a reference to an element have been changed to return a const reference for read-only accessors. The get_pointer() member function has also been changed to return a const pointer for read-only accessors. The value_type and reference member types of accessor have been changed to be const types for read-only accessors.

This change is now apparent from the 20230616_nightly compiler , and in the first CEV testing we see build failures.
```
/export/users/mkltest/cache/mkl/develop/20230705_cev_nightly/__release_lnx/mkl/include/mkl_dfti.h:241:22: note: candidate function not viable: no known conversion from 'std::add_pointer_t<value_type>' (aka 'const float *') to 'void *' for 2nd argument
 241 | DFTI_EXTERN MKL_LONG DftiComputeForward(DFTI_DESCRIPTOR_HANDLE, void*, ...);
```
Fixes # (GitHub issue)

# Checklist
- [x] local build passes with and beyond 20230616_nightly compiler
- [x] local cpu tests pass

## All Submissions

## New interfaces

it was accepted? # (RFC)

## New features

## Bug fixes

      GitHub issue or in this PR)?

### DIFF
--- a/src/dft/backends/mklcpu/backward.cpp
+++ b/src/dft/backends/mklcpu/backward.cpp
@@ -152,11 +152,12 @@ ONEMKL_EXPORT void compute_backward(descriptor_type &desc, sycl::buffer<input_ty
 
     cpu_queue.submit([&](sycl::handler &cgh) {
         auto desc_acc = mklcpu_desc_buffer.template get_access<sycl::access::mode::read>(cgh);
-        auto in_acc = in.template get_access<sycl::access::mode::read_write>(cgh);
+        auto in_acc = in.template get_access<sycl::access::mode::read>(cgh);
         auto out_acc = out.template get_access<sycl::access::mode::write>(cgh);
 
         detail::host_task<class host_kernel_back_outofplace>(cgh, [=]() {
-            DFT_ERROR status = DftiComputeBackward(desc_acc[detail::DIR::bwd], in_acc.get_pointer(),
+            auto inPtr = const_cast<input_type*>(in_acc.get_pointer());
+            DFT_ERROR status = DftiComputeBackward(desc_acc[detail::DIR::bwd], inPtr,
                                                    out_acc.get_pointer());
             if (status != DFTI_NO_ERROR) {
                 throw oneapi::mkl::exception(
@@ -191,8 +192,9 @@ ONEMKL_EXPORT void compute_backward(descriptor_type &desc, sycl::buffer<input_ty
         auto outim_acc = out_im.template get_access<sycl::access::mode::write>(cgh);
 
         detail::host_task<class host_kernel_split_back_outofplace>(cgh, [=]() {
-            DFT_ERROR status = DftiComputeBackward(
-                desc_acc[detail::DIR::bwd], inre_acc.get_pointer(), inim_acc.get_pointer(),
+            auto inrePtr = const_cast<input_type*>(inre_acc.get_pointer());
+            auto inimPtr = const_cast<input_type*>(inim_acc.get_pointer());
+            DFT_ERROR status = DftiComputeBackward(desc_acc[detail::DIR::bwd], inrePtr, inimPtr, 
                 outre_acc.get_pointer(), outim_acc.get_pointer());
             if (status != DFTI_NO_ERROR) {
                 throw oneapi::mkl::exception(
--- a/src/dft/backends/mklcpu/backward.cpp
+++ b/src/dft/backends/mklcpu/backward.cpp
@@ -186,8 +186,8 @@ ONEMKL_EXPORT void compute_backward(descriptor_type &desc, sycl::buffer<input_ty
 
     cpu_queue.submit([&](sycl::handler &cgh) {
         auto desc_acc = mklcpu_desc_buffer.template get_access<sycl::access::mode::read>(cgh);
-        auto inre_acc = in_re.template get_access<sycl::access::mode::read_write>(cgh);
-        auto inim_acc = in_im.template get_access<sycl::access::mode::read_write>(cgh);
+        auto inre_acc = in_re.template get_access<sycl::access::mode::read>(cgh);
+        auto inim_acc = in_im.template get_access<sycl::access::mode::read>(cgh);
         auto outre_acc = out_re.template get_access<sycl::access::mode::write>(cgh);
         auto outim_acc = out_im.template get_access<sycl::access::mode::write>(cgh);
 
--- a/src/dft/backends/mklcpu/backward.cpp
+++ b/src/dft/backends/mklcpu/backward.cpp
@@ -156,8 +156,8 @@ ONEMKL_EXPORT void compute_backward(descriptor_type &desc, sycl::buffer<input_ty
         auto out_acc = out.template get_access<sycl::access::mode::write>(cgh);
 
         detail::host_task<class host_kernel_back_outofplace>(cgh, [=]() {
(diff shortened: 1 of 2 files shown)


anantsrivastava30:
cpu tests : 
[mklcpu_test.log](https://github.com/oneapi-src/oneMKL/files/12022077/mklcpu_test.log)

src/dft/backends/mklcpu/backward.cpp
```diff
@@ -185,10 +185,10 @@ ONEMKL_EXPORT void compute_backward(descriptor_type &desc, sycl::buffer<input_ty
 
     cpu_queue.submit([&](sycl::handler &cgh) {
         auto desc_acc = mklcpu_desc_buffer.template get_access<sycl::access::mode::read>(cgh);
-        auto inre_acc = in_re.template get_access<sycl::access::mode::read>(cgh);
-        auto inim_acc = in_im.template get_access<sycl::access::mode::read>(cgh);
-        auto outre_acc = out_re.template get_access<sycl::access::mode::write>(cgh);
```
hjabird: Why do the the `write` accessors need changing?

src/dft/backends/mklcpu/forward.cpp
```diff
@@ -153,8 +153,8 @@ ONEMKL_EXPORT void compute_forward(descriptor_type &desc, sycl::buffer<input_typ
 
     cpu_queue.submit([&](sycl::handler &cgh) {
         auto desc_acc = mklcpu_desc_buffer.template get_access<sycl::access::mode::read>(cgh);
-        auto in_acc = in.template get_access<sycl::access::mode::read>(cgh);
-        auto out_acc = out.template get_access<sycl::access::mode::write>(cgh);
+        auto in_acc = in.template get_access<sycl::access::mode::read_write>(cgh);
+        auto out_acc = out.template get_access<sycl::access::mode::read_write>(cgh);
 
         detail::host_task<class host_kernel_outofplace>(cgh, [=]() {
             DFT_ERROR status = DftiComputeForward(desc_acc[detail::DIR::fwd], in_acc.get_pointer(),
```
hjabird: Changing the accessors to from `read` to `read_write` might lead to fake dependencies between kernels, and unnecessary copies. Perhaps the const-ness of the pointer can be modifed instead? I've never used `const_cast` but it looks like it would work.

This all assumes that the `DftiComputeXXX` function truely doesn't modify the input - this wasn't the case with other domains.

FMarno: I would not agreed with the use of `const_cast`. If `DftiComputeXXX` want to guarantee that their API doesn't modify the values of the input, then the library authors can change the function signature. Otherwise they could change the implementation to modify the input in the future and we might not notice!

anantsrivastava30: @FMarno the old variadic C API cannot give any programatic gurantee because the first (and second for split) argument can be treated as inout and sometimes as unchangable inputs depending on the runtime argument list. So no we can't *easily* change the function signature! we gurantee the OOP computation preserves input but as a documented promise.

src/dft/backends/mklcpu/backward.cpp
hparks-intel: Same question from me.

src/dft/backends/mklcpu/forward.cpp
hparks-intel: > Otherwise they could change the implementation to modify the input in the future and we might not notice!

I'd agree with this concern if it were for a non-Intel backend. But since it's an Intel backend, we know the chance of this changing is next to zero. And in the unlikely event it does, the same team that would change it is committed to maintaining functionality here. I'm also not personally opposed to taking the potential performance hit of the `read_write` accessor if we want to avoid a superficially dangerous cast.

src/dft/backends/mklcpu/backward.cpp
anantsrivastava30: resolved in 5684389

src/dft/backends/mklcpu/forward.cpp
Rbiessy: It's not ideal but if we can't modify the C API then it's best to use a `const_cast` IMO. Is it something that could be revisited eventually? It sounds useful for users if that API is meant to be used in C++ applications.

hparks-intel: Yeah, the root of the problem is that this API is targeting C applications. We used the C APIs here instead of the C++ APIs to avoid the complications of our C++ API being essentially the same as the interface API here. Maybe @anantsrivastava30 can say more about whether the GPU enabling cleared up enough of those complications to consider revisiting the CPU enabling.

anantsrivastava30: From what I understand, we enabled the CPU backend using the classical API because we wanted to demonstrate that different APIs can be used to create the device specific interfaces, so we will not be, using the C++ API for CPU. 
I believe this use case of using c_api from a c++ source is rare and should not warrant a rather big (closed-source) oneMKL change. 
We can revisit this later if the fact that the c_api is rarely used from a C++ source changes, for now to get the CI working let's resolve this by using const_cast.

FMarno: Alright, but I think we should at least have a link to the place in the documentation where it says that it is guaranteed to not modify the value of input for out-of-place transforms.

hparks-intel: Thanks for the extra context about the CPU enabling @anantsrivastava30. Here's the part of the C documentation showing you'd have to explicitly set [DFTI_DESTROY_INPUT](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-1/dfti-destroy-input.html) to `DFTI_ALLOW` for the out-of-place input to be changed.

src/dft/backends/mklcpu/backward.cpp
```diff
@@ -152,11 +152,12 @@ ONEMKL_EXPORT void compute_backward(descriptor_type &desc, sycl::buffer<input_ty
 
     cpu_queue.submit([&](sycl::handler &cgh) {
         auto desc_acc = mklcpu_desc_buffer.template get_access<sycl::access::mode::read>(cgh);
-        auto in_acc = in.template get_access<sycl::access::mode::read_write>(cgh);
+        auto in_acc = in.template get_access<sycl::access::mode::read>(cgh);
         auto out_acc = out.template get_access<sycl::access::mode::write>(cgh);
 
         detail::host_task<class host_kernel_back_outofplace>(cgh, [=]() {
-            DFT_ERROR status = DftiComputeBackward(desc_acc[detail::DIR::bwd], in_acc.get_pointer(),
+            auto inPtr = const_cast<input_type*>(in_acc.get_pointer());
```
Rbiessy: The variables use snake_case, can you make them consistent?

anantsrivastava30: fixed in 4632e61b2cdbdebb6a63ce986941885e106b684d

Rbiessy (approved):

hjabird (approved):
LGTM. Probably should have another test log for completeness.

anantsrivastava30:
latest logs : [cpu_test.log](https://github.com/oneapi-src/oneMKL/files/12134891/cpu_test.log)

PR #353, opened 2023-07-20
[BLAS] --offload-arch required in both compilation and linking

andrewtbarker:
# Description

Building SYCL-BLAS backend with AMD gpu support requires a slight tweak to config, since a different command-line argument is needed for both compilation and linking.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Log attached: [amd.txt](https://github.com/oneapi-src/oneMKL/files/12113659/amd.txt)

### DIFF
--- a/src/blas/backends/syclblas/CMakeLists.txt
+++ b/src/blas/backends/syclblas/CMakeLists.txt
@@ -59,7 +59,7 @@ if(SYCLBLAS_TUNING_TARGET)
         -fsycl-targets=amdgcn-amd-amdhsa -fsycl-unnamed-lambda
       -Xsycl-target-backend --offload-arch=${HIP_TARGETS})
       target_link_options(ONEMKL::SYCL::SYCL INTERFACE
-        -fsycl-targets=amdgcn-amd-amdhsa -Xsycl-target-backend)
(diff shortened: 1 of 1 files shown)


Rbiessy (approved):
Do you know why almost all of the tests are skipped in your test log? It seems there is a missing `-` in the target, not sure if this is related.

andrewtbarker:
> Do you know why almost all of the tests are skipped in your test log? It seems there is a missing `-` in the target, not sure if this is related.

Good question - I know we skip a lot of tests because some things aren't implemented but I didn't pay attention to how many. Let me take a look.

andrewtbarker:
Based on very rough counts with grep, there are about 32 APIs implemented in `src/blas/backends/syclblas`. They run for CT and RT linking and float and double precisions (only column major layout, no complex precisions). So around 128 permutations, and I see 118 passed tests in the log, which seems pretty close to what we would expect?

Rbiessy:
Ah yes that makes sense. I thought more permutations were supported but haven't followed the progress on SYCL-BLAS recently. Thanks for checking!

mmeterel (approved):

PR #354, opened 2023-07-24
Update MKL Config file to oneMKL 2023.2 version

akabalov:
Since oneMKL 2023.2 is officially released, we need to update MKLConfig in opensource to the one that we have in oneMKL 2023.2. 

This PR addresses these required changes.

### DIFF
--- a/cmake/mkl/MKLConfig.cmake
+++ b/cmake/mkl/MKLConfig.cmake
@@ -40,7 +40,7 @@
 #    Values:  sequential,
 #             intel_thread (Intel OpenMP),
 #             gnu_thread (GNU OpenMP),
-#             pgi_thread (PGI OpenMP),
+#             pgi_thread (PGI OpenMP) [PGI support is deprecated],
 #             tbb_thread
 #    Default: intel_thread
 #       Exceptions:- DPC++ defaults to tbb, PGI compiler on Windows defaults to pgi_thread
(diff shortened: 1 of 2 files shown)


cmake/mkl/MKLConfig.cmake
```diff
@@ -661,126 +758,112 @@ foreach(lib ${MKL_LIBRARIES})
...
+if(MKL_THREADING STREQUAL "tbb_thread" OR MKL_DPCPP_THREADING STREQUAL "tbb_thread")
+  find_package(TBB CONFIG COMPONENTS tbb )
   if(MKL_THREADING STREQUAL "tbb_thread")
-    find_package(TBB CONFIG COMPONENTS tbb )
-    if(TARGET TBB::tbb)
-      set(MKL_THREAD_LIB $<TARGET_LINKER_FILE:TBB::tbb>)
-      set(MKL_SDL_THREAD_ENV "TBB")
-      get_property(TBB_LIB TARGET TBB::tbb PROPERTY IMPORTED_LOCATION_RELEASE)
-      get_filename_component(TBB_LIB_DIR ${TBB_LIB} DIRECTORY)
+    set(MKL_THREAD_LIB $<TARGET_LINKER_FILE:TBB::tbb>)
```
mkrainiuk: The TBB part is not aligned with opensource because at some point I added manual search for TBB to opensource in case cmake file is not found (some packages has only tbb libraries without cmake file). I guess we should add this search to Intel oneMKL instead of removing it from here.

mkrainiuk (commented):
Thanks for the PR, the changes looks good to me. I have one comment for TBB part.

cmake/mkl/MKLConfig.cmake
akabalov: Thank you for letting me know!
I added missed part for tbb

mkrainiuk (approved):

dnhsieh-intel (approved):

PR #358, opened 2023-07-28
[BLAS][CUBLAS] Fixed host accesses for iamax and iamin operators in cublas backend

### LINKED ISSUE #312: [cublas] optimize iamax/iamin return value memory transfers
# Summary

The oneapi spec specifies zero-based indexing for the return value in `iamax` and `iamin`, but the cublas backend uses one-based indexing. If the return value is in host memory this is not a big deal to correct, but if the user provides a pointer for the return value that points to device-only memory our current implementation copies to host, adjusts the indexing, and copies back, which could be a performance problem.

# Version

Affects oneMKL 4c7e7eb. This issue arises from a discussion in PR #304 

# Suggested fix

It is probably possible to adjust the indexing on device without memory transfers with a simple SYCL kernel. A brief attempt to do so didn't work, every kernel I tried produced the error `PI_ERROR_INVALID_BINARY` when running the code (compilation was fine). This was not investigated very thoroughly because of other priorities.

muhammad-tanvir-1211:
# Description

This patch fixes the redundant host accesses for adjusting the output index for `iamax` and `iamin` operators in the `cublas` backend. The host accesses are now replaced by a simple SYCL kernel that performs the same index adjustment directly on the device.

Fixes #312 

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? [cublas_iamax_iamin_fix.log](https://github.com/oneapi-src/oneMKL/files/12195371/cublas_iamax_iamin_fix.log)

### DIFF
--- a/src/blas/backends/cublas/CMakeLists.txt
+++ b/src/blas/backends/cublas/CMakeLists.txt
@@ -37,6 +37,10 @@ target_include_directories(${LIB_OBJ}
           ${PROJECT_SOURCE_DIR}/src
 )
 target_compile_options(${LIB_OBJ} PRIVATE ${ONEMKL_BUILD_COPT})
+target_compile_options(ONEMKL::SYCL::SYCL INTERFACE
+      -fsycl-targets=nvptx64-nvidia-cuda -fsycl-unnamed-lambda)
+target_link_options(ONEMKL::SYCL::SYCL INTERFACE
+      -fsycl-targets=nvptx64-nvidia-cuda)
 target_link_libraries(${LIB_OBJ} PUBLIC ONEMKL::SYCL::SYCL ONEMKL::cuBLAS::cuBLAS)
 target_compile_features(${LIB_OBJ} PUBLIC cxx_std_11)
 set_target_properties(${LIB_OBJ} PROPERTIES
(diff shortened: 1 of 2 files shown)


src/blas/backends/cublas/CMakeLists.txt
```diff
@@ -37,6 +37,10 @@ target_include_directories(${LIB_OBJ}
           ${PROJECT_SOURCE_DIR}/src
 )
 target_compile_options(${LIB_OBJ} PRIVATE ${ONEMKL_BUILD_COPT})
+target_compile_options(ONEMKL::SYCL::SYCL INTERFACE
```
andrewtbarker, line 40: These options are what I was missing when I tried this before...

andrewtbarker (approved):
Thanks for getting this working, this is a nice improvement.

src/blas/backends/cublas/CMakeLists.txt
mehdi-goli: @andrewtbarker  yes, because oneMKL cuBLAS backend did not have any kernel in it, no target triple was needed and it automatically using the DPCPP default target triple which is SPIR64 , This would generate the wrong binary code. once a kernel code is used inside the oneMKL backend we need to pass the relevant target triple according to the chosen device so that the correct binary would be generated

andrewtbarker:
@mkrainiuk Can you take a quick look? (You already approved the almost identical #361)

mkrainiuk (approved):

PR #343, opened 2023-07-04
[DFT] Fix strides for cuFFT

Rbiessy:
# Description

Fix `inembed` and `onembed` for non-default strides with the cuFFT backend. We don't have tests for this configuration at the moment. Adding proper tests is too much work at the moment so I suggest we merge the fix only for now.
We have a task to add this kind of tests in the future but it is unclear when we can start it.

Fixes https://github.com/oneapi-src/oneMKL/issues/341

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Attach a log.
- [x] Have you formatted the code using clang-format?

## New interfaces

- [ ] Have you provided motivation for adding a new feature as part of RFC and
it was accepted? # (RFC)
- [ ] What version of oneAPI spec the interface is targeted?
- [ ] Complete [New features](pull_request_template.md#new-features) checklist

## New features

- [ ] Have you provided motivation for adding a new feature?
- [ ] Have you added relevant tests?

## Bug fixes

- [ ] Have you added relevant regression tests?
- [ ] Have you included information on how to reproduce the issue (either in a
      GitHub issue or in this PR)?

PR #363, opened 2023-08-07
[DFT][cuFFT][rocFFT] Add tests for strides and fix strides in cuFFT and rocFFT backends

t4c1:
# Description

Fixes strides not being properly handles by cuFFT and rocFFT backends and adds tests for strides. 

Some relevant discussion in a previous incomplete PR: https://github.com/oneapi-src/oneMKL/pull/343

# Checklist

## All Submissions

## New features


src/dft/backends/cufft/commit.cpp
```diff
@@ -138,27 +139,83 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+            }
         }
-        else if (rank == 3) {
-            inembed[2] = config_values.input_strides[2];
-            inembed[1] = config_values.input_strides[1] / inembed[2];
+        else {
+            if (o_min - onembed.begin() != i_min - inembed.begin()) {
+                throw mkl::unimplemented(
+                    "dft/backends/cufft", __FUNCTION__,
+                    "cufft requires that if  ordered by stride length, the order of strides is the same for input and output strides!");
```
Rbiessy:
```suggestion
                    "cufft requires that if ordered by stride length, the order of strides is the same for input and output strides!");
```

src/dft/backends/cufft/commit.cpp
```diff
@@ -138,27 +139,83 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
         const int rank = static_cast<int>(config_values.dimensions.size());
-        const int istride = static_cast<int>(config_values.input_strides.back());
-        const int ostride = static_cast<int>(config_values.output_strides.back());
-        const int batch = static_cast<int>(config_values.number_of_transforms);
-        const int fwd_dist = static_cast<int>(config_values.fwd_dist);
-        const int bwd_dist = static_cast<int>(config_values.bwd_dist);
-        std::array<int, max_supported_dims> inembed;
-        if (rank == 2) {
-            inembed[1] = config_values.input_strides[1];
+        std::vector<int> inembed = { config_values.input_strides.begin(),
```
Rbiessy: It is confusing to add the first value (the offset) to `[io]nembed`. If I understand correctly this works because the first value of `[io]nembed` is never read and we later remove the smallest stride. This could use a few comments IMO

tests/unit_tests/dft/include/test_common.hpp
```diff
@@ -197,28 +197,179 @@ inline std::array<std::int64_t, 4> get_default_strides(const std::vector<std::in
     }
 }
 
+template <typename T>
+T getdefault(const std::vector<T> vec, std::size_t idx, T default_) {
```
Rbiessy: `getdefault` -> `get_default`

src/dft/backends/cufft/commit.cpp
```diff
@@ -48,6 +48,7 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
     // We also need this because oneMKL uses a directionless "FWD_DISTANCE" and "BWD_DISTANCE" while cuFFT uses a directional "idist" and "odist".
     // plans[0] is forward, plans[1] is backward
     std::array<std::optional<cufftHandle>, 2> plans = { std::nullopt, std::nullopt };
+    std::array<int64_t, 2> offsets;
```
FMarno:
```suggestion
    std::array<std::int64_t, 2> offsets;
```

src/dft/backends/cufft/backward.cpp
```diff
@@ -35,6 +35,10 @@
 
 namespace oneapi::mkl::dft::cufft {
 namespace detail {
+//forward declaration
+template <dft::precision prec, dft::domain dom>
+std::array<int64_t, 2> get_offsets(dft::detail::commit_impl<prec, dom> *commit);
```
FMarno:
```suggestion
std::array<std::int64_t, 2> get_offsets(dft::detail::commit_impl<prec, dom> *commit);
```

src/dft/backends/cufft/backward.cpp
```diff
@@ -85,6 +100,15 @@ ONEMKL_EXPORT void compute_backward(descriptor_type &desc,
     auto commit = detail::checked_get_commit(desc);
     auto queue = commit->get_queue();
     auto plan = detail::get_bwd_plan(commit);
+    auto offsets = detail::get_offsets(commit);
+
+    if constexpr (std::is_floating_point_v<fwd<descriptor_type>>) {
+        if (offsets[1] % 2 != 0) {
+            throw oneapi::mkl::unimplemented(
+                "DFT", "compute_forward(desc, inout)",
```
FMarno: The function arguments in this string and others are incorrect. can you utilize `func_name` from the host task?

src/dft/backends/cufft/backward.cpp
```diff
@@ -51,6 +55,16 @@ ONEMKL_EXPORT void compute_backward(descriptor_type &desc,
...
     auto queue = commit->get_queue();
     auto plan = detail::get_bwd_plan(commit);
+    auto offsets = detail::get_offsets(commit);
+
+    if constexpr (std::is_floating_point_v<fwd<descriptor_type>>) {
+        offsets[0] *= 2; // offset is supplied in complex but we offset scalar pointer
+        if (offsets[1] % 2 != 0) {
+            throw oneapi::mkl::unimplemented(
+                "DFT", "compute_forward(desc, inout)",
+                "cuFFT requires offset (first value in strides) to be multiple of `sizeof(complex)`!");
```
FMarno: If the value needs to be a multiple of `sizeof(complex)`, then why are you doing mod 2?

src/dft/backends/cufft/commit.cpp
FMarno:
```suggestion
        std::vector<int> inembed { config_values.input_strides.begin(),
```
and for onembed

src/dft/backends/cufft/commit.cpp
```diff
@@ -138,27 +139,83 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+        inembed.erase(i_min);
+        onembed.erase(o_min);
+        if (o_min - onembed.begin() != rank) {
+            // swap dimensions to have the last one have the smallest stride
+            std::swap(n_copy[o_min - onembed.begin() - 1], n_copy[rank - 1]);
         }
-        else if (rank == 3) {
-            onembed[2] = config_values.output_strides[2];
-            onembed[1] = config_values.output_strides[1] / onembed[2];
+        for (int i = 0; i < rank - 1; i++) {
```
FMarno:
```suggestion
        for (int i = 1; i < rank; i++) {
```
then remove the +1 from all the indexes in the loop.

src/dft/backends/cufft/forward.cpp
```diff
@@ -88,6 +104,16 @@ ONEMKL_EXPORT void compute_forward(descriptor_type &desc, sycl::buffer<fwd<descr
     auto queue = commit->get_queue();
     auto plan = detail::get_fwd_plan(commit);
 
+    auto offsets = detail::get_offsets(commit);
+
+    if constexpr (std::is_floating_point_v<fwd<descriptor_type>>) {
+        if (offsets[0] % 2 != 0) {
+            throw oneapi::mkl::unimplemented(
+                "DFT", "compute_forward(desc, inout)",
```
FMarno: wrong parameters in these string too

src/dft/backends/cufft/commit.cpp
t4c1: done

src/dft/backends/cufft/commit.cpp
t4c1: added a comment

tests/unit_tests/dft/include/test_common.hpp
t4c1: done

src/dft/backends/cufft/commit.cpp
t4c1: done

src/dft/backends/cufft/backward.cpp
t4c1: done

src/dft/backends/cufft/backward.cpp
t4c1: yes, fixed

src/dft/backends/cufft/backward.cpp
t4c1: it needs to be a multiple of `sizeof(complex)/sizeof(scalar)`, which equals 2. Fixed misleading error message.

src/dft/backends/cufft/commit.cpp
t4c1: done

src/dft/backends/cufft/commit.cpp
t4c1: done

src/dft/backends/cufft/forward.cpp
t4c1: fixed

t4c1:
Attaching logs form Intel and AMD gpus:
[log_intel_gpu.txt](https://github.com/oneapi-src/oneMKL/files/12302805/log_intel_gpu.txt)
[log_amd_gpu.txt](https://github.com/oneapi-src/oneMKL/files/12302806/log_amd_gpu.txt)

I will attach a log from Nvidia GPU at a later time.

src/dft/backends/cufft/backward.cpp
FMarno: You missed this one I think

src/dft/backends/cufft/forward.cpp
```diff
@@ -37,6 +37,10 @@
 namespace oneapi::mkl::dft::cufft {
 
 namespace detail {
+//forward declaration
+template <dft::precision prec, dft::domain dom>
+std::array<int64_t, 2> get_offsets(dft::detail::commit_impl<prec, dom> *commit);
```
FMarno:
```suggestion
std::array<std::int64_t, 2> get_offsets(dft::detail::commit_impl<prec, dom> *commit);
```

src/dft/backends/rocfft/backward.cpp
```diff
@@ -36,6 +36,10 @@
 
 namespace oneapi::mkl::dft::rocfft {
 namespace detail {
+//forward declaration
+template <dft::precision prec, dft::domain dom>
+std::array<int64_t, 2> get_offsets(dft::detail::commit_impl<prec, dom> *commit);
```
FMarno:
```suggestion
std::array<std::int64_t, 2> get_offsets(dft::detail::commit_impl<prec, dom> *commit);
```

src/dft/backends/rocfft/backward.cpp
```diff
@@ -52,21 +56,33 @@ rocfft_execution_info get_bwd_info(dft::detail::commit_impl<prec, dom> *commit)
...
     auto commit = detail::checked_get_commit(desc);
     auto queue = commit->get_queue();
     auto plan = detail::get_bwd_plan(commit);
     auto info = detail::get_bwd_info(commit);
+    auto offsets = detail::get_offsets(commit);
+
+    if constexpr (std::is_floating_point_v<fwd<descriptor_type>>) {
+        offsets[0] *= 2; // offset is supplied in complex but we offset scalar pointer
+    }
+    if (offsets[0] != offsets[1]) {
```
FMarno, line 71: Can you check this at commit time by looking at the placement commit parameter to see if its needed?

FMarno: Also, why is this a requirement when you are just applying the offset to the pointer yourself? rocFFT doesn't ever see the offset.

t4c1:
Test logs after the last change:
[log_amd_gpu.txt](https://github.com/oneapi-src/oneMKL/files/12311965/log_amd_gpu.txt)
[log_intel_gpu.txt](https://github.com/oneapi-src/oneMKL/files/12311966/log_intel_gpu.txt)
[log_nvidia_gpu.txt](https://github.com/oneapi-src/oneMKL/files/12311967/log_nvidia_gpu.txt)

tests/unit_tests/dft/include/compute_inplace.hpp
```diff
@@ -23,101 +23,69 @@
...
-        const auto real_strides = get_conjugate_even_real_component_strides(sizes);
+    descriptor.set_value(oneapi::mkl::dft::config_param::FWD_DISTANCE, forward_distance);
+    descriptor.set_value(oneapi::mkl::dft::config_param::BWD_DISTANCE, backward_distance);
+    if (strides_fwd.size()) {
+        descriptor.set_value(oneapi::mkl::dft::config_param::INPUT_STRIDES, strides_fwd.data());
+    }
+    if (strides_bwd.size()) {
+        descriptor.set_value(oneapi::mkl::dft::config_param::OUTPUT_STRIDES, strides_bwd.data());
+    }
+    else if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
```
FMarno: Is this still helpful? I think you always set strides_bwd in the REAL case?

tests/unit_tests/dft/include/compute_inplace.hpp
```diff
@@ -126,48 +94,51 @@ int DFT_Test<precision, domain>::test_in_place_buffer() {
...
+                    strides_bwd, abs_error_margin, rel_error_margin, std::cout));
             }
         }
 
-        if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
-            const auto real_strides = get_conjugate_even_real_component_strides(sizes);
+        if (strides_bwd.size()) {
+            descriptor.set_value(oneapi::mkl::dft::config_param::INPUT_STRIDES, strides_bwd.data());
+        }
+        else if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
```
FMarno: again, is this path ever taken

tests/unit_tests/dft/include/compute_out_of_place.hpp
```diff
@@ -23,73 +23,79 @@
...
-            const auto ref_row_stride = sizes.back();
-            const auto backward_row_stride = cast_unsigned(get_backward_row_size<domain>(sizes));
-            const auto backward_row_elements = cast_unsigned(get_backward_row_size<domain>(sizes));
-
-            while (ref_iter < out_host_ref.end()) {
-                EXPECT_TRUE(check_equal_vector(bwd_ptr, ref_iter, backward_row_elements,
-                                               abs_error_margin, rel_error_margin, std::cout));
-                bwd_ptr += backward_row_stride;
-                ref_iter += ref_row_stride;
+            for (int64_t i = 0; i < batches; i++) {
```
FMarno:
```suggestion
            for (std::int64_t i = 0; i < batches; i++) {
```
and other instances in this file

src/dft/backends/rocfft/backward.cpp
t4c1: > Can you check this at commit time by looking at the placement commit parameter to see if its needed?

I could, but it is easier to do here when I also know direction of the transform. At commit time I would need to check both directions to see if either makes sense.

> Also, why is this a requirement when you are just applying the offset to the pointer yourself? rocFFT doesn't ever see the offset.

For in-place transforms rocfft accepts single pointer, so we can only apply one offset. That only makes sense if both offsets are the same.

tests/unit_tests/dft/include/compute_out_of_place.hpp
t4c1: fixed

tests/unit_tests/dft/include/compute_inplace.hpp
t4c1: nope, removed

tests/unit_tests/dft/include/compute_inplace.hpp
t4c1: right, removed.

src/dft/backends/rocfft/backward.cpp
t4c1: fixed

src/dft/backends/cufft/forward.cpp
t4c1: fixed

src/dft/backends/rocfft/backward.cpp
t4c1: fixed

src/dft/backends/rocfft/backward.cpp
FMarno: > For in-place transforms rocfft accepts single pointer, so we can only apply one offset. That only makes sense if both offsets are the same.

Ok that makes sense.


wrt checking at commit time, I'm speaking specifically about the `offset[0] != offset[1]` test, I believe that is orthogonal to direction.

tests/unit_tests/dft/include/compute_out_of_place.hpp
```diff
@@ -112,16 +122,22 @@ int DFT_Test<precision, domain>::test_out_of_place_USM() {
...
+    descriptor.set_value(oneapi::mkl::dft::config_param::FWD_DISTANCE, forward_distance);
     descriptor.set_value(oneapi::mkl::dft::config_param::BWD_DISTANCE, backward_distance);
-    if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
+    if (strides_fwd.size()) {
+        descriptor.set_value(oneapi::mkl::dft::config_param::INPUT_STRIDES, strides_fwd.data());
+    }
+    if (strides_bwd.size()) {
+        descriptor.set_value(oneapi::mkl::dft::config_param::OUTPUT_STRIDES, strides_bwd.data());
+    }
+    else if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
```
FMarno, line 152: Considering this wasn't needed in in-place, I guess it might not be needed here either (and I guess there will be 3 other occurances).

src/dft/backends/rocfft/backward.cpp
t4c1: It is not. For real transforms direction determines which offset is specified in reals and which in complex values. This is the problem of having input/output values instead of forward/backward ...

tests/unit_tests/dft/include/compute_out_of_place.hpp
t4c1: Here it is needed, as in this test strides_bwd can be empty.

src/dft/backends/rocfft/backward.cpp
FMarno: Ok, I see. Thank you for explaining.

Rbiessy (approved):

src/dft/backends/cufft/commit.cpp
```diff
@@ -138,27 +139,84 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
-        const int fwd_dist = static_cast<int>(config_values.fwd_dist);
-        const int bwd_dist = static_cast<int>(config_values.bwd_dist);
-        std::array<int, max_supported_dims> inembed;
-        if (rank == 2) {
-            inembed[1] = config_values.input_strides[1];
+        // cufft ignores the first value in inembed and onembed, so there is no harm in putting offset there
+        std::vector<int> inembed{ config_values.input_strides.begin(),
+                                  config_values.input_strides.end() };
+        std::vector<int> onembed{ config_values.output_strides.begin(),
+                                  config_values.output_strides.end() };
```
raphael-egan, line 146: Shouldn't we enforce `inembed = onembed` in case of in-place c2c transforms (and disregard the output strides given by the user), in order to align with the [specs](https://spec.oneapi.io/versions/latest/elements/oneMKL/source/domains/dft/config_params/strides.html#onemkl-dft-config-strides)' "For in-place transforms(config_param::PLACEMENT=config_value::INPLACE), the configuration set by config_param::OUTPUT_STRIDES is **ignored** when the element types in the forward and backward domains are the same."
The same comment holds for the offsets below and for other backends.
A somewhat less aggressive approach would be to check for equality of the strides in case of in-place c2c and throw an appropriate exception if they're not equal, maybe.

src/dft/backends/cufft/commit.cpp
```diff
@@ -138,27 +139,84 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
         }
-        else if (rank == 3) {
-            inembed[2] = config_values.input_strides[2];
-            inembed[1] = config_values.input_strides[1] / inembed[2];
+        else {
+            if (o_min - onembed.begin() != i_min - inembed.begin()) {
+                throw mkl::unimplemented(
+                    "dft/backends/cufft", __FUNCTION__,
+                    "cufft requires that if ordered by stride length, the order of strides is the same for input and output strides!");
+            }
```
raphael-egan: Wouldn't this fail to catch a case like `input_strides = {0, 36, 1, 6}` and `output_strides = {0, 6, 1, 36}` for a cfo6x6x6, for instance?
EDIT: such a case seems caught below but returns a slightly different exception message. Could we consolidated the two snippets and/or align the exception messages?

src/dft/backends/cufft/commit.cpp
```diff
@@ -171,6 +229,10 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
             ignore_strides || (n_copy[rank - 1] <= onembed[rank - 1] &&
                                (n_copy[rank - 1] / 2 + 1) <= inembed[rank - 1]);
 
+        if (!valid_forward && !valid_backward) {
```
raphael-egan, line 263: These are insightful for real transforms only. Couldn't we generalize the logic to complex transform as well? For instance, catch cases like attempting a cfo6x6 with nonsensically small input or output strides like {0, 3, 1}?
(Similar comment for other backends)

src/dft/backends/mklgpu/commit.cpp
```diff
@@ -113,6 +113,11 @@ class mklgpu_commit final : public dft::detail::commit_impl<prec, dom> {
                        to_mklgpu<onemkl_param::CONJUGATE_EVEN_STORAGE>(config.conj_even_storage));
         desc.set_value(backend_param::PLACEMENT,
                        to_mklgpu<onemkl_param::PLACEMENT>(config.placement));
+
+        if (config.input_strides[0] != 0 || config.output_strides[0] != 0) {
+            throw mkl::unimplemented("dft/backends/mklgpu", "commit",
+                                     "MKLGPU does not support nonzero offsets.");
+        }
```
raphael-egan, line 121: Maybe leave a comment around here that this could be removed for proper exception handling of the exception thrown at commit, once [MKLD-16060](https://jira.devtools.intel.com/browse/MKLD-16060) is completed.

src/dft/backends/rocfft/commit.cpp
```diff
@@ -433,4 +472,24 @@ create_commit(
...
+template std::array<std::int64_t, 2>
+get_offsets<dft::detail::precision::DOUBLE, dft::detail::domain::REAL>(
+    dft::detail::commit_impl<dft::detail::precision::DOUBLE, dft::detail::domain::REAL>*);
+template std::array<std::int64_t, 2>
+get_offsets<dft::detail::precision::DOUBLE, dft::detail::domain::COMPLEX>(
+    dft::detail::commit_impl<dft::detail::precision::DOUBLE, dft::detail::domain::COMPLEX>*);
+
+} //namespace detail
+
 } // namespace oneapi::mkl::dft::rocfft
```
raphael-egan, line 504: Some of my comments in `cufft/commit` apply here above as well (_e.g._ output strides to be ignored for c2c in-place for instance) but I haven't seen anything intrinsically wrong here otherwise, although my knowledge of rocFFT is more than superficial. 
1. I am a little surprise to see that there is no condition checking for similar dimension-ordering on input and output as was done for cuFFT: is a case like input_strides = {0, 36, 1, 6} and output_strides = {0, 6, 1, 36} for a cfo6x6x6 handled smoothly as is by rocFFT? I'm not even sure MKL can do that reliably so  we might want to integrate the ordering thingy into the specs, in fact.
2. Out of curiosity, can rocFFT handle a multi-dimensional real transform with the smallest stride being set for another dimension than the last? I don't think I saw a check for that here above.

tests/unit_tests/dft/include/compute_inplace.hpp
```diff
@@ -23,101 +23,70 @@
...
+            //to be able to calculate in place each row must fit backward data
+            for (size_t i = 0; i < sizes.size() - 1; i++) {
+                strides_bwd.push_back(strides_tmp[i + 1]);
+            }
+            strides_bwd.push_back(1);
+        }
+    }
+    else {
+        // spec says strides_bwd is ignored and strides_fwd is reused for backward domain for in-place complex
+        strides_fwd = strides_bwd;
```
raphael-egan, line 57: Shouldn't this assignment be reversed? It confuses me as `strides_bwd` becomes the `OUTPUT_STRIDES` here below.

tests/unit_tests/dft/include/compute_inplace.hpp
```diff
@@ -179,69 +147,94 @@ int DFT_Test<precision, domain>::test_in_place_USM() {
...
-    if constexpr (domain == oneapi::mkl::dft::domain::REAL) {
-        const auto real_strides = get_conjugate_even_real_component_strides(sizes);
-        const auto complex_strides = get_conjugate_even_complex_strides(sizes);
-
-        descriptor.set_value(oneapi::mkl::dft::config_param::INPUT_STRIDES, complex_strides.data());
+    if (strides_fwd.size()) {
+        descriptor.set_value(oneapi::mkl::dft::config_param::OUTPUT_STRIDES, strides_fwd.data());
+    }
+    else {
+        auto real_strides = get_default_strides(sizes);
```
raphael-egan: always "real" or misnomer?

tests/unit_tests/dft/include/compute_out_of_place.hpp
```diff
@@ -23,73 +23,85 @@
...
+        if (strides_fwd.size()) {
+            descriptor.set_value(oneapi::mkl::dft::config_param::OUTPUT_STRIDES,
+                                 strides_fwd.data());
+        }
+        else {
+            auto real_strides = get_default_strides(sizes);
             descriptor.set_value(oneapi::mkl::dft::config_param::OUTPUT_STRIDES,
                                  real_strides.data());
-            commit_descriptor(descriptor, sycl_queue);
         }
```
raphael-egan: `real_` in `real_strides` used here above (and in other places herein) seems to be a fairly confusing misnomer.

Rbiessy:
Test log running on Intel Arc: [log_intel_gpu.txt](https://github.com/oneapi-src/oneMKL/files/12427223/log_intel_gpu.txt)

src/dft/backends/rocfft/commit.cpp
t4c1: Yes, rocFFT handles both of these cases. I have not checked, but I suspect it uses some additional buffers to do that.

tests/unit_tests/dft/include/compute_inplace.hpp
t4c1: It works either as it is or reversed. Just both must have the same value. Since we are testing transforms in both directions it also at some point becomes INPUT_STRIDES.

src/dft/backends/cufft/commit.cpp
t4c1: Right, fixed to follow the spec.

src/dft/backends/cufft/commit.cpp
t4c1: Aligned the exception message.

src/dft/backends/cufft/commit.cpp
t4c1: Yes, done.

src/dft/backends/mklgpu/commit.cpp
t4c1: Done.

tests/unit_tests/dft/include/compute_inplace.hpp
t4c1: misnomer, renamed.

tests/unit_tests/dft/include/compute_out_of_place.hpp
t4c1: fixed

src/dft/backends/rocfft/commit.cpp
```diff
@@ -281,21 +287,23 @@ class rocfft_commit final : public dft::detail::commit_impl<prec, dom> {
...
                   [&](std::size_t a, std::size_t b) { return out_strides[a] < out_strides[b]; });
         std::array<std::size_t, max_supported_dims> lengths_cplx = lengths;
-        lengths_cplx[0] = lengths_cplx[0] / 2 + 1;
+        if(dom == dft::domain::REAL){
+            lengths_cplx[0] = lengths_cplx[0] / 2 + 1;
+        }
         // When creating real-complex descriptions, the strides will always be wrong for one of the directions.
         // This is because the least significant dimension is symmetric.
         // If the strides are invalid (too small to fit) then just don't bother creating the plan.
         const bool ignore_strides = dom == dft::domain::COMPLEX || dimensions == 1;
```
raphael-egan: It looks like this is unused, now.

raphael-egan (approved):
Thank you for the changes. Just noted one more comment about one seemingly unused variable upon review-related changes.
Approved assuming that the tests still pass successfully after changes.

src/dft/backends/rocfft/commit.cpp
t4c1: Right, removed.

PR #169, opened 2022-02-24
[BLAS] Fix cublas perf

aelizaro:
# Description
 This PR fixes the performance overhead in curand backend. It took a long time to create sycl::context with default constructor in cublas handler

Fixes #111[ # (GitHub issue)](https://github.com/oneapi-src/oneMKL/issues/111)

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? 
[cublas_test.log](https://github.com/oneapi-src/oneMKL/files/8135144/cublas_test.log)

- [x] Have you formatted the code using clang-format?

PR #295, opened 2023-03-28
[BLAS] fix deprecation warnings from cuda.hpp

andrewtbarker:
# Description

Recent compilers issue warnings when we `#include "cuda.hpp"` from the LLVM SYCL backends. This PR removes those warnings.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Log attached: [test.txt](https://github.com/oneapi-src/oneMKL/files/11094562/test.txt)
- [x] Have you formatted the code using clang-format?

Issue #106, opened 2021-06-10
A better way to measure performance of SGEMM using cuBLAS backend


vitduck:
# Summary
onMKL shows underwhelmed SGEMM performance for small matrices.

# Version
oneMKL v0.2

# Environment
* HW you use: NVIDIA V100
* Backend library version: CUDA/10.2 
* OS name and version: CentOS 7.4.1708
* Compiler: Intel-llvm 2021-06-08 

# Steps to reproduce
I attached the source files required to measure the performance of sgemm (GFLOPS). 
Timing is obtained using std;:chrono. 

# Observed behavior
For m = n = k  = 4096, the observed performance is ~ 500 GFlops. In comparison, native cuBLAS archives up to 10 TFlops. 
I think one possible reason is the high cost of creating cuBLAS context under the hood.  
For native cuBLAS, the creation of context can be effectively excluded from cudaEventRecord().
Is there a better way to measure the performance of SGEMM with cuBLAS backend ? 

Thanks 
[sgemm.zip](https://github.com/oneapi-src/oneMKL/files/6629559/sgemm.zip)

vrpascuzzi:
@vitduck,
Thanks for reporting this useful observation. Indeed, I discussed at some length with @mehdi-goli regarding this exact bottleneck. It's probably best he provides insight.

vitduck:
Performance (Gflops): 
| n=m=k | cublas  |  oneMKL |
| -------- | ------- | --------- |  
|   1024   | 10599  |      9.1     | 
|   2048   | 11666  |     72.6    |
|   4096   | 12942  |      568    | 
|   8192   | 13836  |     3411   | 
| 16834   | 13376  |     9981   |

Cost of context creation can be offset by using sufficient large matrix. 
Assuming that the cost of context creation is constant regardless of matrix size, it can be substracted between two run. 
The performnace would be : (2*N2**3 - 2*N1**3)/(t2 - t1)/1e9 
With N2=8182, N1=4096, the performace is ~ 13 TFlops, which is quite close to native cuBLAS.  
In any case, the context creation would be negligible w.r.t walltime for serious application. 

@vrpascuzzi 
I appreciate your comments as well. If nothing else, may the issue be closed ?

mehdi-goli:
Thanks, @mkrainiuk, and @vrpascuzzi. I had a brief chat with @mkrainiuk and we believe that caching the CUDA context for reuse should yield better performance.

mmeterel:
[...]

PR #365, opened 2023-08-14
[LAPACK][cuSOLVER] Get cuSOLVER backend to build with SYCL2020 changes

### LINKED ISSUE #360: oneMKL build failed with the latest syclos compiler
# Summary
oneMKL build failed with the latest syclos compiler (20230731)

# Steps to reproduce
```
cmake .. \
         -DENABLE_CUBLAS_BACKEND=True    \
         -DENABLE_CUSOLVER_BACKEND=True  \
         -DENABLE_CURAND_BACKEND=True    \
         -DENABLE_CUFFT_BACKEND=True     \
         -DENABLE_MKLCPU_BACKEND=False   \
         -DENABLE_MKLGPU_BACKEND=False   \
         -DBUILD_FUNCTIONAL_TESTS=False  \
         -DCMAKE_INSTALL_PREFIX=/export/users/zhiweij1/workspace/onemkl_interfaces/20230801

make -j12 && make install
```

# Observed behavior
Error message
```
In file included from /export/users/zhiweij1/workspace/onemkl_interfaces/oneMKL/src/lapack/backends/cusolver/cusolver_batch.cpp:21:
In file included from /export/users/zhiweij1/workspace/onemkl_interfaces/oneMKL/src/lapack/backends/cusolver/cusolver_task.hpp:32:
In file included from /export/users/zhiweij1/workspace/onemkl_interfaces/oneMKL/src/lapack/backends/cusolver/cusolver_scope_handle.hpp:27:
/nfs/shm/proj/icl/cmplrarch/deploy_syclos/llorgsyclngefi2linux/20230731_160000/build/linux_qa_release/bin/../include/sycl/backend/cuda.hpp:17:2: warning: sycl/backend/cuda.hpp is deprecated and not required anymore [-W#warnings]
   17 | #warning sycl/backend/cuda.hpp is deprecated and not required anymore
      |  ^
In file included from /export/users/zhiweij1/workspace/onemkl_interfaces/oneMKL/src/lapack/backends/cusolver/cusolver_batch.cpp:21:
In file included from /export/users/zhiweij1/workspace/onemkl_interfaces/oneMKL/src/lapack/backends/cusolver/cusolver_task.hpp:32:
/export/users/zhiweij1/workspace/onemkl_interfaces/oneMKL/src/lapack/backends/cusolver/cusolver_scope_handle.hpp:82:11: error: no type named 'interop_handler' in namespace 'sycl'; did you mean 'interop_handle'?
   82 |     sycl::interop_handler &ih;
      |     ~~~~~~^~~~~~~~~~~~~~~
      |           interop_handle
/nfs/shm/proj/icl/cmplrarch/deploy_syclos/llorgsyclngefi2linux/20230731_160000/build/linux_qa_release/bin/../include/sycl/interop_handle.hpp:36:7: note: 'interop_handle' declared here
   36 | class interop_handle {
      |       ^
```

I think oneMKL needs to update the code since the compiler has merged some ABI-breaking PRs (e.g., https://github.com/intel/llvm/pull/10246/)

sknepper:
# Description

Allows oneMKL to build with cuSolver backend and fixes the many resulting segfaults/failures. The only remaining test failures are in sytrd/hetrd, which are due to a failing off-diagonal check. This may be an issue in the underlying cuSOLVER library.

Changes include:
    interop_handler -> interop_handle
    get_* -> get_native_*
    backend::cuda -> backend::ext_oneapi_cuda
    Also applied same fix for performance as in #169
    Remove SYCL2020_DISABLE_DEPRECATION_WARNINGS flag for cuSOLVER backend
    Wait on the queue for each scratchpad_size function

Fixes #360 

# Checklist

## All Submissions

- [No ] Do all unit tests pass locally? Attach a log. [cusolver.txt](https://github.com/oneapi-src/oneMKL/files/12756326/cusolver.txt)
However, only sytrd/hetrd are failing due to an off-diagonal check.
- [Yes ] Have you formatted the code using clang-format?


src/lapack/backends/cusolver/cusolver_scope_handle.hpp
```diff
@@ -23,8 +23,10 @@
 #else
 #include <CL/sycl.hpp>
 #endif
-#if __has_include(<sycl/backend/cuda.hpp>)
+#if __has_include(<sycl/context.hpp>)
+#if __SYCL_COMPILER_VERSION <= 20220930
 #include <sycl/backend/cuda.hpp>
```
jadurazo, line 28: Just checking if we need to have a `has_include` macro for this include file? (It was removed in line 26 above.)

jadurazo (approved):
Looks good! Just one question.

src/lapack/backends/cusolver/cusolver_scope_handle.hpp
sknepper: Good question! I followed the same changes that had been made to the src/blas/backends/cublas/cublas_scope_handle.hpp file, so we're aligned with the cuBLAS backend.

My understanding is that the location of some of the headers changed based on the compiler version. If there is an overarching CL directory, that's one case. And we can capture that case by checking the existence of sycl/context.hpp (without the CL/ directory), so I don't think we need to verify that sycl/backend/cuda.hpp exists.
If there's not a CL/ directory, then we look at the date of the compiler. Based on #295 , it looks like there were deprecation warnings for the cuda.hpp file, so that's why it's only conditionally included.

src/lapack/backends/cusolver/cusolver_scope_handle.cpp
```diff
@@ -39,12 +39,12 @@ thread_local cusolver_handle<pi_context> CusolverScopedContextHandler::handle_he
     cusolver_handle<pi_context>{};
 
 CusolverScopedContextHandler::CusolverScopedContextHandler(sycl::queue queue,
-                                                           sycl::interop_handler &ih)
+                                                           sycl::interop_handle &ih)
         : ih(ih),
           needToRecover_(false) {
-    placedContext_ = queue.get_context();
+    placedContext_ = new sycl::context(queue.get_context());
```
ericlars, line 45: Is there a reason placedContext was changed to a pointer? There is a very contrived scenario where a memory leak could occur. The cuda context get/set functions can throw and the user could recover from the exception, but placedContext_ will have leaked.

sknepper: Thanks for the question! This was done for performance reasons, following the same fix as in https://github.com/oneapi-src/oneMKL/pull/169. According to the description for the PR, it took a long time to create sycl::context with default constructor in cublas handler. I didn't run any performance tests, but I don't see a reason it would be different for the cuSOLVER backend. In #106 , you can see some performance comparisons for sgemm. The change significantly improved the performance of the oneMKL interfaces with cuBLAS backend.

ericlars: It does eliminate a call to the default constructor. I don't see how failing to get/set the cuda context would be recoverable anyways so I'll approve this as is.

ericlars (approved):

PR #369, opened 2023-08-29
[cmake] Add missing dependencies

Rbiessy:
# Description

The library `onemkl` does not depend on any of the libraries that it loads via `dlopen`. This means that a user can build the runtime tests or examples without (re-)building the libraries that will be loaded. This patchs fixes this issue. For example using the blas domain and MKLCPU backend:
```shell
### With the current develop branch:
$ cmake -Bbuild-blas-cpu -DTARGET_DOMAINS=blas -DENABLE_MKLGPU_BACKEND=OFF -DBUILD_FUNCTIONAL_TESTS=OFF .
$ cd build-blas-cpu && make example_blas_gemm_usm
# Only libonemkl.so and the example are built
$ LD_LIBRARY_PATH="lib:$LD_LIBRARY_PATH" ONEAPI_DEVICE_SELECTOR=*:cpu ./bin/example_blas_gemm_usm
[...]
libonemkl_blas_mklcpu.so: cannot open shared object file: No such file or directory
Caught std::exception during GEMM:
	oneMKL: Couldn\'t load selected backend

### With the fix using the same CMake command:
$ make example_blas_gemm_usm
# libonemkl_blas_mklcpu.so is built
$ LD_LIBRARY_PATH="lib:$LD_LIBRARY_PATH" ONEAPI_DEVICE_SELECTOR=*:cpu ./bin/example_blas_gemm_usm
[...]
BLAS GEMM USM example ran OK.
```
This affects all the domains and backends. Tested with CMake 1.13.5 and CMake 3.27.4.

## All Submissions

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -33,6 +33,7 @@ set(ENABLE_PORTBLAS_BACKEND_NVIDIA_GPU OFF CACHE INTERNAL "")
 set(ONEMKL_GENERATED_INCLUDE_PATH ${CMAKE_CURRENT_BINARY_DIR})
 
 # Build loader and backends for each domain
+add_custom_target(onemkl_backend_libs)
 foreach(domain ${TARGET_DOMAINS})
   add_subdirectory(${domain})
 endforeach()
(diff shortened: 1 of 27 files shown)


Rbiessy:
Test log running on Intel CPU and iGPU: [log.txt](https://github.com/oneapi-src/oneMKL/files/12461602/log.txt)
The examples failed because I was using `ONEAPI_DEVICE_SELECTOR` to filter out another device. They run fine without it: [log_example.txt](https://github.com/oneapi-src/oneMKL/files/12461616/log_example.txt)

Rbiessy:
After some discussions I have switched to another approach using custom target. This is a little nicer. Tests are running.

Also to clarify `onemkl` now depends on all the backend libraries so building a blas example will build all of the backend libraries (but not link against them). Users can either disable unwanted domains or backends or we can add more custom targets in the future if this is an issue.

FMarno (approved):

Rbiessy:
@mkrainiuk could you review this PR? I have hit this issue again as I am adding more backends. I think it's a simple fix and can avoid some confusions.

mkrainiuk (approved):
Thanks for the reminder! It looks good to me.

PR #376, opened 2023-09-04
[RNG][rocRAND] Update rocRAND backend

jle-quel:
# Description

This PR updates the rocRAND backend's distribution functions which used to be unimplemented.

The following distribution functions have been implemented for the `philo` and `mrg` generators, for both API (Buffer and USM):

- `gaussian<float, oneapi::mkl::rng::gaussian_method::icdf>`
- `gaussian<double, oneapi::mkl::rng::gaussian_method::icdf>`
- `lognormal<float, oneapi::mkl::rng::lognormal_method::icdf>`
- `lognormal<double, oneapi::mkl::rng::lognormal_method::icdf>`
- `poisson<std::int32_t, poisson_method::gaussian_icdf_based>`
- `poisson<std::uint32_t, poisson_method::gaussian_icdf_based>`

While updating these functions, a race condition has also been fixed, which makes the `gaussian` distribution functions pass again. The reason for the race condition was due to the asynchronous `host_task` which launched asynchronouss rocRAND computation. The stream created for that computation was not synchronized.

Similar issues raised on cuBLAS: https://github.com/oneapi-src/oneMKL/issues/175#issuecomment-1072290446

Furthermore, the copy constructor has also been implemented and a `seed_` and an `offset_` member has been introduced to the `philo_impl` and `mrg_impl` to be able to track and set the same `seed`, and also align the `offset` of the engines if the `* other` already generated values.

Finally, the constructor for `philo` and `mrg`, and the function `skip_ahead` which takes a `std::initializer_list` are still marked as unimplemented due to rocRAND not supporting `std::initializer_list`.
Also, the `bernoulli` distribution function is still marked as unimplemented due to rocRAND not supporting this distribution method.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Attach a log.
- [x] Have you formatted the code using clang-format?

[example_rng_uniform_usm.log.txt](https://github.com/oneapi-src/oneMKL/files/12514300/example_rng_uniform_usm.log.txt)
[test_main_rng_ct.log.txt](https://github.com/oneapi-src/oneMKL/files/12514301/test_main_rng_ct.log.txt)
[test_main_rng_rt.log.txt](https://github.com/oneapi-src/oneMKL/files/12514302/test_main_rng_rt.log.txt)

PR #388, opened 2023-09-29
[LAPACK][rocSOLVER] Get rocSOLVER backend to build with SYCL2020 changes

sknepper:
# Description

Allows oneMKL to build with rocSOLVER backend and fixes the crashing getrf tests. Now all tests pass.

Changes include:
backend::hip -> backend::ext_oneapi_hip
rocblas/rocsolver header locations
Wait on the queue before the integer array copy kernel in getrf (possibly a bug in the software stack as such a change was not necessary for cuSOLVER in #365)

Fixes #360 

# Checklist

## All Submissions
- [x] Do all unit tests pass locally? Attach a log.  [rocsolver.txt](https://github.com/oneapi-src/oneMKL/files/12757381/rocsolver.txt)
- [x] Have you formatted the code using clang-format?

PR #393, opened 2023-10-04
[LAPACK][rocSOLVER] Update rocSOLVER backend

jle-quel:
# Description

This PR updates the rocSOLVER backend in order to fix the multiple race conditions seen in this backend, as well as some small cleaning and `ifdef` to support other sycl implementation modifications.

Here's the full list of modifications and what they fixed:

- [CLEANING]: The add of an `ifdef` to support the different names of the hip backend given by the different sycl implementations.
- [CLEANING]: Removal of all the `devInfo` logic in the library which calls HIP functions which do not accept and set `devInfo`.
- [RACE-CONDITION]: This PR https://github.com/intel/llvm/pull/11344 re-introduced the call to the extendedDeleter which fixes the segfault that was happening at the end of the `example_lapack_getrs_usm`
- [RACE-CONDITION]: The helper function `lapack_info_check` was not correctly waiting on the `sycl::queue` which created a race condition when reading `devInfo`. The wait has been moved to wait for all overloads, as well as the addition of another `wait` to let the `memcpy` finish.
- [RACE-CONDITION]: All `onemkl_rocsolver_host_task` now uses `ROCSOLVER_ERROR_FUNC_T_SYNC` to synchronize the `HIPStream`. (Similar to https://github.com/oneapi-src/oneMKL/pull/376)
- [RACE-CONDITION]: Added the explicit requirements for `parrallel_for` on `host_task` via `depends_on`.

# Checklist

## All Submissions

[example_lapack_getrs_usm.log.txt](https://github.com/oneapi-src/oneMKL/files/12801629/example_lapack_getrs_usm.log.txt)
[test_main_lapack_ct.log.txt](https://github.com/oneapi-src/oneMKL/files/12801630/test_main_lapack_ct.log.txt)
[test_main_lapack_rt.log.txt](https://github.com/oneapi-src/oneMKL/files/12801631/test_main_lapack_rt.log.txt)

### DIFF
--- a/src/lapack/backends/rocsolver/rocsolver_scope_handle.hpp
+++ b/src/lapack/backends/rocsolver/rocsolver_scope_handle.hpp
@@ -57,7 +57,11 @@ public:
     // will be fixed when SYCL-2020 has been implemented for Pi backend.
     template <typename T, typename U>
     inline T get_mem(U acc) {
+#ifdef SYCL_IMPLEMENTATION_ONEAPI
+        hipDeviceptr_t hipPtr = ih.get_native_mem<sycl::backend::ext_oneapi_hip>(acc);
+#else
         hipDeviceptr_t hipPtr = ih.get_native_mem<sycl::backend::hip>(acc);
+#endif
         return reinterpret_cast<T>(hipPtr);
     }
 };
--- a/src/lapack/backends/rocsolver/rocsolver_scope_handle.hpp
+++ b/src/lapack/backends/rocsolver/rocsolver_scope_handle.hpp
@@ -57,11 +57,7 @@ public:
     // will be fixed when SYCL-2020 has been implemented for Pi backend.
     template <typename T, typename U>
     inline T get_mem(U acc) {
-#ifdef SYCL_IMPLEMENTATION_ONEAPI
         hipDeviceptr_t hipPtr = ih.get_native_mem<sycl::backend::ext_oneapi_hip>(acc);
-#else
-        hipDeviceptr_t hipPtr = ih.get_native_mem<sycl::backend::hip>(acc);
-#endif
         return reinterpret_cast<T>(hipPtr);
     }
 };
--- a/src/lapack/backends/rocsolver/rocsolver_helper.hpp
+++ b/src/lapack/backends/rocsolver/rocsolver_helper.hpp
@@ -254,14 +254,15 @@ inline int get_rocsolver_devinfo(sycl::queue &queue, sycl::buffer<int> &devInfo)
 
 inline int get_rocsolver_devinfo(sycl::queue &queue, const int *devInfo) {
     int dev_info_;
-    queue.wait();
     queue.memcpy(&dev_info_, devInfo, sizeof(int));
+    queue.wait();
     return dev_info_;
 }
 
 template <typename DEVINFO_T>
 inline void lapack_info_check(sycl::queue &queue, DEVINFO_T devinfo, const char *func_name,
                               const char *cufunc_name) {
+    queue.wait();
     const int devinfo_ = get_rocsolver_devinfo(queue, devinfo);
     if (devinfo_ > 0)
         throw oneapi::mkl::lapack::computation_error(
--- a/src/lapack/backends/rocsolver/rocsolver_helper.hpp
+++ b/src/lapack/backends/rocsolver/rocsolver_helper.hpp
@@ -27,8 +27,8 @@
 #define _ROCSOLVER_HELPER_HPP_
 
 #include <CL/sycl.hpp>
-#include <rocblas.h>
-#include <rocsolver.h>
+#include <rocblas/rocblas.h>
+#include <rocsolver/rocsolver.h>
 #include <hip/hip_runtime.h>
 #include <complex>
 
--- a/src/lapack/backends/rocsolver/rocsolver_helper.hpp
+++ b/src/lapack/backends/rocsolver/rocsolver_helper.hpp
@@ -82,15 +82,7 @@ void overflow_check(Index index, Next... indices) {
 class rocsolver_error : virtual public std::runtime_error {
 protected:
     inline const char *rocsolver_error_map(rocblas_status error) {
-        switch (error) {
-            case rocblas_status_success: return "ROCBLAS_STATUS_SUCCESS";
-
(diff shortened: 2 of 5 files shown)


src/lapack/backends/rocsolver/rocsolver_scope_handle.hpp
```diff
@@ -57,7 +57,11 @@ class RocsolverScopedContextHandler {
     // will be fixed when SYCL-2020 has been implemented for Pi backend.
     template <typename T, typename U>
     inline T get_mem(U acc) {
+#ifdef SYCL_IMPLEMENTATION_ONEAPI
```
sknepper: Where is "SYCL_IMPLEMENTATION_ONEAPI" defined?
Also, I see on line ~48 of rocsolver_scope_handle.cpp, that  get_native<sycl::backend::ext_oneapi_hip> is used (without a macro protection). If we are protecting it here, should we also protect it there?

src/lapack/backends/rocsolver/rocsolver_scope_handle.hpp
```diff
@@ -57,7 +57,11 @@ class RocsolverScopedContextHandler {
     // will be fixed when SYCL-2020 has been implemented for Pi backend.
```
sknepper, line 57: Do you know if this comment about a workaround is still needed? And/or if the workaround itself is still needed?

sknepper (commented):
This is great, thanks @jle-quel ! I just had a couple questions, plus one small note.
These changes seem to include the main change from my #388 pull request, plus some extra benefits. However, I had also fixed a few deprecation warnings by including <rocblas/rocblas.h> and <rocsolver/rocsolver.h> instead of just <rocblas.h> and <rocsolver.h> in src/lapack/backends/rocsolver/rocsolver_helper.hpp and src/lapack/backends/rocsolver/rocsolver_task.hpp

It's not required, but you could consider making that change in this PR and I can just close the previous one.

jle-quel:
> This is great, thanks @jle-quel ! I just had a couple questions, plus one small note. These changes seem to include the main change from my #388 pull request, plus some extra benefits. However, I had also fixed a few deprecation warnings by including <rocblas/rocblas.h> and <rocsolver/rocsolver.h> instead of just <rocblas.h> and <rocsolver.h> in src/lapack/backends/rocsolver/rocsolver_helper.hpp and src/lapack/backends/rocsolver/rocsolver_task.hpp
> 
> It's not required, but you could consider making that change in this PR and I can just close the previous one.

Thank you,
Oh, I was unaware of such open work. Of course I'll copy the deprecation warning fixes you made and bring them here :+1:

src/lapack/backends/rocsolver/rocsolver_scope_handle.hpp
jle-quel: Here's where it is defined: https://github.com/intel/llvm/blob/c3ca0941c9fa6b6371a1defd32d338d57d6b7b54/sycl/include/sycl/define_vendors.hpp#L10.

That's a good point, the reason I introduced the `ifdef` was that I was assuming that `backend::hip` was working on another sycl implementation, but as you mention there is also a reference to `sycl::backend::ext_oneapi_hip`.

I'll encapsulate all hip backend reference with the same `ifdef`  :+1:

src/lapack/backends/rocsolver/rocsolver_scope_handle.hpp
jle-quel: No, I'll take a look :+1:

jle-quel:
> /intelci: run

Not sure what the CI does, but running tests on this PR now, will not return the same results I shared in the description.
This PR is dependent on the `ExtendedDeleter` which is not yet merged (reference in the description)

src/lapack/backends/rocsolver/rocsolver_scope_handle.hpp
jle-quel: Right, so DPC++ is `sycl::backend::ext_oneapi_hip` while HipSYCL is `sycl::backend::hip`.
Adding the `ifdef` to all references will make sure that the `rocSOLVER` will compile with differents sycl implementations.

sknepper: Ahhh, thanks!
Currently the rocSOLVER backend only supports the LLVM* compiler (according to the table on the readme page). However, it seems like a good idea to encapsulate for the future, if/once the hipSYCL compiler is supported. I believe the history for why the backend::hip was used is because that worked pre-SYCL2020. The cuSOLVER/rocSOLVER backends haven't yet been updated to work with the SYCL2020 changes (see issue #360 ).

sknepper:
> > /intelci: run
> 
> Not sure what the CI does, but running tests on this PR now, will not return the same results I shared in the description. This PR is dependent on the `ExtendedDeleter` which is not yet merged (reference in the description)

Thanks for the clarification! A passing CI is not currently required (and there are other issues that are unrelated to your PR that are also impacting it), but figured I'd run both that and manual testing, but neither of those have the ExtendedDeleter fix. For me, all of the compile-time dispatching tests passed, while some of the run-time dispatching ones segfaulted (until they eventually just stopped running) - which seems to be expected without the ExtendedDeleter fix.

src/lapack/backends/rocsolver/rocsolver_scope_handle.hpp
jle-quel: Oh right, I was under the impression that it also supported the hipSYCL compiler, thus the `ifdef`.
Should we just remove the `ifdef` and replace `hip` with `ext_oneapi_hip` then?

sknepper: That's fine with me. I'm sure there are multiple changes required elsewhere to fully support the hipSYCL compiler, so we can just do the encapsulation in the future. Thanks!

sknepper (approved):
Thanks, @jle-quel ! These changes look good to me.

jle-quel:
I'll mark this PR as `Ready for review` when the following PRs are merged.

- https://github.com/oneapi-src/unified-runtime/pull/943
- https://github.com/intel/llvm/pull/11527

ericlars (approved):

PR #395, opened 2023-10-16
[tests] skip DP tests on devices without DP support

mkrainiuk:
# Description

PR adds double precision support check to unit tests for skipping double precision tests on devices that do not support it.

I've tested the changes on Intel Arc A770, all double precision tests are skipped as expected, all other tests are passed except for `GemmBatchUsm` for half precision (not related to the changes).

# Checklist

## All Submissions

### DIFF
--- a/tests/unit_tests/include/test_helper.hpp
+++ b/tests/unit_tests/include/test_helper.hpp
@@ -53,6 +53,10 @@
             EXPECT_EQ(res, test_passed); \
     } while (0);
 
+#define CHECK_DOUBLE_ON_DEVICE(d)                                        \
+    if (d->get_info<sycl::info::device::double_fp_config>().size() == 0) \
+    GTEST_SKIP()
+
 #if defined(ENABLE_MKLCPU_BACKEND) || defined(ENABLE_NETLIB_BACKEND)
 #ifdef ENABLE_MKLCPU_BACKEND
 #define TEST_RUN_INTELCPU_SELECT_NO_ARGS(q, func) \
--- a/tests/unit_tests/include/test_helper.hpp
+++ b/tests/unit_tests/include/test_helper.hpp
@@ -55,7 +55,7 @@
 
 #define CHECK_DOUBLE_ON_DEVICE(d)                                        \
     if (d->get_info<sycl::info::device::double_fp_config>().size() == 0) \
-    GTEST_SKIP()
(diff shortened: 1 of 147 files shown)


mmeterel (commented):
Thanks for the PR. Would it be possible to add a macro for the check and use it everywhere? It would make it easier if we need to update it later.

mkrainiuk:
> Thanks for the PR. Would it be possible to add a macro for the check and use it everywhere? It would make it easier if we need to update it later.

This is a good idea, I will update the tests.

mmeterel (approved):
Thanks for the update. Looks good to me.

ericlars (approved):

aelizaro (approved):

Rbiessy:
The DFT tests can also use double precision, should we add the same kind of checks there in this PR? Last time I checked the CI is running very few DFT tests for the GPU device but I don't know why.

mkrainiuk:
> The DFT tests can also use double precision, should we add the same kind of checks there in this PR? Last time I checked the CI is running very few DFT tests for the GPU device but I don't know why.

As I see DFT already has the check for DP support on the device during commit stage https://github.com/oneapi-src/oneMKL/blob/develop/tests/unit_tests/dft/source/descriptor_tests.cpp#L599-L601 
Let me know if I missed something for DFT.

tests/unit_tests/include/test_helper.hpp
```diff
@@ -53,6 +53,10 @@
             EXPECT_EQ(res, test_passed); \
     } while (0);
 
+#define CHECK_DOUBLE_ON_DEVICE(d)                                        \
+    if (d->get_info<sycl::info::device::double_fp_config>().size() == 0) \
+    GTEST_SKIP()
```
Rbiessy: Consider adding a message to explain why the test is skipped like so:

```suggestion
    GTEST_SKIP() << "Double precision is not supported on the device"
```

Rbiessy:
> As I see DFT already has the check for DP support on the device during commit stage https://github.com/oneapi-src/oneMKL/blob/develop/tests/unit_tests/dft/source/descriptor_tests.cpp#L599-L601 Let me know if I missed something for DFT.

Indeed, I missed that, thanks!
I'll make sure to add this to https://github.com/oneapi-src/oneMKL/pull/374 once this is merged.

PR #398, opened 2023-10-18
[BLAS::portBLAS backend] Update nrm2 operator call routine

s-Nick:
# Description

The nrm2 operator available in portBLAS requires that result memory is initialized to zero before performing the routine. I updated the routine call accordingly.

# Checklist

## All Submissions

[nrm2.txt](https://github.com/oneapi-src/oneMKL/files/13002778/nrm2.txt)

### DIFF
--- a/src/blas/backends/portblas/portblas_level1.cxx
+++ b/src/blas/backends/portblas/portblas_level1.cxx
@@ -238,8 +238,9 @@ sycl::event asum(sycl::queue &queue, std::int64_t n, const real_t *x, std::int64
     // before starting the computation.
     auto init_res_val = queue.submit(
         [&](sycl::handler &cgh) { cgh.single_task([=]() { result[0] = real_t(0); }); });
-    init_res_val.wait();
-    CALL_PORTBLAS_USM_FN(::blas::_asum, queue, n, x, incx, result, dependencies);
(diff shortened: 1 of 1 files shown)


src/blas/backends/portblas/portblas_level1.cxx
```diff
@@ -299,6 +305,11 @@ sycl::event nrm2(sycl::queue &queue, std::int64_t n, const std::complex<real_t>
 
 sycl::event nrm2(sycl::queue &queue, std::int64_t n, const real_t *x, std::int64_t incx,
                  real_t *result, const std::vector<sycl::event> &dependencies) {
+    // portBLAS nrm2 implementation requires result to be initializes to zero
+    // before starting the computation.
+    auto init_res_val = queue.submit(
+        [&](sycl::handler &cgh) { cgh.single_task([=]() { result[0] = real_t(0); }); });
+    init_res_val.wait();
```
muhammad-tanvir-1211: Instead of waiting on this event here, I think we can add this event to the `dependencies` vector and that should take care of things.

muhammad-tanvir-1211 (approved):
LGTM, Thanks.

mkrainiuk (approved):
Looks good to me, thank you!

andrewtbarker (approved):

PR #402, opened 2023-10-23
Fix layout enum to match specification

### LINKED ISSUE #400: Layout mismatch
While working on sparse_blas I ran into an issue with the `layout` enum. In OneMKL interface it is defined [here](https://github.com/oneapi-src/oneMKL/blob/develop/include/oneapi/mkl/types.hpp#L44) as
```
enum class layout : char { column_major = 0, row_major = 1, C = 0, R = 1 };
```
whereas in the close-source headers it is defined as:
```
enum class layout : char {
    row_major = 0,
    col_major = 1,
    R = 0,
    C = 1,
};
```

One difference is the name `column_major` vs `col_major`. The [oneMKL interface documentation](https://oneapi-src.github.io/oneAPI-spec/spec/elements/oneMKL/source/architecture/architecture.html) mentions that it should be `col_major`.

The main issue is that the underlying values are swapped. **Should the interface enum be changed to match the close-source headers or is this intentional?** I think changing this will only affect the blas tests but should not change any of the existing logic.

It is an issue for sparse_blas as this is the only domain where the `layout` is a function parameter AFAIK. There is no good solution to properly convert the open-source layout to the close-source one as they have the same name and namespace. The only way would be to swap the `char` values before calling the backend function.

Rbiessy:
# Description

* layout::column_major is renamed to layout::col_major to match the [specification](https://oneapi-src.github.io/oneAPI-spec/spec/elements/oneMKL/source/architecture/architecture.html)
* The values of the enum are swapped to match with close-source layout enum

Fixes https://github.com/oneapi-src/oneMKL/issues/400

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Attach a log. I have triggered an internal CI (see email)

### DIFF
--- a/tests/unit_tests/include/test_helper.hpp
+++ b/tests/unit_tests/include/test_helper.hpp
@@ -220,9 +220,8 @@ class LayoutDeviceNamePrint {
 public:
     std::string operator()(
         testing::TestParamInfo<std::tuple<sycl::device *, oneapi::mkl::layout>> dev) const {
-        std::string layout_name = std::get<1>(dev.param) == oneapi::mkl::layout::column_major
-                                      ? "Column_Major"
-                                      : "Row_Major";
+        std::string layout_name =
+            std::get<1>(dev.param) == oneapi::mkl::layout::col_major ? "Column_Major" : "Row_Major";
         std::string dev_name = std::get<0>(dev.param)->get_info<sycl::info::device::name>();
         for (std::string::size_type i = 0; i < dev_name.size(); ++i) {
             if (!isalnum(dev_name[i]))
--- a/include/oneapi/mkl/types.hpp
+++ b/include/oneapi/mkl/types.hpp
@@ -41,7 +41,7 @@ enum class side : char { left = 0, right = 1, L = 0, R = 1 };
 
 enum class offset : char { row = 0, column = 1, fix = 2, R = 0, C = 1, F = 2 };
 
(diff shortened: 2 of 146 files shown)


tests/unit_tests/include/test_helper.hpp
```diff
@@ -220,9 +220,8 @@ class LayoutDeviceNamePrint {
 public:
     std::string operator()(
         testing::TestParamInfo<std::tuple<sycl::device *, oneapi::mkl::layout>> dev) const {
-        std::string layout_name = std::get<1>(dev.param) == oneapi::mkl::layout::column_major
-                                      ? "Column_Major"
-                                      : "Row_Major";
+        std::string layout_name =
+            std::get<1>(dev.param) == oneapi::mkl::layout::col_major ? "Column_Major" : "Row_Major";
```
Rbiessy, line 224: Note that I have not changed the layout name here. This relates to the namespace to use which uses the full name `column_major` so I've kept it with the full name as well.

Similarly I have not changed the unimplemented exception message `"for column_major layout"` that refers to the namespace.

Rbiessy:
Other than the file `include/oneapi/mkl/types.hpp` the diff was generated by replacing `layout::column_major` to `layout::col_major`.

mmeterel (approved):
Thanks for the PR. Looks good.

gajanan-choudhary (approved):
LGTM!

PR #403, opened 2023-10-23
[SPARSE] Add support for sparse GEMM using MKLCPU

Rbiessy:
# Description

Add support for sparse GEMM using MKLCPU.
This relies on https://github.com/oneapi-src/oneMKL/pull/402 and https://github.com/oneapi-src/oneMKL/pull/374

# Checklist

## All Submissions

## New features


Rbiessy:
Log using a recent close-source MKL and DPCPP builds, all tests are passing: [sparse_gemm_log.txt](https://github.com/oneapi-src/oneMKL/files/13070682/sparse_gemm_log.txt)
I confirmed that with the 2023.2.0 release everything is building and the tests are either skipped or are passing.

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
```diff
@@ -0,0 +1,128 @@
...
+                              detail::get_handle(A_handle), B, columns, ldb, beta, C, ldc);
+}
+
+template <typename fpType>
+std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> gemm(
+    sycl::queue& queue, layout dense_matrix_layout, transpose transpose_A, transpose transpose_B,
+    const fpType alpha, detail::matrix_handle* A_handle, const fpType* B,
+    const std::int64_t columns, const std::int64_t ldb, const fpType beta, fpType* C,
+    const std::int64_t ldc, const std::vector<sycl::event>& dependencies) {
+    // TODO: Remove const_cast with 2024.0 oneMKL release
```
Rbiessy: @gajanan-choudhary even with the latest MKL build this `const_cast` cannot be removed yet. Are you planning to make the `B` argument const? This may be too late for the 2024.0 release right?

Rbiessy: Similar question for `optimize_gemm` which does not seem to be implemented yet.

Rbiessy:
I have rebased the branch and confirmed that this is still building and running fine with oneapi 2023.2.0 as well as with recent DPCPP + MKL builds (see log above).

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
noffermans: Yes, we are planning to add `const` to the `B` argument and implement `optimize_gemm`, but that will not be in the 2024.0 release. I think that you can change the comment and mention the 2024.1 release instead.

```suggestion
    // TODO: Remove const_cast with 2024.1 oneMKL release
```

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
```diff
@@ -17,6 +17,13 @@
 * SPDX-License-Identifier: Apache-2.0
 *******************************************************************************/
 
+sycl::event optimize_gemm(sycl::queue& /*queue*/, transpose /*transpose_A*/,
+                          detail::matrix_handle* /*handle*/,
+                          const std::vector<sycl::event>& /*dependencies*/) {
+    // TODO: Call to optimize_gemm with 2024.0 oneMKL release
```
noffermans: It won't be implemented in 2024.0. Let's target 2024.1 instead.

```suggestion
    // TODO: Call to optimize_gemm with 2024.1 oneMKL release
```

tests/unit_tests/sparse_blas/include/test_common.hpp
```diff
@@ -113,6 +127,28 @@ void rand_vector(std::vector<fpType> &v, std::size_t n) {
...
+    }
+    m.resize(outer_size * ld);
+    rand_scalar<fpType> rand;
+    for (std::size_t i = 0; i < outer_size; ++i) {
+        std::size_t j = 0;
+        for (; j < inner_size; ++j) {
+            m[i * ld + j] = rand(fpRealType(-0.5), fpRealType(0.5));
+        }
+        for (; j < ld; ++j) {
+            m[i * ld + j] = set_fp_value<fpType>()(-1.f, 0.f);
```
noffermans, line 147: Out of curiosity, is there any reason to choose `m = (-1., 0.)` for `outer_size <= j < ld`? Does it make it easier to catch an error?

tests/unit_tests/sparse_blas/source/sparse_gemm_buffer.cpp
```diff
@@ -0,0 +1,233 @@
...
+    intType int_index = (index == oneapi::mkl::index_base::zero) ? 0 : 1;
+    std::size_t opa_nrows = static_cast<std::size_t>(
+        transpose_A == oneapi::mkl::transpose::nontrans ? nrows_A : ncols_A);
+    std::size_t opa_ncols = static_cast<std::size_t>(
+        transpose_A == oneapi::mkl::transpose::nontrans ? ncols_A : nrows_A);
+
+    // Input matrix in CSR format
+    std::vector<intType> ia_host, ja_host;
+    std::vector<fpType> a_host;
+    intType nnz = generate_random_matrix<fpType, intType>(nrows_A, ncols_A, density_A_matrix,
```
noffermans, line 62: Do I understand correctly that `std::rand()` is not seeded and that the same matrices would be generated every time the test suite is run?

tests/unit_tests/sparse_blas/source/sparse_gemm_buffer.cpp
```diff
@@ -0,0 +1,233 @@
...
+                           oneapi::mkl::layout::row_major, transpose_A, transpose_B, fp_one,
+                           fp_zero, ncols_C, ncols_C));
+    // Test int64 indices
+    long long_nrows_A = 27, long_ncols_A = 13, long_ncols_C = 6;
+    long long_ldb = transpose_A == oneapi::mkl::transpose::nontrans ? long_ncols_A : long_nrows_A;
+    long long_ldc = transpose_A == oneapi::mkl::transpose::nontrans ? long_nrows_A : long_ncols_A;
+    EXPECT_TRUEORSKIP(test(dev, long_nrows_A, long_ncols_A, long_ncols_C, density_A_matrix,
+                           index_zero, col_major, transpose_A, transpose_B, fp_one, fp_zero,
+                           long_ldb, long_ldc));
+}
```
noffermans, line 209: Do we plan to add tests with invalid input arguments in the future? I don't think that it should be done                                                                                                                                                                                                                                                                                                                                                 in this PR, I'm just curious :)

noffermans:
This PR looks good to me! I mostly had questions for my own sake.
I'll wait to see whether @gajanan-choudhary wants to review the PR himself before giving approval.

tests/unit_tests/sparse_blas/include/test_common.hpp
Rbiessy: Yes this is just an arbitrary fixed value, outside of the actual input range to catch potential errors.

tests/unit_tests/sparse_blas/source/sparse_gemm_buffer.cpp
Rbiessy: [`std::rand()`](https://en.cppreference.com/w/cpp/numeric/random/rand) is seeded with `std::srand(1)` by default so the inputs are always the same.

tests/unit_tests/sparse_blas/source/sparse_gemm_buffer.cpp
Rbiessy: That's a good point, this would be nice. I've created a task so we may be able to work on that in the future.

tests/unit_tests/sparse_blas/source/sparse_gemm_buffer.cpp
noffermans: Good, thanks!

tests/unit_tests/sparse_blas/source/sparse_gemm_buffer.cpp
noffermans: great!

tests/unit_tests/sparse_blas/include/test_common.hpp
```diff
@@ -113,6 +127,28 @@ void rand_vector(std::vector<fpType> &v, std::size_t n) {
     }
 }
 
+template <typename fpType>
+void rand_matrix(std::vector<fpType> &m, oneapi::mkl::layout layout_val, std::size_t nrows,
```
FMarno, line 131: It seems a bit pointless to have m as an in-out parameter instead of just creating the vector in the function.

FMarno (approved):
I think it looks good overall. Feel free to ignore my comment.

tests/unit_tests/sparse_blas/include/test_common.hpp
Rbiessy: I agree but this is more consistent with other domains, blas has very similar functions but that code is not shared across domains. I have a preference to keep it as is.

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
```diff
@@ -17,6 +17,13 @@
 * SPDX-License-Identifier: Apache-2.0
 *******************************************************************************/
 
+sycl::event optimize_gemm(sycl::queue& /*queue*/, transpose /*transpose_A*/,
+                          detail::matrix_handle* /*handle*/,
+                          const std::vector<sycl::event>& /*dependencies*/) {
+    // TODO: Call to optimize_gemm with 2024.1 oneMKL release
+    return {};
```
gajanan-choudhary: We can't just return an empty `sycl::event`, otherwise the dependence on `dependencies` is lost. We need to return a `sycl::event` that carries over the dependencies. Essentially:
```suggestion
    // TODO: Call to optimize_gemm with 2024.1 oneMKL release
    return queue.submit([=](sycl::handler &cgh) {
        cgh.depends_on(dependencies);
        cgh.host_task([=](){/* Empty kernel */});
    });
```

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
```diff
@@ -67,21 +74,24 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> trsv(
...
-    transpose /*transpose_B*/, const fpType /*alpha*/, detail::matrix_handle* /*A_handle*/,
-    const fpType* /*B*/, const std::int64_t /*columns*/, const std::int64_t /*ldb*/,
-    const fpType /*beta*/, fpType* /*C*/, const std::int64_t /*ldc*/,
-    const std::vector<sycl::event>& /*dependencies*/) {
-    throw unimplemented("SPARSE_BLAS", "gemm");
+    sycl::queue& queue, layout dense_matrix_layout, transpose transpose_A, transpose transpose_B,
+    const fpType alpha, detail::matrix_handle* A_handle, const fpType* B,
+    const std::int64_t columns, const std::int64_t ldb, const fpType beta, fpType* C,
+    const std::int64_t ldc, const std::vector<sycl::event>& dependencies) {
+    // TODO: Remove const_cast with 2024.1 oneMKL release
```
gajanan-choudhary: Removing `const_cast` can't happen in 2024.1 release. It can only happen in a major release of oneMKL (e.g. 2025.0 Gold or 2026.0 Gold) since it will be a breaking change in public headers (adding `const` to the `B` argument in oneMKL header.)

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
Rbiessy: Indeed, thanks! I've re-used your suggestion in https://github.com/oneapi-src/oneMKL/pull/403/commits/e33658c220cbcf61461c7289b770356151b09ccc

tests/unit_tests/sparse_blas/include/test_common.hpp
gajanan-choudhary: Good idea!

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
Rbiessy: That's unfortunate. I've updated the comment in https://github.com/oneapi-src/oneMKL/pull/403/commits/5964f7eafc18206ef96b6a312908acda60f01c1c

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
gajanan-choudhary: Looks good now.

include/oneapi/mkl/sparse_blas/detail/onemkl_sparse_blas_backends.hxx
```diff
@@ -37,6 +37,10 @@ set_csr_data(sycl::queue &queue, matrix_handle_t handle, intType num_rows, intTy
              intType nnz, index_base index, intType *row_ptr, intType *col_ind, fpType *val,
              const std::vector<sycl::event> &dependencies = {});
 
+ONEMKL_EXPORT sycl::event optimize_gemm(sycl::queue &queue, transpose transpose_A,
+                                        matrix_handle_t handle,
+                                        const std::vector<sycl::event> &dependencies = {});
+
```
gajanan-choudhary, line 43: There are 2 `optimize_gemm()` APIs in the oneMKL spec v1.3. There's the one you have here, and there's another one that also has the dense matrix info in the arguments. See [optimize_gemm() in oneMKL spec](https://spec.oneapi.io/versions/1.3-provisional-rev-1/elements/oneMKL/source/domains/spblas/gemmoptimize.html#onemkl-sparse-optimize-gemm). Could you please add both versions in?

gajanan-choudhary (commented):
Once you add the second `optimize_gemm()` API, I will approve. Overall the changes look fine.
I just need to take a closer look at the gemm unit test to understand ldb/ldc/inner dim/outer dim, etc. and the combinations that play out in there, but that is not a blocker for the PR.

include/oneapi/mkl/sparse_blas/detail/onemkl_sparse_blas_backends.hxx
Rbiessy: Thank you, I didn't notice that. I've added it here https://github.com/oneapi-src/oneMKL/pull/403/commits/943c1f430f214787fc577e955f51662d05e18f30
Currently the tests will always call the 2 `optimize_gemm` versions. I did not feel the need to duplicate the number of tests for this.

gajanan-choudhary: We definitely don't want users calling both versions of GEMM one after the other, that's not the intended use case of the APIs (although it would work, but be pointless and possibly less performant). You could use one version of optimize for buffer tests and another one for USM tests, that would be better I think.

gajanan-choudhary: But these are unit tests, so both versions need to be tested. How about adding a loop over a flag and/or additional `EXPECT_TRUEORSKIP()` tests in your list of tests in the CPP files?
```cpp
if (optimize_flag == 1)      {optimize_gemm(/*version 1*/);}
else if (optimize_flag == 2) {optimize_gemm(/*version 2*/);}
else                         { /* skip call to optimize_gemm(); that must work too! */ }
```

src/sparse_blas/function_table.hpp
```diff
@@ -79,6 +79,15 @@ typedef struct {
     FOR_EACH_FP_AND_INT_TYPE(DEFINE_SET_CSR_DATA);
 
     // optimize_*
+    sycl::event (*optimize_gemm_1_input)(sycl::queue &queue, oneapi::mkl::transpose transpose_A,
+                                         oneapi::mkl::sparse::matrix_handle_t handle,
+                                         const std::vector<sycl::event> &dependencies);
+    sycl::event (*optimize_gemm_2_inputs)(sycl::queue &queue, oneapi::mkl::transpose transpose_A,
```
gajanan-choudhary: Not trying to nitpick here, but maybe simply naming them as `optimize_gemm_v1` and `optimize_gemm_v2` would be fine. The `_input` flag is a bit misleading when other `optimize_*` functions don't have that in the name. I won't insist on that though, since it's just 4 lines of code in total.

tests/unit_tests/sparse_blas/source/sparse_gemm_buffer.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+    sycl::event ev_release;
+    oneapi::mkl::sparse::matrix_handle_t handle = nullptr;
+    try {
+        CALL_RT_OR_CT(oneapi::mkl::sparse::init_matrix_handle, main_queue, &handle);
+
+        CALL_RT_OR_CT(oneapi::mkl::sparse::set_csr_data, main_queue, handle, nrows_A, ncols_A, nnz,
+                      index, ia_buf, ja_buf, a_buf);
+
+        CALL_RT_OR_CT(oneapi::mkl::sparse::optimize_gemm, main_queue, transpose_A, handle);
```
gajanan-choudhary: The idea of the `optimize_*` functions is that they can be used optionally to possibly speed up computations. That means we support both "non-optimized" and "optimized" versions of the calls. Our unit tests should cover calling `gemm` with or without calls to `optimize_gemm`.

include/oneapi/mkl/sparse_blas/detail/onemkl_sparse_blas_backends.hxx
Rbiessy: That's not something that we tested for gemv but sure. I have added this for gemv in https://github.com/oneapi-src/oneMKL/pull/403/commits/9fab428f7f6c3214f23710593914c64d03e14d43 and for gemm in https://github.com/oneapi-src/oneMKL/pull/403/commits/0441c3ee899823eb0587f729b3cc448dfe7c6a9c
I've also added a test that uses both optimize_gemm to be complete.

src/sparse_blas/function_table.hpp
Rbiessy: That's fine with me, it's done in https://github.com/oneapi-src/oneMKL/pull/403/commits/94b8b8e73f279d2f35887bce1f0c77116496fa43

tests/unit_tests/sparse_blas/source/sparse_gemm_buffer.cpp
Rbiessy: Done in https://github.com/oneapi-src/oneMKL/pull/403/commits/0441c3ee899823eb0587f729b3cc448dfe7c6a9c

include/oneapi/mkl/sparse_blas/detail/onemkl_sparse_blas_backends.hxx
gajanan-choudhary: Yeah, I missed out on this during my review. Thanks for adding it for GEMV too!

gajanan-choudhary (approved):
LGTM!

PR #407, opened 2023-11-10
[SPARSE] Add support for sparse TRSV with MKLCPU backend

Rbiessy:
# Description

Add support for sparse TRSV using MKLCPU.
Relies on https://github.com/oneapi-src/oneMKL/pull/403

## All Submissions

## New features

### DIFF
--- a/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
+++ b/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
@@ -47,10 +47,11 @@ sycl::event optimize_gemv(sycl::queue& queue, transpose transpose_val,
                                               dependencies);
 }
 
-sycl::event optimize_trsv(sycl::queue& /*queue*/, uplo /*uplo_val*/, transpose /*transpose_val*/,
-                          diag /*diag_val*/, detail::matrix_handle* /*handle*/,
-                          const std::vector<sycl::event>& /*dependencies*/) {
-    throw unimplemented("SPARSE_BLAS", "optimize_trsv");
+sycl::event optimize_trsv(sycl::queue& queue, uplo uplo_val, transpose transpose_val,
+                          diag diag_val, detail::matrix_handle* handle,
+                          const std::vector<sycl::event>& dependencies) {
+    return oneapi::mkl::sparse::optimize_trsv(queue, uplo_val, transpose_val, diag_val, detail::get_handle(handle),
+                                              dependencies);
 }
 
 template <typename fpType>
@@ -71,21 +72,24 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> gemv(
 }
 
 template <typename fpType>
-std::enable_if_t<detail::is_fp_supported_v<fpType>> trsv(sycl::queue& /*queue*/, uplo /*uplo_val*/,
-                                                         transpose /*transpose_val*/,
-                                                         diag /*diag_val*/,
-                                                         detail::matrix_handle* /*A_handle*/,
-                                                         sycl::buffer<fpType, 1>& /*x*/,
-                                                         sycl::buffer<fpType, 1>& /*y*/) {
-    throw unimplemented("SPARSE_BLAS", "trsv");
+std::enable_if_t<detail::is_fp_supported_v<fpType>> trsv(sycl::queue& queue, uplo uplo_val,
+                                                         transpose transpose_val,
+                                                         diag diag_val,
+                                                         detail::matrix_handle* A_handle,
+                                                         sycl::buffer<fpType, 1>& x,
+                                                         sycl::buffer<fpType, 1>& y) {
+    oneapi::mkl::sparse::trsv(queue, uplo_val, transpose_val, diag_val, detail::get_handle(A_handle), x, y);
 }
 
 template <typename fpType>
 std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> trsv(
-    sycl::queue& /*queue*/, uplo /*uplo_val*/, transpose /*transpose_val*/, diag /*diag_val*/,
-    detail::matrix_handle* /*A_handle*/, const fpType* /*x*/, fpType* /*y*/,
-    const std::vector<sycl::event>& /*dependencies*/) {
-    throw unimplemented("SPARSE_BLAS", "trsv");
+    sycl::queue& queue, uplo uplo_val, transpose transpose_val, diag diag_val,
+    detail::matrix_handle* A_handle, const fpType* x, fpType* y,
+    const std::vector<sycl::event>& dependencies) {
+    // TODO: Remove const_cast with 2024.1 oneMKL release
+    return oneapi::mkl::sparse::trsv(queue, uplo_val, transpose_val, diag_val,
+                                     detail::get_handle(A_handle), const_cast<fpType*>(x), y,
+                                     dependencies);
 }
 
 template <typename fpType>
--- a/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
+++ b/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
@@ -51,9 +51,8 @@ sycl::event optimize_trsv(sycl::queue& queue, uplo uplo_val, transpose transpose
                           detail::matrix_handle* handle,
                           const std::vector<sycl::event>& dependencies) {
     if (transpose_val != transpose::nontrans) {
-      throw mkl::unimplemented(
-                    "sparse_blas/backends/mkl", __FUNCTION__,
-                    "Transposed or conjugate trsv is not supported");
+        throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
+                                 "Transposed or conjugate trsv is not supported");
     }
     return oneapi::mkl::sparse::optimize_trsv(queue, uplo_val, transpose_val, diag_val,
                                               detail::get_handle(handle), dependencies);
@@ -83,9 +82,8 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>> trsv(sycl::queue& queue, upl
                                                          sycl::buffer<fpType, 1>& x,
                                                          sycl::buffer<fpType, 1>& y) {
     if (transpose_val != transpose::nontrans) {
-      throw mkl::unimplemented(
-                    "sparse_blas/backends/mkl", __FUNCTION__,
-                    "Transposed or conjugate trsv is not supported");
+        throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
+                                 "Transposed or conjugate trsv is not supported");
     }
     oneapi::mkl::sparse::trsv(queue, uplo_val, transpose_val, diag_val,
                               detail::get_handle(A_handle), x, y);
@@ -97,9 +95,8 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> trsv(
     detail::matrix_handle* A_handle, const fpType* x, fpType* y,
     const std::vector<sycl::event>& dependencies) {
     if (transpose_val != transpose::nontrans) {
-      throw mkl::unimplemented(
-                    "sparse_blas/backends/mkl", __FUNCTION__,
-                    "Transposed or conjugate trsv is not supported");
+        throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
+                                 "Transposed or conjugate trsv is not supported");
     }
     // TODO: Remove const_cast in future oneMKL release
     return oneapi::mkl::sparse::trsv(queue, uplo_val, transpose_val, diag_val,
--- a/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
+++ b/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
@@ -47,11 +47,11 @@ sycl::event optimize_gemv(sycl::queue& queue, transpose transpose_val,
                                               dependencies);
 }
 
-sycl::event optimize_trsv(sycl::queue& queue, uplo uplo_val, transpose transpose_val,
-                          diag diag_val, detail::matrix_handle* handle,
+sycl::event optimize_trsv(sycl::queue& queue, uplo uplo_val, transpose transpose_val, diag diag_val,
+                          detail::matrix_handle* handle,
                           const std::vector<sycl::event>& dependencies) {
-    return oneapi::mkl::sparse::optimize_trsv(queue, uplo_val, transpose_val, diag_val, detail::get_handle(handle),
-                                              dependencies);
+    return oneapi::mkl::sparse::optimize_trsv(queue, uplo_val, transpose_val, diag_val,
+                                              detail::get_handle(handle), dependencies);
 }
 
 template <typename fpType>
@@ -73,12 +73,12 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> gemv(
 
 template <typename fpType>
 std::enable_if_t<detail::is_fp_supported_v<fpType>> trsv(sycl::queue& queue, uplo uplo_val,
-                                                         transpose transpose_val,
-                                                         diag diag_val,
+                                                         transpose transpose_val, diag diag_val,
                                                          detail::matrix_handle* A_handle,
                                                          sycl::buffer<fpType, 1>& x,
                                                          sycl::buffer<fpType, 1>& y) {
-    oneapi::mkl::sparse::trsv(queue, uplo_val, transpose_val, diag_val, detail::get_handle(A_handle), x, y);
+    oneapi::mkl::sparse::trsv(queue, uplo_val, transpose_val, diag_val,
+                              detail::get_handle(A_handle), x, y);
 }
 
 template <typename fpType>
--- a/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
+++ b/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
@@ -50,6 +50,11 @@ sycl::event optimize_gemv(sycl::queue& queue, transpose transpose_val,
 sycl::event optimize_trsv(sycl::queue& queue, uplo uplo_val, transpose transpose_val, diag diag_val,
                           detail::matrix_handle* handle,
                           const std::vector<sycl::event>& dependencies) {
+    if (transpose_val != transpose::nontrans) {
+      throw mkl::unimplemented(
+                    "sparse_blas/backends/mkl", __FUNCTION__,
+                    "Transposed or conjugate trsv is not supported");
+    }
     return oneapi::mkl::sparse::optimize_trsv(queue, uplo_val, transpose_val, diag_val,
                                               detail::get_handle(handle), dependencies);
 }
@@ -77,6 +82,11 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>> trsv(sycl::queue& queue, upl
                                                          detail::matrix_handle* A_handle,
                                                          sycl::buffer<fpType, 1>& x,
                                                          sycl::buffer<fpType, 1>& y) {
+    if (transpose_val != transpose::nontrans) {
+      throw mkl::unimplemented(
+                    "sparse_blas/backends/mkl", __FUNCTION__,
+                    "Transposed or conjugate trsv is not supported");
+    }
     oneapi::mkl::sparse::trsv(queue, uplo_val, transpose_val, diag_val,
                               detail::get_handle(A_handle), x, y);
 }
@@ -86,6 +96,11 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> trsv(
     sycl::queue& queue, uplo uplo_val, transpose transpose_val, diag diag_val,
     detail::matrix_handle* A_handle, const fpType* x, fpType* y,
     const std::vector<sycl::event>& dependencies) {
+    if (transpose_val != transpose::nontrans) {
+      throw mkl::unimplemented(
+                    "sparse_blas/backends/mkl", __FUNCTION__,
+                    "Transposed or conjugate trsv is not supported");
+    }
     // TODO: Remove const_cast in future oneMKL release
     return oneapi::mkl::sparse::trsv(queue, uplo_val, transpose_val, diag_val,
                                      detail::get_handle(A_handle), const_cast<fpType*>(x), y,
--- a/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
+++ b/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
@@ -86,7 +86,7 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> trsv(
     sycl::queue& queue, uplo uplo_val, transpose transpose_val, diag diag_val,
     detail::matrix_handle* A_handle, const fpType* x, fpType* y,
     const std::vector<sycl::event>& dependencies) {
-    // TODO: Remove const_cast with 2024.1 oneMKL release
+    // TODO: Remove const_cast in future oneMKL release
     return oneapi::mkl::sparse::trsv(queue, uplo_val, transpose_val, diag_val,
                                      detail::get_handle(A_handle), const_cast<fpType*>(x), y,
                                      dependencies);
--- a/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
+++ b/src/sparse_blas/backends/mkl_common/mkl_operations.cxx
@@ -50,6 +50,7 @@ sycl::event optimize_gemv(sycl::queue& queue, transpose transpose_val,
 sycl::event optimize_trsv(sycl::queue& queue, uplo uplo_val, transpose transpose_val, diag diag_val,
                           detail::matrix_handle* handle,
                           const std::vector<sycl::event>& dependencies) {
+    // TODO: Remove this if condition once Intel oneMKL adds support for trans/conjtrans to optimize_trsv
     if (transpose_val != transpose::nontrans) {
         throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
                                  "Transposed or conjugate trsv is not supported");
@@ -81,6 +82,7 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>> trsv(sycl::queue& queue, upl
                                                          detail::matrix_handle* A_handle,
                                                          sycl::buffer<fpType, 1>& x,
                                                          sycl::buffer<fpType, 1>& y) {
+    // TODO: Remove this if condition once Intel oneMKL adds support for trans/conjtrans to trsv
     if (transpose_val != transpose::nontrans) {
         throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
                                  "Transposed or conjugate trsv is not supported");
@@ -94,6 +96,7 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> trsv(
     sycl::queue& queue, uplo uplo_val, transpose transpose_val, diag diag_val,
     detail::matrix_handle* A_handle, const fpType* x, fpType* y,
     const std::vector<sycl::event>& dependencies) {
+    // TODO: Remove this if condition once Intel oneMKL adds support for trans/conjtrans to trsv
     if (transpose_val != transpose::nontrans) {
         throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
                                  "Transposed or conjugate trsv is not supported");
--- a/tests/unit_tests/include/test_helper.hpp
+++ b/tests/unit_tests/include/test_helper.hpp
@@ -53,6 +53,17 @@
             EXPECT_EQ(res, test_passed); \
     } while (0);
 
+// GTEST_SKIP stops the execution of the program.
+// This macro lets a test use multiple EXPECT_TRUE_OR_FUTURE_SKIP and mark a test as skipped only once at the end.
+#define EXPECT_TRUE_OR_FUTURE_SKIP(a, skip) \
+    do {                                    \
+        int res = a;                        \
+        if (res == test_skipped)            \
+            skip = true;                    \
+        else                                \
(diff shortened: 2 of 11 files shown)


Rbiessy:
Tests using a recent MKL and DPCPP build: [sparse_trsv_log.txt](https://github.com/oneapi-src/oneMKL/files/13320446/sparse_trsv_log.txt)

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
```diff
@@ -48,40 +72,46 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> gemv(
...
 template <typename fpType>
 std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> trsv(
-    sycl::queue& /*queue*/, uplo /*uplo_val*/, transpose /*transpose_val*/, diag /*diag_val*/,
-    detail::matrix_handle* /*A_handle*/, const fpType* /*x*/, fpType* /*y*/,
-    const std::vector<sycl::event>& /*dependencies*/) {
-    throw unimplemented("SPARSE_BLAS", "trsv");
+    sycl::queue& queue, uplo uplo_val, transpose transpose_val, diag diag_val,
+    detail::matrix_handle* A_handle, const fpType* x, fpType* y,
+    const std::vector<sycl::event>& dependencies) {
+    // TODO: Remove const_cast with 2024.1 oneMKL release
```
gajanan-choudhary:
```suggestion
    // TODO: Remove const_cast in future oneMKL release
```

gajanan-choudhary (commented):
TRSV parts look fine to me. Depending on how the `optimize_gemm` discussions go in #374, this PR may or may not need further changes.

tests/unit_tests/sparse_blas/source/sparse_trsv_buffer.cpp
```diff
@@ -0,0 +1,199 @@
+/*******************************************************************************
+* Copyright 2023 Intel Corporation
```
hjabird, line 2: Codeplay copyright? I'm not sure what the policy on this is.

tests/unit_tests/sparse_blas/source/sparse_trsv_buffer.cpp
```diff
@@ -0,0 +1,199 @@
...
+    try {
+        CALL_RT_OR_CT(oneapi::mkl::sparse::init_matrix_handle, main_queue, &handle);
+
+        CALL_RT_OR_CT(oneapi::mkl::sparse::set_csr_data, main_queue, handle, m, m, nnz, index,
+                      ia_buf, ja_buf, a_buf);
+
+        if (use_optimize) {
+            CALL_RT_OR_CT(oneapi::mkl::sparse::optimize_trsv, main_queue, uplo_val, transpose_val,
+                          diag_val, handle);
+        }
```
hjabird, line 94: Why is the interface different on CPU / GPU?

tests/unit_tests/sparse_blas/include/sparse_reference.hpp
```diff
@@ -232,4 +266,32 @@ void prepare_reference_gemm_data(const intType *ia, const intType *ja, const fpT
     }
 }
 
+template <typename fpType, typename intType>
+void prepare_reference_trsv_data(const intType *ia, const intType *ja, const fpType *a, intType m,
+                                 intType a_ind, oneapi::mkl::uplo uplo_val,
+                                 oneapi::mkl::transpose opA, oneapi::mkl::diag diag_val,
+                                 const fpType *x, fpType *y_ref) {
+    std::size_t mu = static_cast<std::size_t>(m);
+    auto dense_a = sparse_to_dense(ia, ja, a, mu, mu, a_ind, opA, diag_val);
```
hjabird, line 275: As I understand, `trsv` solves a system of linear equations. I'm not totally sure what you're doing here, but wouldn't the most simple solution be to generate the x to solve via matrix multiplication first from a generated y?

On host do a matrix multiply with a known *y*:
$$x = A y$$
and then use trsv to try and recover the *y* value.

tests/unit_tests/sparse_blas/source/sparse_trsv_buffer.cpp
```diff
@@ -0,0 +1,199 @@
...
+    // Output and reference dense vectors.
+    // They are both initialized with a dummy value to catch more errors.
+    std::vector<fpType> y_host(mu, -2.0f);
+    std::vector<fpType> y_ref_host(y_host);
+
+    // Shuffle ordering of column indices/values to test sortedness
+    shuffle_data(ia_host.data(), ja_host.data(), a_host.data(), mu);
+
+    auto ia_buf = make_buffer(ia_host);
+    auto ja_buf = make_buffer(ja_host);
```
hjabird, line 78: Does the spec allow these to be mutated by the sparse_blas API?

tests/unit_tests/sparse_blas/source/sparse_trsv_usm.cpp
```diff
@@ -0,0 +1,224 @@
...
+*
+* SPDX-License-Identifier: Apache-2.0
+*******************************************************************************/
+
+#include <algorithm>
+#include <complex>
+#include <cstdlib>
+#include <iostream>
+#include <limits>
+#include <vector>
```
hjabird: Are we sure we need all of these?

tests/unit_tests/sparse_blas/source/sparse_trsv_buffer.cpp
Rbiessy: I've used Codeplay copyright when I wrote the files from scratch. The tests files are heavily inspired from existing tests in oneMKL interface and MKL tests that were approved for being open-sourced so I've used the Intel one here.

tests/unit_tests/sparse_blas/source/sparse_trsv_buffer.cpp
Rbiessy: The interface is the same but currently not using `optimize_trsv` is not supported with MKLCPU and throws an `invalid_argument` exception.
Ideally we should throw an `unimplemented` exception in oneMKL interface in that case but that would complexify the implementation a bit just for this. We would need a wrapper for the handler and detect whether `optimize_trsv` has been called when calling `trsv`. It may be easier to improve this in the future after the spec has been updated.

tests/unit_tests/sparse_blas/include/sparse_reference.hpp
Rbiessy: That's a good idea. In my suggestion I solve the system one variable after the other and re-use each previously calculated variable to solve the next equation. I find that it's simple enough given that the system is triangular.
What I like is that this keeps the same global structure for the tests in the sense that the inputs are initialized randomly at the beginning and the output is computed from this.
I'm not sure it's really worth changing now that it's done unless others have a strong preference to change this?

tests/unit_tests/sparse_blas/source/sparse_trsv_buffer.cpp
Rbiessy: Do you mean is the user allowed to mutate them or is the backend allowed to mutate them? If I understand the [close-source spec](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2023-2/sparse-blas-matrix-handle-contract-between-user.html#USER-AND-LIBRARY-AGREEMENTS-WITH-RESPECT-TO-SPARSE-MATRIX-HANDLE-T-OBJECT) correctly the answer would be no and yes respectively.
That's not something that is described in the open-source spec currently AFAIK.

tests/unit_tests/sparse_blas/source/sparse_trsv_usm.cpp
Rbiessy: Indeed, I'll remove some of them on Monday.

tests/unit_tests/sparse_blas/include/sparse_reference.hpp
hjabird: As you say it probably isn't worth changing now. If we find in the future that we have trouble with numerical accuracy, it might be a solution.

tests/unit_tests/sparse_blas/source/sparse_trsv_buffer.cpp
hjabird: The backend allowed to mutate them. This relates to the DFT domain where mutating the input buffer is allowed! I couldn't find anything on this either.

tests/unit_tests/sparse_blas/source/sparse_trsv_buffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+    std::vector<fpType> x_host;
+    rand_vector(x_host, mu);
+
+    // Output and reference dense vectors.
+    // They are both initialized with a dummy value to catch more errors.
+    std::vector<fpType> y_host(mu, -2.0f);
+    std::vector<fpType> y_ref_host(y_host);
+
+    // Shuffle ordering of column indices/values to test sortedness
+    shuffle_data(ia_host.data(), ja_host.data(), a_host.data(), mu);
```
gajanan-choudhary: It's a miss in oneMKL documentation that we don't clarify sortedness requirements. We'll fix it eventually.
```suggestion

    // Intel oneMKL does not support unsorted data if
    // `sparse::optimize_trsv()` is not called first.
    if (use_optimize) {
        // Shuffle ordering of column indices/values to test sortedness.
        shuffle_data(ia_host.data(), ja_host.data(), a_host.data(), mu);
    }
```

tests/unit_tests/sparse_blas/source/sparse_trsv_buffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+    EXPECT_TRUEORSKIP(test<fpType>(dev, m, density_A_matrix, index_zero, lower, transpose_val,
+                                   oneapi::mkl::diag::unit, use_optimize));
+    // Test int64 indices
+    EXPECT_TRUEORSKIP(test<fpType>(dev, 15L, density_A_matrix, index_zero, lower, transpose_val,
+                                   nonunit, use_optimize));
+    if (!dev->is_cpu()) {
+        // Test without optimize_trsv
+        EXPECT_TRUEORSKIP(test<fpType>(dev, m, density_A_matrix, index_zero, lower, transpose_val,
+                                       nonunit, false));
+    }
```
gajanan-choudhary: Provided you accept the other suggestion for conditional call to `shuffle_data()`, this should work for both CPU and GPU devices.
```suggestion
    // Test without optimize_trsv
    EXPECT_TRUEORSKIP(test<fpType>(dev, m, density_A_matrix, index_zero, lower, transpose_val,
                                       nonunit, false));
```

tests/unit_tests/sparse_blas/source/sparse_trsv_usm.cpp
```diff
@@ -0,0 +1,221 @@
...
+ * @param dev Device to test
+ * @param transpose_val Transpose value for the input matrix
+ */
+template <typename fpType>
+void test_helper(sycl::device *dev, oneapi::mkl::transpose transpose_val) {
+    double density_A_matrix = 0.8;
+    oneapi::mkl::index_base index_zero = oneapi::mkl::index_base::zero;
+    oneapi::mkl::uplo lower = oneapi::mkl::uplo::lower;
+    oneapi::mkl::diag nonunit = oneapi::mkl::diag::nonunit;
+    int m = 5;
```
gajanan-choudhary: I would suggest using a larger matrix with lower density for the tests unless they are really slow. Maybe `m = 277`, a prime number, with about 40 non-zeros per row, so:
```suggestion
    double density_A_matrix = 0.144;
    oneapi::mkl::index_base index_zero = oneapi::mkl::index_base::zero;
    oneapi::mkl::uplo lower = oneapi::mkl::uplo::lower;
    oneapi::mkl::diag nonunit = oneapi::mkl::diag::nonunit;
    int m = 277;
```

gajanan-choudhary (commented):
All files other than the unit tests are fine. I have a few suggestions for the tests.

tests/unit_tests/sparse_blas/source/sparse_trsv_usm.cpp
```diff
@@ -0,0 +1,221 @@
...
+    // Test unit diagonal matrix
+    EXPECT_TRUEORSKIP(test<fpType>(dev, m, density_A_matrix, index_zero, lower, transpose_val,
+                                   oneapi::mkl::diag::unit, use_optimize));
+    // Test int64 indices
+    EXPECT_TRUEORSKIP(test<fpType>(dev, 15L, density_A_matrix, index_zero, lower, transpose_val,
+                                   nonunit, use_optimize));
+    if (!dev->is_cpu()) {
+        // Test without optimize_trsv
+        EXPECT_TRUEORSKIP(test<fpType>(dev, m, density_A_matrix, index_zero, lower, transpose_val,
+                                       nonunit, false));
```
gajanan-choudhary: Also, please test this with both lower and upper.

tests/unit_tests/sparse_blas/source/sparse_trsv_usm.cpp
Rbiessy: Done in https://github.com/oneapi-src/oneMKL/pull/407/commits/9f420d9e8bb6df5e64bcc7c8b4823247ea33474a

tests/unit_tests/sparse_blas/source/sparse_trsv_buffer.cpp
Rbiessy: Thank you! Fixed in https://github.com/oneapi-src/oneMKL/pull/407/commits/a762d79de03400ebf05b3ede85db6bb6242d9940

tests/unit_tests/sparse_blas/source/sparse_trsv_buffer.cpp
Rbiessy: Yes, that's better. It's fixed in https://github.com/oneapi-src/oneMKL/pull/407/commits/a762d79de03400ebf05b3ede85db6bb6242d9940

tests/unit_tests/sparse_blas/source/sparse_trsv_usm.cpp
Rbiessy: Done in https://github.com/oneapi-src/oneMKL/pull/407/commits/3d1f1e425b9ed95196e70b4778398c53e65cf0b4

tests/unit_tests/sparse_blas/source/sparse_trsv_usm.cpp
Rbiessy: Done in https://github.com/oneapi-src/oneMKL/pull/407/commits/a9e4a3096690eb931ead246fd332d62006b17cf2

Rbiessy:
Note for the reviewers, I have made a change that affects the other sparse_blas tests to introduce `EXPECT_TRUE_OR_FUTURE_SKIP`. I was concerned about the number of tests that weren't properly run due to `GTEST_SKIP` that would interrupt the test on the first configuration that is not supported.
Log running all the tests on CPU using the 2024.0 package: [log_sparse_blas-cpu.txt](https://github.com/oneapi-src/oneMKL/files/13427606/log_sparse_blas-cpu.txt)

gajanan-choudhary:
> Note for the reviewers, I have made a change that affects the other sparse_blas tests to introduce `EXPECT_TRUE_OR_FUTURE_SKIP`. I was concerned about the number of tests that weren't properly run due to `GTEST_SKIP` that would interrupt the test on the first configuration that is not supported. Log running all the tests on CPU using the 2024.0 package: [log_sparse_blas-cpu.txt](https://github.com/oneapi-src/oneMKL/files/13427606/log_sparse_blas-cpu.txt)

@Rbiessy, it looks from your [log_sparse_blas-cpu.txt](https://github.com/oneapi-src/oneMKL/files/13427606/log_sparse_blas-cpu.txt) file that the only tests that ran were GEMV. All GEMM and TRSV tests appear to have been skipped. Shouldn't all of them be running and passing?

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
```diff
@@ -47,10 +47,15 @@ sycl::event optimize_gemv(sycl::queue& queue, transpose transpose_val,
...
-                          diag /*diag_val*/, detail::matrix_handle* /*handle*/,
-                          const std::vector<sycl::event>& /*dependencies*/) {
-    throw unimplemented("SPARSE_BLAS", "optimize_trsv");
+sycl::event optimize_trsv(sycl::queue& queue, uplo uplo_val, transpose transpose_val, diag diag_val,
+                          detail::matrix_handle* handle,
+                          const std::vector<sycl::event>& dependencies) {
+    if (transpose_val != transpose::nontrans) {
+        throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
+                                 "Transposed or conjugate trsv is not supported");
+    }
```
gajanan-choudhary, line 57: Add `TODO` comment and replace `trsv` in thrown text with `optimize_trsv`:
```suggestion
    // TODO: Remove this if condition once Intel oneMKL adds support for
    // trans/conjtrans to optimize_trsv
    if (transpose_val != transpose::nontrans) {
        throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
                                 "Transposed or conjugate optimize_trsv is not supported");
    }
```

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
```diff
@@ -71,21 +76,32 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> gemv(
...
-    throw unimplemented("SPARSE_BLAS", "trsv");
+std::enable_if_t<detail::is_fp_supported_v<fpType>> trsv(sycl::queue& queue, uplo uplo_val,
+                                                         transpose transpose_val, diag diag_val,
+                                                         detail::matrix_handle* A_handle,
+                                                         sycl::buffer<fpType, 1>& x,
+                                                         sycl::buffer<fpType, 1>& y) {
+    if (transpose_val != transpose::nontrans) {
+        throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
+                                 "Transposed or conjugate trsv is not supported");
+    }
```
gajanan-choudhary, line 89: Add `TODO` comment:
```suggestion
    // TODO: Remove this if condition once Intel oneMKL adds support for
    // trans/conjtrans to trsv
    if (transpose_val != transpose::nontrans) {
        throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
                                 "Transposed or conjugate trsv is not supported");
    }
```

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
```diff
@@ -71,21 +76,32 @@ std::enable_if_t<detail::is_fp_supported_v<fpType>, sycl::event> gemv(
...
-    detail::matrix_handle* /*A_handle*/, const fpType* /*x*/, fpType* /*y*/,
-    const std::vector<sycl::event>& /*dependencies*/) {
-    throw unimplemented("SPARSE_BLAS", "trsv");
+    sycl::queue& queue, uplo uplo_val, transpose transpose_val, diag diag_val,
+    detail::matrix_handle* A_handle, const fpType* x, fpType* y,
+    const std::vector<sycl::event>& dependencies) {
+    if (transpose_val != transpose::nontrans) {
+        throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
+                                 "Transposed or conjugate trsv is not supported");
+    }
```
gajanan-choudhary, line 103: Add `TODO` comment:
```suggestion
    // TODO: Remove this if condition once Intel oneMKL adds support for
    // trans/conjtrans to trsv
    if (transpose_val != transpose::nontrans) {
        throw mkl::unimplemented("sparse_blas/backends/mkl", __FUNCTION__,
                                 "Transposed or conjugate trsv is not supported");
    }
```

gajanan-choudhary (commented):
LGTM except for the `EXPECT_TRUE_OR_FUTURE_SKIP` related changes which make the test logs misleading because:
```
    bool skip = test_helper_transpose<fpType>(GetParam());
    if (skip) {
        // Mark that some tests were skipped
        GTEST_SKIP();
    }
```
If there are 10 input combinations that are run and 9 pass, 1 is skipped, then the entire test is marked as skipped. There's a [way to also pass a string to `GTEST_SKIP()`](https://stackoverflow.com/questions/7208070/googletest-how-to-skip-a-test). We should use that to clearly indicate how many tests passed, how many tests failed, and how many were skipped.

gajanan-choudhary:
> > Note for the reviewers, I have made a change that affects the other sparse_blas tests to introduce `EXPECT_TRUE_OR_FUTURE_SKIP`. I was concerned about the number of tests that weren't properly run due to `GTEST_SKIP` that would interrupt the test on the first configuration that is not supported. Log running all the tests on CPU using the 2024.0 package: [log_sparse_blas-cpu.txt](https://github.com/oneapi-src/oneMKL/files/13427606/log_sparse_blas-cpu.txt)
> 
> @Rbiessy, it looks from your [log_sparse_blas-cpu.txt](https://github.com/oneapi-src/oneMKL/files/13427606/log_sparse_blas-cpu.txt) file that the only tests that ran were GEMV. All GEMM and TRSV tests appear to have been skipped. Shouldn't all of them be running and passing?

I see what happened there. Multiple tests are run, but some may throw some acceptable exception such as `unimplemented` and is therefore skipped, but we mark the entire set in the test as "Skipped" with a `GTEST_SKIP()` call in the end. This is misleading, case in point being my own comment above and confusion before reviewing the changes. Can we come up with a better long-term solution for this? Each gtest run could report how many subtests were run, how many passed, how many were skipped, and how many failed. Maybe 3 variables could be used to track that along to summarize in the end?

For that matter, I don't think a throwing an acceptable exception such as `unimplemented` should be considered a skipped test, we should mark it as passed. Any opinions here, @spencerpatty?

gajanan-choudhary:
> For that matter, I don't think a throwing an acceptable exception such as `unimplemented` should be considered a skipped test, we should mark it as passed. Any opinions here, @spencerpatty?

On second thought, I take that back. I myself don't know what's the appropriate thing to do here. It is a skipped test, but if we mark that set as skipped, then it's misleading. If we mark it as passed, then it may appear to some as if there's no exception being thrown and that the input combination is supported. Maybe counting passed, skipped, failed tests is the way to go? Is there a better way?

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
Rbiessy: Done in https://github.com/oneapi-src/oneMKL/pull/407/commits/e49f285b7e41f354c6423359238ce2ace2d9b50d

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
Rbiessy: Done in https://github.com/oneapi-src/oneMKL/pull/407/commits/e49f285b7e41f354c6423359238ce2ace2d9b50d

src/sparse_blas/backends/mkl_common/mkl_operations.cxx
Rbiessy: Done in https://github.com/oneapi-src/oneMKL/pull/407/commits/e49f285b7e41f354c6423359238ce2ace2d9b50d

Rbiessy:
@gajanan-choudhary regarding the issue that you mention, I agree that this is not ideal but this is the current way the tests are done for other domains as well.
The proper solution would be to split the tests so that each test runs only 1 configuration. I considered improving this but I didn't want to spend more time than needed.
I was able to confirm that none of the tests are skipped with a recent MKL build.

hjabird (approved):

gajanan-choudhary:
> I was able to confirm that none of the tests are skipped with a recent MKL build.

If none of the tests are skipped, then why do the log files such as [log_sparse_blas-cpu.txt](https://github.com/oneapi-src/oneMKL/files/13427606/log_sparse_blas-cpu.txt) you posted earlier have "Skipped" written in them?

I understand that it is not that easy to have each test run just 1 configuration; that's fine, but at least can we have to logs print how many configurations were run, how many passed/failed/skipped? There's an option to [pass a custom print to gtest_skip()](https://stackoverflow.com/questions/7208070/googletest-how-to-skip-a-test), for instance. We could cascade the count of test statuses to the final return point.

tests/unit_tests/include/test_helper.hpp
```diff
@@ -53,6 +53,17 @@
...
+// This macro lets a test use multiple EXPECT_TRUE_OR_FUTURE_SKIP and mark a test as skipped only once at the end.
+#define EXPECT_TRUE_OR_FUTURE_SKIP(a, skip) \
+    do {                                    \
+        int res = a;                        \
+        if (res == test_skipped)            \
+            skip = true;                    \
+        else                                \
+            EXPECT_EQ(res, test_passed);    \
+    } while (0);
+
```
gajanan-choudhary: I think it's not that difficult to change this macro a bit to total things up into passed/failed/skipped tests.
```suggestion
#define EXPECT_TRUE_OR_FUTURE_SKIP(a, skip) \
    do {                                    \
        int res = a;                        \
        if (res == test_skipped)            \
            skip = true;                    \
            skipped_count += 1;             \
        else {                              \
            passed_count += 1;              \
            EXPECT_EQ(res, test_passed);    \
        }                                   \
    } while (0);

```
You might have to return a `pair<int, int>` from your `test_helper` functions though to propagate both the status and the counts. Finally in your `gtest_skip()` call, you can have:
```cpp
    if (skip) {
        // Mark that some tests were skipped
        GTEST_SKIP() << "Passed: " << passed_count << ", Skipped: " << skipped_count << std::endl;
    }
```

Rbiessy: I have done that in https://github.com/oneapi-src/oneMKL/pull/407/commits/7954345dcda900b24c1dcf4678b060ac3dd63303

Rbiessy:
@gajanan-choudhary, the log I attached was using oneapi 2024.0. This one has skipped configurations for gemm when `transpose_B != nontrans` and for trsv if the matrix is transposed because the backend is throwing unsupported exceptions. If I use a more recent nightly build of MKL none of the tests are skipped on CPU.
I've applied your suggestion but it's not working well with ctest. I don't know why the skipped messages are not printed even with the extra-verbose flag. The skipped messages are printed if we run a binary `./bin/test_main_sparse_blas_ct` directly and use the `--terse_output` flag.
Here are the logs using ctest [log_cpu_2024_0.txt](https://github.com/oneapi-src/oneMKL/files/13488573/log_cpu_2024_0.txt) and another one with the the `terse_output` [log_cpu_2024_0_terse_output.txt](https://github.com/oneapi-src/oneMKL/files/13488574/log_cpu_2024_0_terse_output.txt).
I was not able to run this last change on a nightly build anymore due to permissions issue. I don't think this is blocking this PR.

tests/unit_tests/sparse_blas/source/sparse_gemm_buffer.cpp
```diff
@@ -214,41 +233,67 @@ void test_helper(sycl::device *dev, oneapi::mkl::transpose transpose_A,
...
 
 TEST_P(SparseGemmBufferTests, RealSinglePrecision) {
     using fpType = float;
-    test_helper_transpose<fpType>(GetParam());
+    int num_passed = 0, num_skipped = 0;
+    test_helper_transpose<fpType>(GetParam(), num_passed, num_skipped);
+    if (num_skipped > 0) {
+        // Mark that some tests were skipped
+        GTEST_SKIP() << "Passed: " << num_passed << ", Skipped: " << num_skipped
+                     << " configurations." << std::endl;
```
gajanan-choudhary, line 260:
```suggestion
        // Mark that some tests were skipped
        // This may not print/work well with ctest. However, running the binary
        // files of the tests directly with the `--terse-output` flag prints this
        // properly. E.g., ./bin/test_main_sparse_blas_ct --terse-output
        GTEST_SKIP() << "Passed: " << num_passed << ", Skipped: " << num_skipped
                     << " configurations." << std::endl;
```

gajanan-choudhary (approved):
LGTM!

I do have a final minor suggestion of adding a comment above the `gtest_skip()` call indicating that the printing may not work well with `ctest`, in which case running the binary with the `--terse-output` flag should print things. Just so that this info doesn't get lost in history if someone tries digging in!

Thanks for going above and beyond in seeing through my change requests on this PR. I really appreciate your effort in making the product better! Great work on the PR!

tests/unit_tests/sparse_blas/source/sparse_gemm_buffer.cpp
Rbiessy: This would need to be added for every `GTEST_SKIP` occurrences if we want a chance for someone to find it and to be consistent. I'm not sure if there is a single place that would be suitable for this comment so I would rather not add it.

gajanan-choudhary: In that case, maybe add it above `EXPECT_TRUE_OR_FUTURE_SKIP` macro definition? Or in some appropriate CMake file maybe around the location where the `ctest` tests are defined? Any other location you can think of? Worst case, there's the option of converting the `if` condition into a macro, so that we could have the comment be in a single place above the macro, but I'd rather avoid that if possible.

gajanan-choudhary: I'm also fine with it being added in the first `GTEST_SKIP()` occurrence in each individual file (which would be 6 files right now, I believe).

Rbiessy: I have added a more generic comment at the top of test_helper.hpp in https://github.com/oneapi-src/oneMKL/pull/407/commits/c558637aeb73c0b4d8c6730d9fc10cacc2e5b7b3

The issue is not specific to sparse_blas and a message is also used with `GTEST_SKIP()` in the macro `CHECK_DOUBLE_ON_DEVICE`.

gajanan-choudhary (approved):

Issue #396, opened 2023-10-17
Support external workspace and workspace size query in DFT domain


zhiweij1:
# Summary
Support external workspace and workspace size query in DFT domain.

# Problem statement
Currently, in the Intel(R) oneAPI Math Kernel Library, we can use external workspace and query workspace size and estimated workspace size.

# Preferred solution
Support those functionalities in the opensource oneMKL interfaces.

Rbiessy:
Thank you for the issue. We'll start working on this in the coming weeks.

hjabird:
I'm looking at this at the moment. A challenging aspect is providing pre-commit estimates of workspace size. This is because *rocFFT* (and perhaps other future backends) do not support workspace estimates, only exact workspace requirements after committing.

* What is the usecase for pre-commit workspace estimates?
  * Do we need pre-commit estimates at all?
* Do these estimates need to be conservative?

Thanks!

zhiweij1:
Hi @hjabird , I noticed that https://github.com/ROCmSoftwarePlatform/hipFFT/blob/ef5eb0d5facf9d997b00bb0b94c8e37791b72ce4/library/src/amd_detail/hipfft.cpp#L1150 uses `hipfftGetSize1d` to implement `hipfftEstimate1d` which is the migrated result of `cufftEstimate1d` with HIPFIY.

It just queries the actual workspace size after creating a plan and destroys the plan before return.

hjabird:
Hi @zhiweij1, that is an option, but creating a plan just to query workspace size is an expensive thing to do. Especially if you then create that exact same plan once again on commit!

At the moment I don't understand the scenario in which you use an estimate instead of committing a descriptor and getting the exact workspace requirement from that. In such a scenario, does the estimate *need* to be conservative (if not, we could introduce some kind of heuristic for rocFFT's workspace requirements, or just return 0)?

The specification may be a better place to continue this discussion. I've created a spec issue https://github.com/oneapi-src/oneAPI-spec/issues/508 and also a draft PR in the spec https://github.com/oneapi-src/oneAPI-spec/pull/509

hjabird:
@zhiweij1 I've had some feedback on the above linked PR. It would be great to have your feedback too, especially since it doesn't include the estimation feature right now. It doesn't included this feature since I'm not sure how it would be used (and so what to specify), as stated in my above comment.

zhiweij1:
[...]

PR #411, opened 2023-11-17
[DFT] Implement external workspace according to draft spec

hjabird:
# Description

This PR implements the draft specification for external workspace at https://github.com/oneapi-src/oneAPI-spec/pull/509, which relates to oneMKL issue https://github.com/oneapi-src/oneMKL/issues/396.

# Checklist

## All Submissions

## New interfaces

it was accepted? # (RFC) **Draft PR for oneAPI spec linked, but this is not yet merged**

## New features

Logs:
* [inteltest.txt](https://github.com/oneapi-src/oneMKL/files/13473985/inteltest.txt)
* [testamd.txt](https://github.com/oneapi-src/oneMKL/files/13473986/testamd.txt)
* [testnvidia.txt](https://github.com/oneapi-src/oneMKL/files/13473989/testnvidia.txt)

### DIFF
--- a/include/oneapi/mkl/dft/detail/commit_impl.hpp
+++ b/include/oneapi/mkl/dft/detail/commit_impl.hpp
@@ -139,7 +139,7 @@ public:
      * @param function_name The function name to user in generated exceptions.
     */
     template <typename... ArgTs>
-    inline void compute_call_throw(const char *function_name) {
+    void compute_call_throw(const char *function_name) {
         external_workspace_helper_.template compute_call_throw<ArgTs...>(function_name);
     }
 
@@ -148,8 +148,23 @@ public:
      * @param function_name The function name to user in generated exceptions.
      * @param cgh The command group handler to associate the accessor with.
     */
-    inline void get_workspace_buffer_access_if_rqd(const char *function_name, sycl::handler &cgh) {
-        external_workspace_helper_.get_workspace_buffer_access_if_rqd(function_name, cgh);
+    void add_buffer_dependency_if_rqd(const char *function_name, sycl::handler &cgh) {
+        external_workspace_helper_.add_buffer_dependency_if_rqd(function_name, cgh);
+    }
+
+    /** If WORKSPACE_EXTERNAL is set, depend on the last USM workspace event added via set_last_usm_workspace_event.
+     * @param cgh The command group handler to associate the accessor with.
+    */
+    void depend_on_last_usm_workspace_event_if_rqd(sycl::handler &cgh) {
+        external_workspace_helper_.depend_on_last_usm_workspace_event(cgh);
+    }
+
+    /** If WORKSPACE_EXTERNAL is set, store the given event internally to allow it to be depended upon by
+     * subsequent calls to depend_on_last_usm_workspace_event.
+     * @param sycl_event The last usage of the USM workspace.
+    */
+    void set_last_usm_workspace_event(sycl::event &sycl_event) {
+        external_workspace_helper_.set_last_usm_workspace_event(sycl_event);
     }
 
 protected:
--- a/include/oneapi/mkl/dft/detail/commit_impl.hpp
+++ b/include/oneapi/mkl/dft/detail/commit_impl.hpp
@@ -149,7 +149,7 @@ public:
      * @param cgh The command group handler to associate the accessor with.
     */
     void add_buffer_workspace_dependency_if_rqd(const char *function_name, sycl::handler &cgh) {
-        external_workspace_helper_.add_buffer_workspace_dependency_if_rqd(function_name, cgh);
+        external_workspace_helper_.add_buffer_dependency_if_rqd(function_name, cgh);
     }
 
     /** If WORKSPACE_EXTERNAL is set, depend on the last USM workspace event added via set_last_usm_workspace_event.
--- a/include/oneapi/mkl/dft/detail/commit_impl.hpp
+++ b/include/oneapi/mkl/dft/detail/commit_impl.hpp
@@ -84,6 +84,9 @@ public:
     // set_workspace should be overrided for any backend that enables external workspaces.
     // If these are overrided, get_workspace_external_bytes_impl must also be overrided.
     // For backends that do not support external workspaces, these functions do not need to be overrided.
+    // When not overrided, external workspace support is faked: an external workspace can be set, 
+    // and errors will be generated according to the specificiation, 
+    // but the required workspace size will always be zero, and any given workspace will not actually be used.
     virtual void set_workspace(scalar_type *usm_workspace) {
         external_workspace_helper_.set_workspace_throw(*this, usm_workspace);
     };
--- a/include/oneapi/mkl/dft/detail/commit_impl.hpp
+++ b/include/oneapi/mkl/dft/detail/commit_impl.hpp
@@ -148,8 +148,8 @@ public:
      * @param function_name The function name to user in generated exceptions.
      * @param cgh The command group handler to associate the accessor with.
     */
-    void add_buffer_dependency_if_rqd(const char *function_name, sycl::handler &cgh) {
-        external_workspace_helper_.add_buffer_dependency_if_rqd(function_name, cgh);
+    void add_buffer_workspace_dependency_if_rqd(const char *function_name, sycl::handler &cgh) {
+        external_workspace_helper_.add_buffer_workspace_dependency_if_rqd(function_name, cgh);
     }
 
     /** If WORKSPACE_EXTERNAL is set, depend on the last USM workspace event added via set_last_usm_workspace_event.
--- a/include/oneapi/mkl/dft/detail/commit_impl.hpp
+++ b/include/oneapi/mkl/dft/detail/commit_impl.hpp
@@ -156,15 +156,15 @@ public:
      * @param cgh The command group handler to associate the accessor with.
     */
     void depend_on_last_usm_workspace_event_if_rqd(sycl::handler &cgh) {
-        external_workspace_helper_.depend_on_last_usm_workspace_event(cgh);
+        external_workspace_helper_.depend_on_last_usm_workspace_event_if_rqd(cgh);
     }
 
     /** If WORKSPACE_EXTERNAL is set, store the given event internally to allow it to be depended upon by
      * subsequent calls to depend_on_last_usm_workspace_event.
      * @param sycl_event The last usage of the USM workspace.
     */
-    void set_last_usm_workspace_event(sycl::event &sycl_event) {
-        external_workspace_helper_.set_last_usm_workspace_event(sycl_event);
+    void set_last_usm_workspace_event_if_rqd(sycl::event &sycl_event) {
+        external_workspace_helper_.set_last_usm_workspace_event_if_rqd(sycl_event);
     }
 
 protected:
--- a/include/oneapi/mkl/dft/detail/commit_impl.hpp
+++ b/include/oneapi/mkl/dft/detail/commit_impl.hpp
@@ -81,9 +81,9 @@ public:
         return external_workspace_helper_.get_rqd_workspace_bytes(*this);
     };
 
-    // Set workspace by be overrided for any backend that actually uses the workspace.
+    // set_workspace should be overrided for any backend that enables external workspaces.
     // If these are overrided, get_workspace_external_bytes_impl must also be overrided.
-    // For backends that do not support external workspace, these functions do not need to be overrided.
+    // For backends that do not support external workspaces, these functions do not need to be overrided.
     virtual void set_workspace(scalar_type *usm_workspace) {
         external_workspace_helper_.set_workspace_throw(*this, usm_workspace);
     };
--- a/include/oneapi/mkl/dft/detail/commit_impl.hpp
+++ b/include/oneapi/mkl/dft/detail/commit_impl.hpp
@@ -27,6 +27,7 @@
 #endif
 
 #include "descriptor_impl.hpp"
+#include "external_workspace_helper.hpp"
 
 namespace oneapi::mkl {
 enum class backend;
@@ -48,7 +49,16 @@ public:
     using bwd_type = typename descriptor_info<descriptor_type>::backward_type;
     using scalar_type = typename descriptor_info<descriptor_type>::scalar_type;
 
-    commit_impl(sycl::queue queue, mkl::backend backend) : queue_(queue), backend_(backend) {}
+protected:
+    external_workspace_helper<prec, dom> external_workspace_helper_;
+
+public:
+    commit_impl(sycl::queue queue, mkl::backend backend,
+                const dft::detail::dft_values<prec, dom> &config_values)
+            : queue_(queue),
+              backend_(backend),
+              external_workspace_helper_(config_values.workspace_placement ==
+                                         dft::detail::config_value::WORKSPACE_EXTERNAL) {}
 
     // rule of three
     commit_impl(const commit_impl &other) = delete;
@@ -66,6 +76,20 @@ public:
     virtual void *get_handle() noexcept = 0;
 
     virtual void commit(const dft_values<prec, dom> &) = 0;
+
+    inline std::int64_t get_workspace_external_bytes() {
+        return external_workspace_helper_.get_rqd_workspace_bytes(*this);
+    };
+
+    // Set workspace by be overrided for any backend that actually uses the workspace.
+    // If these are overrided, get_workspace_external_bytes_impl must also be overrided.
+    // For backends that do not support external workspace, these functions do not need to be overrided.
+    virtual void set_workspace(scalar_type *usm_workspace) {
+        external_workspace_helper_.set_workspace_throw(*this, usm_workspace);
+    };
+    virtual void set_workspace(sycl::buffer<scalar_type> &buffer_workspace) {
+        external_workspace_helper_.set_workspace_throw(*this, buffer_workspace);
+    };
 
     virtual void forward_ip_cc(descriptor_type &desc, sycl::buffer<fwd_type, 1> &inout) = 0;
     virtual void forward_ip_rr(descriptor_type &desc, sycl::buffer<scalar_type, 1> &inout_re,
@@ -108,6 +132,33 @@ public:
     virtual sycl::event backward_op_rr(descriptor_type &desc, scalar_type *in_re,
                                        scalar_type *in_im, scalar_type *out_re, scalar_type *out_im,
                                        const std::vector<sycl::event> &dependencies) = 0;
+
+    /** For compute calls, throw errors for the external workspace as required.
+     * @tparam ArgTs The non-descriptor arg(s) for the compute call. First one is used to check
+     * buffer or USM call.
+     * @param function_name The function name to user in generated exceptions.
+    */
+    template <typename... ArgTs>
+    inline void compute_call_throw(const char *function_name) {
+        external_workspace_helper_.template compute_call_throw<ArgTs...>(function_name);
+    }
+
+    /** Create an accessor out of the workspace buffer when required, to ensure correct dependency
+     *  management for the buffer. To be used by backends that don't natively support sycl::buffers.
+     * @param function_name The function name to user in generated exceptions.
+     * @param cgh The command group handler to associate the accessor with.
+    */
+    inline void get_workspace_buffer_access_if_rqd(const char *function_name, sycl::handler &cgh) {
+        external_workspace_helper_.get_workspace_buffer_access_if_rqd(function_name, cgh);
+    }
+
+protected:
+    friend class external_workspace_helper<prec, dom>;
+
+    // This must be reimplemented for backends that support external workspaces.
+    virtual std::int64_t get_workspace_external_bytes_impl() {
+        return 0;
+    };
 };
 
 } // namespace oneapi::mkl::dft::detail
--- a/include/oneapi/mkl/dft/detail/commit_impl.hpp
+++ b/include/oneapi/mkl/dft/detail/commit_impl.hpp
@@ -81,11 +81,11 @@ public:
         return external_workspace_helper_.get_rqd_workspace_bytes(*this);
     };
 
-    // set_workspace should be overrided for any backend that enables external workspaces.
-    // If these are overrided, get_workspace_external_bytes_impl must also be overrided.
-    // For backends that do not support external workspaces, these functions do not need to be overrided.
-    // When not overrided, external workspace support is faked: an external workspace can be set, 
-    // and errors will be generated according to the specificiation, 
+    // set_workspace should be overridden for any backend that enables external workspaces.
+    // If these are overridden, get_workspace_external_bytes_impl must also be overridden.
+    // For backends that do not support external workspaces, these functions do not need to be overridden.
+    // When not overridden, external workspace support is faked: an external workspace can be set,
+    // and errors will be generated according to the specificiation,
     // but the required workspace size will always be zero, and any given workspace will not actually be used.
     virtual void set_workspace(scalar_type *usm_workspace) {
         external_workspace_helper_.set_workspace_throw(*this, usm_workspace);
--- a/include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
+++ b/include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
@@ -63,24 +63,16 @@ private:
 public:
     /** Constructor.
      *  @param ext_workspace_rqd True if WORKSPACE_PLACEMENT is set to WORKSPACE_EXTERNAL.
-     *  @param backend_rqd_bytes_fn The function to call to obtain the backend's required workspace size.
     */
-    inline constexpr external_workspace_helper(bool ext_workspace_rqd)
+    constexpr external_workspace_helper(bool ext_workspace_rqd)
             : m_ext_workspace_rqd(ext_workspace_rqd),
               m_workspace_type(ext_workspace_type::not_set),
               m_workspace_bytes_rqd(-1) {}
 
-    /** Reset. May be required for changing descriptor settings before recommitting.
-     *  @param ext_workspace_rqd True if WORKSPACE_PLACEMENT is set to WORKSPACE_EXTERNAL.
-    */
-    inline void reset(bool ext_workspace_rqd) {
-        *this = external_workspace_helper(ext_workspace_rqd);
-    }
-
     /** Get the required workspace bytes for the backend's external workspace.
      *  @param committed_desc The backend's native descriptor.
     */
-    inline std::int64_t get_rqd_workspace_bytes(commit_impl_t& committed_desc) {
+    std::int64_t get_rqd_workspace_bytes(commit_impl_t& committed_desc) {
         if (m_workspace_bytes_rqd == -1) {
             m_workspace_bytes_rqd = committed_desc.get_workspace_external_bytes_impl();
         }
@@ -91,11 +83,10 @@ public:
      *  @param committed_desc The backend's native descriptor.
      *  @param usm_workspace A USM allocation for the workspace. Assumed to be sufficeintly large.
     */
-    inline void set_workspace_throw(commit_impl_t& committed_desc, scalar_t* usm_workspace) {
+    void set_workspace_throw(commit_impl_t& committed_desc, scalar_t* usm_workspace) {
         if (get_rqd_workspace_bytes(committed_desc) > 0 && usm_workspace == nullptr) {
             throw mkl::invalid_argument("DFT", "set_workspace",
-                                        "Provided workspace is too small (nullptr)");
-            return;
+                                        "Backend expected a non-null workspace pointer.");
         }
         m_ext_workspace_rqd = true;
         m_workspace_type = ext_workspace_type::usm;
@@ -105,8 +96,8 @@ public:
      *  @param committed_desc The backend's native descriptor.
      *  @param buffer_workspace A buffer for the workspace
     */
-    inline void set_workspace_throw(commit_impl_t& committed_desc,
-                                    sycl::buffer<scalar_t>& buffer_workspace) {
+    void set_workspace_throw(commit_impl_t& committed_desc,
+                             sycl::buffer<scalar_t>& buffer_workspace) {
         if (static_cast<std::size_t>(get_rqd_workspace_bytes(committed_desc)) / sizeof(scalar_t) >
             buffer_workspace.size()) {
             throw mkl::invalid_argument("DFT", "set_workspace", "Provided workspace is too small");
@@ -123,7 +114,7 @@ public:
     }
 
     template <typename FirstArgT, typename... ArgTs>
-    inline void compute_call_throw(const char* function_name) const {
+    void compute_call_throw(const char* function_name) const {
         constexpr bool is_pointer = std::is_pointer_v<std::remove_reference_t<FirstArgT>>;
         if constexpr (is_pointer) {
             usm_compute_call_throw(function_name);
@@ -133,17 +124,18 @@ public:
         }
     }
 
-    inline void get_workspace_buffer_access_if_rqd(const char* function_name, sycl::handler& cgh) {
+    void get_workspace_buffer_access_if_rqd(const char* function_name, sycl::handler& cgh) {
         if (m_ext_workspace_rqd) {
             if (m_workspace_buffer) {
                 if (m_workspace_buffer->size()) {
                     m_workspace_buffer->template get_access<sycl::access::mode::read_write>(cgh);
                 }
-                return;
             }
-            throw mkl::invalid_argument(
-                "DFT", function_name,
-                "Buffer external workspace must be used with buffer compute calls");
+            else {
+                throw mkl::invalid_argument(
+                    "DFT", function_name,
+                    "Buffer external workspace must be used with buffer compute calls");
+            }
         }
     }
 
@@ -151,7 +143,7 @@ private:
     /** When a compute function using USM arguments is called, throw an exception if an incorrect workspace has been set.
      *  @param function_name The name of the function to use in the error.
     */
-    inline void usm_compute_call_throw(const char* function_name) const {
+    void usm_compute_call_throw(const char* function_name) const {
         if (m_ext_workspace_rqd && m_workspace_type != ext_workspace_type::usm) {
             throw mkl::invalid_argument(
                 "DFT", function_name, "USM external workspace must be used with usm compute calls");
@@ -161,7 +153,7 @@ private:
     /** When a compute function using buffer arguments is called, throw an exception if an incorrect workspace has been set.
      *  @param function_name The name of the function to use in the error.
     */
-    inline void buffer_compute_call_throw(const char* function_name) const {
+    void buffer_compute_call_throw(const char* function_name) const {
         if (m_ext_workspace_rqd && m_workspace_type != ext_workspace_type::buffer) {
             throw mkl::invalid_argument(
                 "DFT", function_name,
--- a/include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
+++ b/include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
@@ -77,13 +77,6 @@ public:
         *this = external_workspace_helper(ext_workspace_rqd);
     }
 
-    /** Reset. May be required for changing descriptor settings before recommitting. Assumes no change
-     *  in workspace setting.
-    */
-    inline void reset() {
-        *this = external_workspace_helper(m_ext_workspace_rqd);
-    }
-
     /** Get the required workspace bytes for the backend's external workspace.
      *  @param committed_desc The backend's native descriptor.
     */
--- a/include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
+++ b/include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
@@ -1,0 +1,185 @@
+/*******************************************************************************
+* Copyright Codeplay Software Ltd
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing,
+* software distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions
+* and limitations under the License.
+*
+*
+* SPDX-License-Identifier: Apache-2.0
+*******************************************************************************/
+
+#ifndef _ONEMKL_DFT_EXTERNAL_WORKSPACE_HELPER_HPP_
+#define _ONEMKL_DFT_EXTERNAL_WORKSPACE_HELPER_HPP_
+
+#if __has_include(<sycl/sycl.hpp>)
+#include <sycl/sycl.hpp>
+#else
+#include <CL/sycl.hpp>
+#endif
+
+#include "oneapi/mkl/detail/export.hpp"
+#include "oneapi/mkl/dft/detail/types_impl.hpp"
+#include "oneapi/mkl/dft/detail/commit_impl.hpp"
+
+namespace oneapi {
+namespace mkl {
+namespace dft {
+namespace detail {
+
+template <precision prec, domain dom>
+class external_workspace_helper {
+public:
+    using commit_impl_t = commit_impl<prec, dom>;
+    using scalar_t = typename commit_impl_t::scalar_type;
+
+private:
+    // Enum to represent whatever the workspace was set as.
+    enum class ext_workspace_type {
+        not_set,
+        usm,
+        buffer,
+    };
+
+    // Is an external workspace required?
+    bool m_ext_workspace_rqd;
+
+    // Set workspace type, with optional workspaces.
+    ext_workspace_type m_workspace_type;
+
+    // Minimum size of workspace in bytes. -1 indicates not set.
+    std::int64_t m_workspace_bytes_rqd;
+
+    std::optional<sycl::buffer<scalar_t>> m_workspace_buffer;
+
+public:
+    /** Constructor.
+     *  @param ext_workspace_rqd True if WORKSPACE_PLACEMENT is set to WORKSPACE_EXTERNAL.
+     *  @param backend_rqd_bytes_fn The function to call to obtain the backend's required workspace size.
+    */
+    inline constexpr external_workspace_helper(bool ext_workspace_rqd)
+            : m_ext_workspace_rqd(ext_workspace_rqd),
+              m_workspace_type(ext_workspace_type::not_set),
+              m_workspace_bytes_rqd(-1) {}
+
+    /** Reset. May be required for changing descriptor settings before recommitting.
+     *  @param ext_workspace_rqd True if WORKSPACE_PLACEMENT is set to WORKSPACE_EXTERNAL.
+    */
+    inline void reset(bool ext_workspace_rqd) {
+        *this = external_workspace_helper(ext_workspace_rqd);
+    }
+
+    /** Reset. May be required for changing descriptor settings before recommitting. Assumes no change
+     *  in workspace setting.
+    */
+    inline void reset() {
+        *this = external_workspace_helper(m_ext_workspace_rqd);
+    }
+
+    /** Get the required workspace bytes for the backend's external workspace.
+     *  @param committed_desc The backend's native descriptor.
+    */
+    inline std::int64_t get_rqd_workspace_bytes(commit_impl_t& committed_desc) {
+        if (m_workspace_bytes_rqd == -1) {
+            m_workspace_bytes_rqd = committed_desc.get_workspace_external_bytes_impl();
+        }
+        return m_workspace_bytes_rqd;
+    }
+
+    /** Throw according to spec for setting the workspace. USM version.
+     *  @param committed_desc The backend's native descriptor.
+     *  @param usm_workspace A USM allocation for the workspace. Assumed to be sufficeintly large.
+    */
+    inline void set_workspace_throw(commit_impl_t& committed_desc, scalar_t* usm_workspace) {
+        if (get_rqd_workspace_bytes(committed_desc) > 0 && usm_workspace == nullptr) {
+            throw mkl::invalid_argument("DFT", "set_workspace",
+                                        "Provided workspace is too small (nullptr)");
+            return;
+        }
+        m_ext_workspace_rqd = true;
+        m_workspace_type = ext_workspace_type::usm;
+    }
+
+    /** Throw according to spec for setting the workspace. Buffer version.
+     *  @param committed_desc The backend's native descriptor.
+     *  @param buffer_workspace A buffer for the workspace
+    */
+    inline void set_workspace_throw(commit_impl_t& committed_desc,
+                                    sycl::buffer<scalar_t>& buffer_workspace) {
+        if (static_cast<std::size_t>(get_rqd_workspace_bytes(committed_desc)) / sizeof(scalar_t) >
+            buffer_workspace.size()) {
+            throw mkl::invalid_argument("DFT", "set_workspace", "Provided workspace is too small");
+            return;
+        }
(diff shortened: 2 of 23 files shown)


include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
```diff
@@ -0,0 +1,178 @@
...
+#endif
+
+#include "oneapi/mkl/detail/export.hpp"
+#include "oneapi/mkl/dft/detail/types_impl.hpp"
+#include "oneapi/mkl/dft/detail/commit_impl.hpp"
+
+namespace oneapi {
+namespace mkl {
+namespace dft {
+namespace detail {
```
FMarno, line 36:
```suggestion
namespace oneapi::mkl::dft::detail {
```
for me, this makes it clearer that there is only actually 1 workspace here so I don't need to worry about finding the brace where it ends.

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+    // Minimum size of workspace in bytes. -1 indicates not set.
+    std::int64_t m_workspace_bytes_rqd;
+
+    std::optional<sycl::buffer<scalar_t>> m_workspace_buffer;
+
+public:
+    /** Constructor.
+     *  @param ext_workspace_rqd True if WORKSPACE_PLACEMENT is set to WORKSPACE_EXTERNAL.
+     *  @param backend_rqd_bytes_fn The function to call to obtain the backend's required workspace size.
```
FMarno:
```suggestion

```

include/oneapi/mkl/dft/detail/commit_impl.hpp
```diff
@@ -67,6 +77,20 @@ class commit_impl {
 
     virtual void commit(const dft_values<prec, dom> &) = 0;
 
+    inline std::int64_t get_workspace_external_bytes() {
+        return external_workspace_helper_.get_rqd_workspace_bytes(*this);
+    };
+
+    // Set workspace by be overrided for any backend that actually uses the workspace.
```
Rbiessy: Not sure what you meant here, there is a typo I think?

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
```diff
@@ -0,0 +1,178 @@
...
+    */
+    inline constexpr external_workspace_helper(bool ext_workspace_rqd)
+            : m_ext_workspace_rqd(ext_workspace_rqd),
+              m_workspace_type(ext_workspace_type::not_set),
+              m_workspace_bytes_rqd(-1) {}
+
+    /** Reset. May be required for changing descriptor settings before recommitting.
+     *  @param ext_workspace_rqd True if WORKSPACE_PLACEMENT is set to WORKSPACE_EXTERNAL.
+    */
+    inline void reset(bool ext_workspace_rqd) {
```
FMarno: was `reset` required when you had the `backend_rqd_bytes_fn` param?

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
```diff
@@ -0,0 +1,178 @@
...
+    };
+
+    // Is an external workspace required?
+    bool m_ext_workspace_rqd;
+
+    // Set workspace type, with optional workspaces.
+    ext_workspace_type m_workspace_type;
+
+    // Minimum size of workspace in bytes. -1 indicates not set.
+    std::int64_t m_workspace_bytes_rqd;
```
FMarno, line 59: Why are we "caching" this?

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    /** Throw according to spec for setting the workspace. USM version.
+     *  @param committed_desc The backend's native descriptor.
+     *  @param usm_workspace A USM allocation for the workspace. Assumed to be sufficeintly large.
+    */
+    inline void set_workspace_throw(commit_impl_t& committed_desc, scalar_t* usm_workspace) {
+        if (get_rqd_workspace_bytes(committed_desc) > 0 && usm_workspace == nullptr) {
+            throw mkl::invalid_argument("DFT", "set_workspace",
+                                        "Provided workspace is too small (nullptr)");
```
FMarno: I think it would be better to clearly say that the function was expecting a non-null pointer.

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
```diff
@@ -0,0 +1,178 @@
...
+            return;
```
FMarno: surely the return is unreachable?

src/dft/backends/cufft/commit.cpp
```diff
@@ -44,15 +44,20 @@ namespace detail {
...
 private:
+    using scalar_type = typename dft::detail::commit_impl<prec, dom>::scalar_type;
+
     // For real to complex transforms, the "type" arg also encodes the direction (e.g. CUFFT_R2C vs CUFFT_C2R) in the plan so we must have one for each direction.
     // We also need this because oneMKL uses a directionless "FWD_DISTANCE" and "BWD_DISTANCE" while cuFFT uses a directional "idist" and "odist".
     // plans[0] is forward, plans[1] is backward
     std::array<std::optional<cufftHandle>, 2> plans = { std::nullopt, std::nullopt };
     std::array<std::int64_t, 2> offsets;
+    // For using a buffer external workspace, the buffer object itself needs to be kept between compute calls such that accessors can be used.
+    std::optional<sycl::buffer<scalar_type, 1>> buffer_workspace;
```
Rbiessy: Is this needed? I don't think it's being used since it is already part of `commit_impl` and `external_workspace_helper`

src/dft/backends/cufft/commit.cpp
```diff
@@ -323,6 +351,76 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
         return offsets;
     }
 
+    virtual void set_workspace(scalar_type* usmWorkspace) override {
```
Rbiessy: I think we should stick to snake_case for function arguments, same for `bufferWorkspace`.

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
```diff
@@ -0,0 +1,178 @@
...
+        }
+        m_ext_workspace_rqd = true;
+        m_workspace_type = ext_workspace_type::usm;
+    }
+
+    /** Throw according to spec for setting the workspace. Buffer version.
+     *  @param committed_desc The backend's native descriptor.
+     *  @param buffer_workspace A buffer for the workspace
+    */
+    inline void set_workspace_throw(commit_impl_t& committed_desc,
```
FMarno: Do you really need to mark the functions inline? This is a templated class so I wouldn't think ODR is an issue (could be wrong)

src/dft/backends/cufft/commit.cpp
```diff
@@ -323,6 +351,76 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+    std::int64_t get_plan0_workspace_size_bytes() {
+        if (!plans[0]) {
+            return 0;
+        }
+        return get_plan_workspace_size_bytes(*plans[0]);
+    }
+
+    virtual std::int64_t get_workspace_external_bytes_impl() override {
+        std::int64_t size = plans[1] ? get_plan_workspace_size_bytes(*plans[1]) : 0;
+        return size + get_plan0_workspace_size_bytes();
```
Rbiessy: I didn't think of the issue of multiple plans with the external workspace. I think we should return the maximum of the 2 workspace size. I'm concerned that this would artificially increase the workspace size for users that only use one direction.
This would add the requirement that we must wait between `compute` functions of a same descriptor that uses an external workspace which I think is acceptable.

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
```diff
@@ -0,0 +1,178 @@
...
+        constexpr bool is_pointer = std::is_pointer_v<std::remove_reference_t<FirstArgT>>;
+        if constexpr (is_pointer) {
+            usm_compute_call_throw(function_name);
+        }
+        else {
+            buffer_compute_call_throw(function_name);
+        }
+    }
+
+    inline void get_workspace_buffer_access_if_rqd(const char* function_name, sycl::handler& cgh) {
```
FMarno: What does this do?

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
```diff
@@ -0,0 +1,178 @@
...
+        if (m_ext_workspace_rqd) {
+            if (m_workspace_buffer) {
+                if (m_workspace_buffer->size()) {
+                    m_workspace_buffer->template get_access<sycl::access::mode::read_write>(cgh);
+                }
+                return;
+            }
+            throw mkl::invalid_argument(
+                "DFT", function_name,
+                "Buffer external workspace must be used with buffer compute calls");
```
FMarno:
```suggestion
            if (m_workspace_buffer) {
                if (m_workspace_buffer->size()) {
                    m_workspace_buffer->template get_access<sycl::access::mode::read_write>(cgh);
                }

            } else {
            throw mkl::invalid_argument(
                "DFT", function_name,
                "Buffer external workspace must be used with buffer compute calls");
                }
```

src/dft/backends/cufft/backward.cpp
```diff
@@ -69,6 +69,7 @@ ONEMKL_EXPORT void compute_backward(descriptor_type &desc,
 
     queue.submit([&](sycl::handler &cgh) {
         auto inout_acc = inout.template get_access<sycl::access::mode::read_write>(cgh);
+        commit->get_workspace_buffer_access_if_rqd("compute_backward", cgh);
```
FMarno: I presume this is here to show that the buffer is being used here so should be added to the dependency graph, but it makes me a little nervous that there is no variable with the scope of the whole function.

tests/unit_tests/dft/source/workspace_external_tests.cpp
```diff
@@ -0,0 +1,358 @@
...
+        forward_t rescale =
+            static_cast<forward_t>(1) / static_cast<forward_t>(static_cast<scalar_t>(dftSize));
+        sanityCheckPasses = parseval_check(dftSize, hostFwd.data(), hostBwd.data(), rescale);
+    }
+
+    sycl::free(deviceFwd, sycl_queue);
+    sycl::free(deviceBwd, sycl_queue);
+    sycl::free(workspace, sycl_queue);
+    return sanityCheckPasses ? !::testing::Test::HasFailure() : ::testing::Test::HasFailure();
+    ;
```
Rbiessy:
```suggestion
```

tests/unit_tests/dft/source/workspace_external_tests.cpp
```diff
@@ -0,0 +1,358 @@
...
+            compute_backward<decltype(desc), backward_t, forward_t>(desc, bufBwd, bufFwd);
+            sycl_queue.wait_and_throw();
+        }
+        forward_t rescale =
+            static_cast<forward_t>(1) / static_cast<forward_t>(static_cast<scalar_t>(dftSize));
+        sanityCheckPasses = parseval_check(dftSize, hostFwd.data(), hostBwdCpy.data(), rescale);
+    }
+
+    return sanityCheckPasses ? !::testing::Test::HasFailure() : ::testing::Test::HasFailure();
+    ;
```
Rbiessy:
```suggestion
```

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
```diff
@@ -0,0 +1,178 @@
...
+            m_workspace_bytes_rqd = committed_desc.get_workspace_external_bytes_impl();
+        }
+        return m_workspace_bytes_rqd;
+    }
+
+    /** Throw according to spec for setting the workspace. USM version.
+     *  @param committed_desc The backend's native descriptor.
+     *  @param usm_workspace A USM allocation for the workspace. Assumed to be sufficeintly large.
+    */
+    inline void set_workspace_throw(commit_impl_t& committed_desc, scalar_t* usm_workspace) {
```
FMarno: I don't really like the name `set_workspace_throw` since it doesn't set the workspace and hopefully doesn't throw.
I would prefer something like `check_workspace_is_valid` or just `check_workspace_valid` or maybe `check_workspace_size`.

src/dft/backends/cufft/commit.cpp
```diff
@@ -315,6 +332,17 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
         clean_plans();
     }
 
+    static void apply_external_workspace_setting(cufftHandle& handle,
+                                                 const config_value& workspace_setting) {
```
FMarno:
```suggestion
    static void apply_external_workspace_setting(cufftHandle handle,
                                                 config_value workspace_setting) {
```
cufftHandle is just an `unsigned int` so should be passed by value, and `config_value` is something like that too.

src/dft/backends/cufft/commit.cpp
```diff
@@ -323,6 +351,76 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+                    throw oneapi::mkl::exception(
+                        "dft/backends/cufft", "set_workspace",
+                        "cufftSetStream returned " + std::to_string(result));
+                }
+                auto workspace_native = reinterpret_cast<scalar_type*>(
+                    ih.get_native_mem<sycl::backend::ext_oneapi_cuda>(workspace_acc));
+                cufftSetWorkArea(plan, workspace_native);
+            });
+        });
+        this->get_queue().wait_and_throw();
```
FMarno:
```suggestion
        }).wait_and_throw();
```

src/dft/backends/cufft/commit.cpp
hjabird: I'd prefer not to add a user requirement about waits between compute calls. But I think we can avoid this:
* I think the buffer implementation already will already have implicit dependencies between compute forward / compute backwards, since they're getting access to the same underlying buffer. 
  * I kept it simple doing this. Anyone migrating from CUDA probably isn't using a buffer anyway.
* For USM we can always have a dependency on the last use of the allocation.
  * I don't *think* it can cause hidden deadlocks. Open to thoughts on that though.

Finally this might be good because in the implementation of this PR I introduced errors in both rocFFT and cuFFT when splitting up the buffers/USM allocs. 

Are you happy with the idea of implicit dependencies in the USM impl?

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: Sorry, I don't understand the comment.

When a descriptor is recommitted, if there is an external workspace, user needs to:
* Re-query the external workspace size requirements.
* set_workspace (even if it had previously been set).

`reset()` resets the internal state of the workspace helper to make sure oneMKL throws the appropriate errors.

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: Because:
* We need to provide this to the user, and also internally when set_workspace is called.
* We don't know the cost of asking the backend for the workspace space requirement. 
* We only need to obtain this value once per commit, so caching it is dead simple.

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
FMarno: Sorry, I should have been clearer.
I was thinking users of the class could just do this action themselves, as in override their existing `external_workspace_helper` using the constructor and copy assignment operator.

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: The backend might not require an external workspace, in which case the user might legally provide a `nullptr`.

FMarno: I'm just commenting on the explanation string. "too small" seems odd to me.
```suggestion
                                        "Backend expected a non-null workspace");
```

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
FMarno: It may also be super cheap to query!
I was just think it might be unneeded complexity. Its very neatly contained so I'm not so bothered really.

src/dft/backends/cufft/backward.cpp
hjabird: Yes, this is about adding the buffer dependency to the dependency graph. I also get nervous about the interaction of native APIs and buffers.

My understanding is that an accessor is safe to access within the scope of the enclosing command group scope. I think that this means that the dependency on `cgh` gives it the correct lifetime for the dependecy graph.

src/dft/backends/cufft/commit.cpp
Rbiessy: Yes that makes sense. As you say it should work for buffers and we can add the dependency ourselves for USM.
I think we could also add a comment somewhere or make a note in the spec that using external workspace serializes the computations of a same descriptor.

src/dft/backends/cufft/commit.cpp
hjabird: Done in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

src/dft/backends/cufft/commit.cpp
hjabird: Done in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: Done in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: Removed `inline` in several places in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: Removed `return` in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: Done in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: Done in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: Removed in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

tests/unit_tests/dft/source/workspace_external_tests.cpp
hjabird: Done in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

tests/unit_tests/dft/source/workspace_external_tests.cpp
hjabird: Done in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

src/dft/backends/cufft/commit.cpp
hjabird: That would be consistent with the contribution guidelines for oneMKL. I've updated your suggestions here, and a few other places I missed in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

src/dft/backends/cufft/commit.cpp
hjabird: Its leftover from where my initial attempts didn't neatly fit into the existing code. Removed in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

include/oneapi/mkl/dft/detail/commit_impl.hpp
hjabird: I've reworded this slightly in [1da13ee](https://github.com/oneapi-src/oneMKL/pull/411/commits/1da13eef0fbb663fcadc1abbcc15ad930f977589).

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
FMarno: There are a number of other instances of this

tests/unit_tests/dft/include/parseval_check.hpp
```diff
@@ -0,0 +1,81 @@
...
+ * 
+ * @tparam TypeFwd Forward domain type
+ * @tparam TypeBwd Backward domain type
+ * @param dft_len DFT size
+ * @param in forward domain data
+ * @param out bwd domain data
+ * @param rescale_forward A value to multiply the in data by.
+*/
+template <typename TypeFwd, typename TypeBwd>
+bool parseval_check(std::size_t dft_len, const TypeFwd* in, TypeBwd* out,
```
FMarno, line 44:
```suggestion
bool parseval_check(std::size_t dft_len, const TypeFwd* in, const TypeBwd* out,
```

tests/unit_tests/dft/source/workspace_external_tests.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+template <oneapi::mkl::dft::precision prec, oneapi::mkl::dft::domain dom>
+int test_workspace_external_usm_impl(std::size_t dft_size, sycl::device* dev) {
+    using namespace oneapi::mkl::dft;
+    using scalar_t = std::conditional_t<prec == precision::DOUBLE, double, float>;
+    using forward_t = std::conditional_t<dom == domain::COMPLEX, std::complex<scalar_t>, scalar_t>;
+    using backward_t = std::complex<scalar_t>;
+
+    sycl::queue sycl_queue(*dev);
+    if (prec == precision::DOUBLE && !sycl_queue.get_device().has(sycl::aspect::fp64)) {
```
FMarno, line 46:
```suggestion
    if (prec == precision::DOUBLE && !sycl_queue.get_device().has(sycl::aspect::fp64)) {
                    std::cout << "Device does not support double precision." << std::endl;
```

tests/unit_tests/dft/source/workspace_external_tests.cpp
```diff
@@ -0,0 +1,357 @@
...
+    desc.get_value(config_param::WORKSPACE_EXTERNAL_BYTES, &workspace_bytes);
+    if (workspace_bytes < 0) {
+        return ::testing::Test::HasFailure();
+    }
+    sycl::buffer<scalar_t> workspace(static_cast<std::size_t>(workspace_bytes) / sizeof(scalar_t));
+    desc.set_workspace(workspace);
+    // Generate data
+    std::vector<forward_t> host_fwd(static_cast<std::size_t>(dft_size));
+    std::size_t bwd_size =
+        dom == domain::COMPLEX ? dft_size : dft_size / 2 + 1; // TODO: Check this!
```
FMarno, line 142: It seems correct to me!

src/dft/backends/cufft/commit.cpp
```diff
@@ -323,6 +350,77 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+    }
+
+    // Separate workspaces are needed for the forward and backward plans. Because the plans[1]'s workspace is offset
+    // by the size of plans[0]'s workspace, we need to be able to get plans[0]'s size individually.
+    std::int64_t get_plan0_workspace_size_bytes() {
+        if (!plans[0]) {
+            return 0;
+        }
+        return get_plan_workspace_size_bytes(*plans[0]);
+    }
```
FMarno:
```suggestion
    std::int64_t get_plan_workspace_size_bytes(std::size_t dir) {
        if (!plans[dir]) {
            return 0;
        }
        return get_plan_workspace_size_bytes(*plans[dir]);
    }
```
I don't think it's much to add a `0` to your above call and then this can be reused for `get_workspace_external_bytes_impl`

src/dft/backends/cufft/commit.cpp
```diff
@@ -323,6 +350,77 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+            set_buffer_workspace(*plans[0], buffer_workspace, plan0_count, 0);
+        }
+        if (plans[1]) {
+            set_buffer_workspace(*plans[1], buffer_workspace, plan1_count, plan0_count);
+        }
+    }
+
+    std::int64_t get_plan_workspace_size_bytes(cufftHandle handle) {
+        std::size_t size = 0;
+        cufftGetSize(*plans[0], &size);
```
FMarno, line 396:
```suggestion
        cufftGetSize(handle, &size);
```

src/dft/backends/cufft/commit.cpp
FMarno: I think the synchronisation should be offloaded to the user. In much the same way that the user shouldn't try to launch two kernels that use the same USM memory without the use of events to synchronize, then it should be clear from our specification that the external workspace is in use while the `compute_forward` kernel is in flight.
Kernels launched from the same descriptor should be independent.
Trying to sort this ourselves goes against USM I would say.

src/dft/backends/cufft/commit.cpp
```diff
@@ -323,6 +350,77 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+                    buffer_workspace.template get_access<sycl::access::mode::read_write>(cgh, range,
+                                                                                         offset);
+                cgh.host_task([=](sycl::interop_handle ih) {
+                    auto stream = ih.get_native_queue<sycl::backend::ext_oneapi_cuda>();
+                    auto result = cufftSetStream(plan, stream);
+                    if (result != CUFFT_SUCCESS) {
+                        throw oneapi::mkl::exception(
+                            "dft/backends/cufft", "set_workspace",
+                            "cufftSetStream returned " + std::to_string(result));
+                    }
```
FMarno, line 375: Why are you setting the stream? the stream is changed with each call to compute forward

include/oneapi/mkl/dft/detail/commit_impl.hpp
```diff
@@ -67,6 +77,20 @@ class commit_impl {
 
     virtual void commit(const dft_values<prec, dom> &) = 0;
 
+    inline std::int64_t get_workspace_external_bytes() {
+        return external_workspace_helper_.get_rqd_workspace_bytes(*this);
+    };
+
+    // set_workspace should be overrided for any backend that enables external workspaces.
+    // If these are overrided, get_workspace_external_bytes_impl must also be overrided.
+    // For backends that do not support external workspaces, these functions do not need to be overrided.
```
t4c1: this comment could be improved by explaining what these defaults do if not overridden.

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
```diff
@@ -0,0 +1,170 @@
...
+        else {
+            buffer_compute_call_throw(function_name);
+        }
+    }
+
+    void get_workspace_buffer_access_if_rqd(const char* function_name, sycl::handler& cgh) {
+        if (m_ext_workspace_rqd) {
+            if (m_workspace_buffer) {
+                if (m_workspace_buffer->size()) {
+                    m_workspace_buffer->template get_access<sycl::access::mode::read_write>(cgh);
```
t4c1, line 136: What is the point of creating accessor if it is not used anywhere?

src/dft/backends/cufft/commit.cpp
```diff
@@ -323,6 +350,77 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
     }
 
+    virtual void set_workspace(scalar_type* usm_workspace) override {
+        this->external_workspace_helper_.set_workspace_throw(*this, usm_workspace);
+        if (plans[0]) {
+            cufftSetWorkArea(*plans[0], usm_workspace);
+        }
+        if (plans[1]) {
+            cufftSetWorkArea(
+                *plans[1], usm_workspace + get_plan0_workspace_size_bytes() / sizeof(scalar_type));
```
t4c1: Do we need separate workspaces for both plans? Does the spec allow forward and backward compute of a descriptor running concurrently?

src/dft/backends/cufft/commit.cpp
```diff
@@ -323,6 +350,77 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+                    auto stream = ih.get_native_queue<sycl::backend::ext_oneapi_cuda>();
+                    auto result = cufftSetStream(plan, stream);
+                    if (result != CUFFT_SUCCESS) {
+                        throw oneapi::mkl::exception(
+                            "dft/backends/cufft", "set_workspace",
+                            "cufftSetStream returned " + std::to_string(result));
+                    }
+                    auto workspace_native = reinterpret_cast<scalar_type*>(
+                        ih.get_native_mem<sycl::backend::ext_oneapi_cuda>(workspace_acc));
+                    cufftSetWorkArea(plan, workspace_native);
```
t4c1, line 378: I do not think this is OK. SYCL runtime only expects the memory to be used within the host_task call. This could lead to multiple subtle bugs. So accessor should be constructed within the same host_task call that launches the cufft kernel.

src/dft/backends/cufft/commit.cpp
Rbiessy: There are a few more occurrences of `bufferWorkspace` and `usmWorkspace` in other files.

src/dft/backends/cufft/commit.cpp
hjabird: I've implemented this with implicit USM dependencies in [9c2b2ff](https://github.com/oneapi-src/oneMKL/pull/411/commits/9c2b2ffc008faf9b5da5a137fa53e2e74f935785) and [19869bf](https://github.com/oneapi-src/oneMKL/pull/411/commits/19869bf06518db4983183a24ca7f68f907f1c834).

I agree that this doesn't fit with the typical usage of USM. However, anyone porting from closed-source *oneMKL* and expecting their code to work with *cuFFT* and *rocFFT* would expect their code to just work without additional changes to dependency tracking.

I'm also under the impression that CUDA streams are in-order (perhaps I'm wrong here - I've never used CUDA directly). Consequently, kernels launched from the same sycl queue would not run concurrently anyway. *cuFFT* launches multiple kernels per DFT, so the USM dependency would enforce that kernels from different DFTs could not be interleaved. In this context, the implicit dependency for the USM implementation does not seem so bad.

FMarno: I'm having trouble finding the docs about closed-source oneMKL synchronizing on the use of the workspace, do you have a link?

CUDA streams are in-order, but we change the stream for each call to `compute_xxxward` to allow transforms to happen concurrently.

I think launching N kernels from the same descriptor is similar to launching 1 kernel from each of N descriptors that share the same external workspace. If you had limited memory then you could save a lot by sharing the same workspace between all the descriptors.

hjabird: The reference for closed-source oneMKL is [here](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2023-0/user-allocated-workspaces.html). It states that:
> The descriptor assumes that the workspace will be solely used for FFT, so the user must be careful not to use the same workspace area for other purposes in parallel or before finishing compute_forward/compute_backward (especially for USM as explicit synchronization is required).

My interpretation of this is that for anything other than FFTs, explicit synchronisation is required. Therefore we don't have to worry about synchronisation for multiple DFTs with a single descriptor. Perhaps @raphael-egan can comment?

Regarding CUDA streams, my understanding of how streams map to CUDA was wrong. Thanks for clearing that up.

src/dft/backends/cufft/commit.cpp
hjabird: Now that a workspace is shared for forwards and backwards in the *cuFFT* and *rocFFT* backends, this function no longer exists.

src/dft/backends/cufft/commit.cpp
hjabird: Hopefully all the stragglers caught in [fd528ea](https://github.com/oneapi-src/oneMKL/pull/411/commits/fd528eae5034fec9ca50f349c74f8c7b3925f443).

src/dft/backends/cufft/commit.cpp
FMarno: Thanks for the link!
To me, that reads only as a reminder that the descriptor cannot detect if the memory is being used elsewhere. I don't see a guarantee of synchronization between calls to compute_xxxward.

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: The *cuFFT* and *rocFFT* backends take their workspace arguments as pointers. Consequently, implementing the buffer API with then poses a few issues. 

The issue that this aims to resolve is that of the SYCL runtime knowing that a workspace buffer is used when a DFT accesses it via a captured pointer (captured in `set_workspace`). Here, we create an accessor out of the buffer (to create the expected kernel dependencies) but then access the buffer via the captured pointer instead of via the accessor itself.

src/dft/backends/cufft/commit.cpp
hjabird: This is discussed in https://github.com/oneapi-src/oneMKL/pull/411#discussion_r1399082325. The implementation has been changed to use a single shared workspace, but the discussion is ongoing, so please add your thoughts there!

src/dft/backends/cufft/commit.cpp
raphael-egan: > The reference for closed-source oneMKL is [here](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2023-0/user-allocated-workspaces.html). It states that:
> 
> > The descriptor assumes that the workspace will be solely used for FFT, so the user must be careful not to use the same workspace area for other purposes in parallel or before finishing compute_forward/compute_backward (especially for USM as explicit synchronization is required).
> 
> My interpretation of this is that for anything other than FFTs, explicit synchronisation is required. Therefore we don't have to worry about synchronisation for multiple DFTs with a single descriptor. Perhaps @raphael-egan can comment?
> 
> Regarding CUDA streams, my understanding of how streams map to CUDA was wrong. Thanks for clearing that up.

Answer given by email and shared with Cheetah team.

src/dft/backends/cufft/commit.cpp
hjabird: You probably correct from the standpoint of the spec (although the interop spec for CUDA or AMD are not yet published). My understanding is that this should work with Codeplay's oneAPI for Nvidia GPUs and oneAPI for AMD GPUs.

include/oneapi/mkl/dft/detail/commit_impl.hpp
hjabird: I've expanded the comment in [d31868a](https://github.com/oneapi-src/oneMKL/pull/411/commits/d31868aefc19b5b7dfc15f661f62cd54b122a72c).

src/dft/backends/cufft/commit.cpp
hjabird: I was able to do this more tidily than expected in [fe3273f](https://github.com/oneapi-src/oneMKL/pull/411/commits/fe3273f53bb6b3a0449e94b65bba4d77877a62ee).

src/dft/backends/cufft/commit.cpp
hjabird: No longer calling setStream in [bee7d30](https://github.com/oneapi-src/oneMKL/pull/411/commits/bee7d3068a28fa90a868f23baab198dffad3440b).

src/dft/backends/cufft/commit.cpp
hjabird: Done in [bee7d30](https://github.com/oneapi-src/oneMKL/pull/411/commits/bee7d3068a28fa90a868f23baab198dffad3440b)

tests/unit_tests/dft/source/workspace_external_tests.cpp
hjabird: TODO removed in [bee7d30](https://github.com/oneapi-src/oneMKL/pull/411/commits/bee7d3068a28fa90a868f23baab198dffad3440b).

tests/unit_tests/dft/source/workspace_external_tests.cpp
hjabird: Done in [bee7d30](https://github.com/oneapi-src/oneMKL/pull/411/commits/bee7d3068a28fa90a868f23baab198dffad3440b).

tests/unit_tests/dft/include/parseval_check.hpp
hjabird: Done in [bee7d30](https://github.com/oneapi-src/oneMKL/pull/411/commits/bee7d3068a28fa90a868f23baab198dffad3440b).

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: I agree its not a great name! It sets the workspace information within the external_workspace_helper, and should throw anything required by the specification.

I've renamed this `set_workspace_metadata` since this isn't just a check that its valid, and as you say, ideally doesn't throw. Let me know if you think this is good enough. In [bee7d30](https://github.com/oneapi-src/oneMKL/pull/411/commits/bee7d3068a28fa90a868f23baab198dffad3440b).

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: Now renamed `get_workspace_accessor_if_rqd`, and also returning `std::optional<sycl::accessor>`.

For backends that don't natively support buffers (eg. anything that isn't SYCL), the workspace buffer dependency of a compute function must be added the SYCL dependency graph when the call is submitted to the queue. This is achieved by creating an accessor from te workspace buffer. This function does that if its needed (ie. a buffer workspace is needed, and its size is greater than zero), and also now returns this workspace accessor.

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: I've removed more instances in [bee7d30](https://github.com/oneapi-src/oneMKL/pull/411/commits/bee7d3068a28fa90a868f23baab198dffad3440b).

include/oneapi/mkl/dft/detail/external_workspace_helper.hpp
hjabird: Done in [bee7d30](https://github.com/oneapi-src/oneMKL/pull/411/commits/bee7d3068a28fa90a868f23baab198dffad3440b).

src/dft/backends/cufft/commit.cpp
hjabird: It looks as if a number of backends don't make any gaurantees regarding calling multiple *compute_XXX* functions. For example, I believe that calling multiple *cuFFT* compute functions without interdependencies using the same descriptor could lead to problems in the current oneMKL implementation since the SYCL runtime will schedule those without explicit dependencies to run concurrently by changing streams - see [how cuFFT doesn't suppor this on Stackoverflow](https://stackoverflow.com/questions/39030949/cufftsetstream-causes-garbage-output-am-i-doing-something-wrong).

The problem of not scheduling concurrent `compute_*` calls therefore has a larger scope than just for explict workspace implementation, and I believe this should be addressed at a oneAPI spec level.

I'll be removing the interdepencies from the USM version of the code. For SYCL spec conformance, using accessors on workspace buffers is still the correct thing to do, so these will still implicitly result in implicit synchronisation.

hjabird: Removed dependencies for USM compute calls in [9856b7d](https://github.com/oneapi-src/oneMKL/pull/411/commits/9856b7d5d320c5807e7a17cefad32b3a5074af9f).

Rbiessy: That looks good to me so I'll resolve this big discussion, thanks.

Rbiessy (approved):

tests/unit_tests/dft/source/workspace_external_tests.cpp
FMarno: I don't see the change

hjabird: Well the change was in my editor... unsaved. Oops. Done in [aa0967a](https://github.com/oneapi-src/oneMKL/pull/411/commits/aa0967a974e68d6701c408969b43132fd50be684).

src/dft/backends/cufft/commit.cpp
```diff
@@ -323,6 +350,63 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+        this->external_workspace_helper_.set_workspace_metadata(*this, usm_workspace);
+        if (plans[0]) {
+            cufftSetWorkArea(*plans[0], usm_workspace);
+        }
+        if (plans[1]) {
+            cufftSetWorkArea(*plans[1], usm_workspace);
+        }
+    }
+
+    void set_buffer_workspace(cufftHandle plan, sycl::buffer<scalar_type>& buffer_workspace) {
```
FMarno, line 363: Why do we need this function if we are going to set the workspace with each compute call using buffers?

hjabird: I give my reasoning in the comment starting on line 364 - to allow internal memory to be freed (if required), and to allow cuFFT to generate any errors (I think its preferable to get errors for setting the workspace at this point in comparison to later when compute calls are made). I'm not sure what errors are likely. I'd guess the most likely is that the buffer can't be accessed on the Nvidia device.

FMarno: Sorry, yes you did write that.

FMarno (approved):

t4c1 (approved):

hjabird:
I've been having trouble with access to AMD hardware since the rebase, so test here is test data for MKL-CPU, MKL-GPU and Nvidia for now:
[testintel.txt](https://github.com/oneapi-src/oneMKL/files/13582967/testintel.txt)
[testnvidia.txt](https://github.com/oneapi-src/oneMKL/files/13582968/testnvidia.txt)

hjabird:
This PR has been rebased:
* DFT domain has had portFFT added as a backend
* Additional work was required to support portFFT in the tests

Test results for Nvidia (cuFFT backend), Intel MKL-CPU and MKL-GPU, and portFFT on Intel are attached:
[testintel_mkl.txt](https://github.com/oneapi-src/oneMKL/files/13679187/testintel_mkl.txt)
[testintel_portfft.txt](https://github.com/oneapi-src/oneMKL/files/13679188/testintel_portfft.txt)
[testnvidia.txt](https://github.com/oneapi-src/oneMKL/files/13679189/testnvidia.txt)

Unfortunately I cannot currently access AMD hardware to test that the rocFFT backend. However, it has not required any changes since the last test results were provided.

include/oneapi/mkl/dft/detail/commit_impl.hpp
```diff
@@ -67,6 +77,23 @@ class commit_impl {
 
     virtual void commit(const dft_values<prec, dom> &) = 0;
 
+    inline std::int64_t get_workspace_external_bytes() {
+        return external_workspace_helper_.get_rqd_workspace_bytes(*this);
+    };
+
+    // set_workspace should be overrided for any backend that enables external workspaces.
+    // If these are overrided, get_workspace_external_bytes_impl must also be overrided.
```
anantsrivastava30: overridden perhaps instead of overrided.

anantsrivastava30 (approved):
LGTM!

tests/unit_tests/main_test.cpp
```diff
@@ -102,9 +102,9 @@ int main(int argc, char** argv) {
             for (auto dev : plat_devs) {
                 try {
                     /* Do not test for OpenCL backend on GPU */
-                    if (dev.is_gpu() && plat.get_info<sycl::info::platform::name>().find(
-                                            "OpenCL") != std::string::npos)
-                        continue;
+                    // if (dev.is_gpu() && plat.get_info<sycl::info::platform::name>().find(
+                    //                         "OpenCL") != std::string::npos)
+                    //     continue;
```
Rbiessy: I think you didn't mean to push that.

tests/unit_tests/dft/source/workspace_external_tests.cpp
```diff
@@ -0,0 +1,402 @@
...
+    sycl::queue sycl_queue(*GetParam());
+    const int dft_len = 1024 * 3 * 5 * 7 * 16; // A size likely to require an external workspace.
+    float* fft_data_usm = sycl::malloc_device<float>(dft_len * 2, sycl_queue);
+    descriptor<precision::SINGLE, domain::COMPLEX> desc_usm(dft_len);
+    try {
+        // WORKSPACE_EXTERNAL is NOT set.
+        commit_descriptor(desc_usm, sycl_queue);
+    }
+    catch (oneapi::mkl::unimplemented&) {
+        // Admit defeat, and just test that the API does as expected.
```
Rbiessy: Is there a reason this comment is different from the ones above in a similar situation?
```suggestion
        // DFT size may not be supported. Use a DFT size that probably will be, even if it
        // won't actually use an external workspace internally.
```

tests/unit_tests/main_test.cpp
hjabird: I did not - thank you for spotting this. Reverted in [d0d10fe](https://github.com/oneapi-src/oneMKL/pull/411/commits/d0d10febbaf898e36bdd05f8579ef1a5338e0020).

include/oneapi/mkl/dft/detail/commit_impl.hpp
hjabird: Yes, my English was incorrect there. Corrected in [2c4c096](https://github.com/oneapi-src/oneMKL/pull/411/commits/2c4c0962030d285710c0faf485be289711ff36bf).

tests/unit_tests/dft/source/workspace_external_tests.cpp
hjabird: Applied in [985ea94](https://github.com/oneapi-src/oneMKL/pull/411/commits/985ea9406ed8cbb204e3bdff3b587df010374f1f).

Rbiessy (approved):

include/oneapi/mkl/dft/detail/commit_impl.hpp
hjabird: @anantsrivastava30 if you're happy with this (and everything else), it looks like I'm good to merge this PR!

Issue #306, opened 2023-04-21
Improve CMake to make oneMKL easier to use


Rbiessy:
# Summary
I have faced issues when trying to use OneMKL as a dependency using CMake. From my testing currently users of oneMKL need to manually set the right includes or libraries to link in CMake to have oneMKL as a dependency. I would like to discuss if we can improve this.

# Version
Latest oneMKL commit, AFAIK this is not a recent issue.

# Environment
I am using the oneAPI 2023.1.0 package and using the [modules files](https://www.intel.com/content/www/us/en/docs/oneapi/programming-guide/2023-1/use-modulefiles-with-linux.html) to load `mkl` and `compiler`.
CMake 3.26.3.

# Steps to reproduce

## First approach
I have tried to use OneMKL in 2 ways. First using `FetchContent` as shown with the following `CMakeLists.txt`:
```
cmake_minimum_required(VERSION 3.13)
project(sample)

include(FetchContent)
set(BUILD_FUNCTIONAL_TESTS OFF)
set(BUILD_EXAMPLES OFF)
FetchContent_Declare(
    open_onemkl
    GIT_REPOSITORY https://github.com/oneapi-src/oneMKL.git
    GIT_TAG dc22279396def65b1af1742a45dda7afcf7c0a22
)
FetchContent_MakeAvailable(open_onemkl)

add_executable(test test.cpp)
target_link_libraries(test PRIVATE onemkl)
```
and the simplest `test.cpp`:
```
int main() { return 0; }
```
The CMake command `cmake -Bbuild .` is failing with:
```
CMake Error at build/_deps/open_onemkl-src/src/blas/CMakeLists.txt:43 (target_link_libraries):
  Target "onemkl_blas" links to:

    ONEMKL::SYCL::SYCL

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```
and the same errors for all the other domains and backends. I think `FetchContent` is quite common in modern CMake. I have had more success with `ExternalProject` but this method has limitations and requires the user to manually set the include directory and link to the libraries as far as I know. Do you think it would be valuable to support this? I'm not sure what changes are needed to fix this so help would be appreciated.

## Second approach
Another common approach is to use `find_package(oneMKL)` like so:
```
cmake_minimum_required(VERSION 3.13)
project(sample)

find_package(oneMKL REQUIRED)
[...]

PR #413, opened 2023-11-22
Add one more Include to have the product workable with FetchContent

akabalov:
The issue with FetchContent was opened here https://github.com/oneapi-src/oneMKL/issues/306

The problem is that when we use FetchContent_Declare, it creates ${CMAKE_BINARY_DIR}/_deps dir where we have the build folder, downloaded sources, etc. 

We do have generation of or config.hpp file as a part of oneMKL cmake and it puts this config file into ${CMAKE_CURRENT_BINARY_DIR}/oneapi/mkl/config.hpp folder

So if we don't add paths to the ${CMAKE_CURRENT_BINARY_DIR} into the includes, we end up with failures from domains that this config.hpp is not found. 

To fix the problem, we just need to add ${CMAKE_CURRENT_BINARY_DIR} into includes.


mkrainiuk (approved):

Rbiessy:
Thanks for the patch. I can see you are only adding the include path for the mklcpu and mklgpu backends, don't we need this for all the other backends too?
Also just a suggestion I think it would be more readable to store the path to include in `src/CMakeLists.txt` as an internal variable `ONEMKL_GENERATED_INCLUDE_PATH` for instance and use it everywhere instead of relying on relative paths.

akabalov:
> Thanks for the patch. I can see you are only adding the include path for the mklcpu and mklgpu backends, don't we need this for all the other backends too? Also just a suggestion I think it would be more readable to store the path to include in `src/CMakeLists.txt` as an internal variable `ONEMKL_GENERATED_INCLUDE_PATH` for instance and use it everywhere instead of relying on relative paths.

Thank you for pointing me to the missed backends! I love your ideas! Does it look better now?

Rbiessy (approved):
Thanks, LGTM

src/rng/backends/rocrand/CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ add_library(${LIB_OBJ} OBJECT ${SOURCES})
 
 target_include_directories(
   ${LIB_OBJ} PRIVATE ${PROJECT_SOURCE_DIR}/include ${PROJECT_SOURCE_DIR}/src
-                     ${CMAKE_BINARY_DIR}/bin ${MKL_INCLUDE})
+                     ${CMAKE_BINARY_DIR}/bin ${MKL_INCLUDE} ${ONEMKL_GENERATED_INCLUDE_PATH})
```
mkrainiuk, line 67: Could you please check that clang-format won't complain about this line?

akabalov: Checked Clang format using our regular CI job. Everything is fine. Merging changes now.

PR #425, opened 2023-12-07
[CUDA][HIP] Use device to get native context

hdelan:
Since https://github.com/oneapi-src/unified-runtime/pull/999 it is no longer valid to get the native context from the SYCL context on a multi GPU system. The get native func for contexts has been deprecated for this reason. See https://github.com/intel/llvm/pull/10975

Similar ticket: https://github.com/oneapi-src/oneDNN/pull/1765

### DIFF
--- a/src/blas/backends/cublas/cublas_scope_handle.cpp
+++ b/src/blas/backends/cublas/cublas_scope_handle.cpp
@@ -42,10 +42,11 @@ CublasScopedContextHandler::CublasScopedContextHandler(sycl::queue queue, sycl::
         : ih(ih),
           needToRecover_(false) {
     placedContext_ = new sycl::context(queue.get_context());
-    auto device = queue.get_device();
-    auto desired = sycl::get_native<sycl::backend::ext_oneapi_cuda>(*placedContext_);
+    auto cudaDevice = ih.get_native_device<sycl::backend::ext_oneapi_cuda>();
     CUresult err;
+    CUcontext desired;
     CUDA_ERROR_FUNC(cuCtxGetCurrent, err, &original_);
+    CUDA_ERROR_FUNC(cuDevicePrimaryCtxRetain, err, &desired, cudaDevice);
     if (original_ != desired) {
         // Sets the desired context as the active one for the thread
         CUDA_ERROR_FUNC(cuCtxSetCurrent, err, desired);
@@ -87,8 +88,11 @@ void ContextCallback(void *userData) {
 }
 
 cublasHandle_t CublasScopedContextHandler::get_handle(const sycl::queue &queue) {
-    auto piPlacedContext_ = reinterpret_cast<pi_context>(
-        sycl::get_native<sycl::backend::ext_oneapi_cuda>(*placedContext_));
+    auto cudaDevice = ih.get_native_device<sycl::backend::ext_oneapi_cuda>();
+    CUresult cuErr;
+    CUcontext desired;
+    CUDA_ERROR_FUNC(cuDevicePrimaryCtxRetain, cuErr, &desired, cudaDevice);
+    auto piPlacedContext_ = reinterpret_cast<pi_context>(desired);
     CUstream streamId = get_stream(queue);
     cublasStatus_t err;
     auto it = handle_helper.cublas_handle_mapper_.find(piPlacedContext_);
--- a/src/dft/backends/cufft/commit.cpp
+++ b/src/dft/backends/cufft/commit.cpp
@@ -81,9 +81,10 @@ public:
         }
         if (fix_context) {
             // cufftDestroy changes the context so change it back.
-            CUcontext interopContext =
-                sycl::get_native<sycl::backend::ext_oneapi_cuda>(this->get_queue().get_context());
-            if (cuCtxSetCurrent(interopContext) != CUDA_SUCCESS) {
+            CUdevice interopDevice =
+                sycl::get_native<sycl::backend::ext_oneapi_cuda>(this->get_queue().get_device());
+            CUcontext interopContext;
+            if (cuDevicePrimaryCtxRetain(&interopContext, interopDevice) != CUDA_SUCCESS) {
                 throw mkl::exception("dft/backends/cufft", __FUNCTION__,
                                      "Failed to change cuda context.");
             }
--- a/src/blas/backends/rocblas/rocblas_scope_handle.cpp
+++ b/src/blas/backends/rocblas/rocblas_scope_handle.cpp
@@ -58,10 +58,11 @@ RocblasScopedContextHandler::RocblasScopedContextHandler(sycl::queue queue,
         : interop_h(ih),
           needToRecover_(false) {
     placedContext_ = new sycl::context(queue.get_context());
-    auto device = queue.get_device();
-    auto desired = sycl::get_native<sycl::backend::ext_oneapi_hip>(*placedContext_);
+    auto hipDevice = ih.get_native_device<sycl::backend::ext_oneapi_hip>();
     hipError_t err;
+    hipCtx_t desired;
     HIP_ERROR_FUNC(hipCtxGetCurrent, err, &original_);
+    HIP_ERROR_FUNC(hipDevicePrimaryCtxRetain, err, &desired, hipDevice);
     if (original_ != desired) {
         // Sets the desired context as the active one for the thread
         HIP_ERROR_FUNC(hipCtxSetCurrent, err, desired);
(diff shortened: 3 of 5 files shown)


jinz2014:
reading your changes, I have a question.

For example,

auto cudaDevice = sycl::get_native<sycl::backend::ext_oneapi_cuda>(queue.get_device());

Is the type of cudaDevice "CUdevice" ?

hdelan:
Hi @jinz2014 yes you are correct!

src/dft/backends/cufft/commit.cpp
```diff
@@ -78,8 +78,13 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
-            CUcontext interopContext =
-                sycl::get_native<sycl::backend::ext_oneapi_cuda>(this->get_queue().get_context());
+            CUdevice interopDevice =
+                sycl::get_native<sycl::backend::ext_oneapi_cuda>(this->get_queue().get_device());
+            CUcontext interopContext;
+            if (cuDevicePrimaryCtxRetain(&interopContext, interopDevice) != CUDA_SUCCESS) {
+                throw mkl::exception("dft/backends/cufft", __FUNCTION__,
+                                     "Failed to change cuda context.");
+            }
             if (cuCtxSetCurrent(interopContext) != CUDA_SUCCESS) {
```
FMarno: is `cuCtxSetCurrent` still required after `cuDevicePrimaryCtxRetain`? I'm not sure about the difference between retaining a context and setting it.
The reason for `cuCtxSetCurrent` was that we found that calling `cufftDestroy` would change the context for some reason and then some data accesses became invalid.

hdelan: We still need to set the current context. Retaining the primary context is just getting it, not setting it. It is analogous to `cuCtxCreate` in that it gets a (primary or not) context that can then be set as current

FMarno:
[cufft_run.txt](https://github.com/oneapi-src/oneMKL/files/13638282/cufft_run.txt)
All the DFT changes look good to me and I've run the DFT tests successfully.
I'd like to see test logs for the other backends before I approve.

src/dft/backends/cufft/commit.cpp
hdelan: I was actually wrong about this. I revisited the docs and `cuDevicePrimaryContextRetain` does indeed set the context. So I have removed the following `cuCtxSetCurrent`

hdelan:
AMD tests for lapack and blas all passing: 
[test_amd.txt](https://github.com/oneapi-src/oneMKL/files/13754532/test_amd.txt)

8 lapack nvidia test failing on GTX1050 but these tests are also failing on `develop` branch:
[test_cuda_lapack.txt](https://github.com/oneapi-src/oneMKL/files/13754558/test_cuda_lapack.txt)

Nvidia blas tests passing 
[test_cuda_blas.txt](https://github.com/oneapi-src/oneMKL/files/13754554/test_cuda_blas.txt)

src/dft/backends/cufft/commit.cpp
hdelan: I was wrong about being wrong. That is I was right! The CUDA docs have ambiguous wording around this, see here https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__PRIMARY__CTX.html#group__CUDA__PRIMARY__CTX_1g9051f2d5c31501997a6cb0530290a300 They say that the context becomes

> active and available to the user until the user releases it

But this does not mean that it is set as the current context! I have verified this behaviour in my own testing. It is strange that tests still passed without the `cuCtxSetCurrent`.

muhammad-tanvir-1211:
I see all the buffer tests failing for the `rocblas` backend with `PI_ERROR_INVALID_OPERATION`.

Logs:
[PR_425.txt](https://github.com/oneapi-src/oneMKL/files/13918415/PR_425.txt)

The failures are not because of the changes in this PR, but rather a recent change in the compiler. All these tests are expected to pass once https://github.com/oneapi-src/unified-runtime/pull/1226 and https://github.com/intel/llvm/pull/12297 are merged.

src/blas/backends/cublas/cublas_scope_handle.cpp
```diff
@@ -42,13 +42,13 @@ CublasScopedContextHandler::CublasScopedContextHandler(sycl::queue queue, sycl::
...
           needToRecover_(false) {
     placedContext_ = new sycl::context(queue.get_context());
-    auto device = queue.get_device();
-    auto desired = sycl::get_native<sycl::backend::ext_oneapi_cuda>(*placedContext_);
+    auto cudaDevice = ih.get_native_device<sycl::backend::ext_oneapi_cuda>();
     CUresult err;
+    CUcontext desired;
     CUDA_ERROR_FUNC(cuCtxGetCurrent, err, &original_);
+    // Getting the primary context also sets it as the active context
+    CUDA_ERROR_FUNC(cuDevicePrimaryCtxRetain, err, &desired, cudaDevice);
```
Rbiessy: Should we expect a performance cost from this change? From what I understand `cuCtxSetCurrent` was expected to be called only once before, assuming the context active was not changed outside of oneMKL.
This constructor is called once before each calls to blas functions so I am wary that the cost may add up.

Rbiessy (commented):
Would you be able to attach test logs again? Also does this change compile with the 2024.0 icpx release?

hjabird (commented):
Generally LGTM. I've tested the rocFFT and cuFFT backends with DPC++ 2024.0's icpx.
It would be good to see logs again after the rebase as Rbiessy suggests.

src/blas/backends/cublas/cublas_scope_handle.cpp
hdelan: The cost of `cuDevicePrimaryContextRetain` is minimal once the primary context is not being initialized for the first time, which it should not be here. A simple benchmark like this:
```c++
for (int i = 0; i < NUM_ITERATIONS; i++) {
    CHECK(cuDevicePrimaryCtxRetain(&context, device));
    CHECK(cuDevicePrimaryCtxRelease(device));
}
```

Gives 32ns per loop, so calls to these funcs are almost free. 

Using setup:
```
$ nvidia-smi
Mon Mar 25 16:27:32 2024       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.104.12             Driver Version: 535.104.12   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA GeForce GTX 1050 Ti     On  | 00000000:01:00.0 Off |                  N/A |
| 31%   24C    P8              N/A /  75W |     14MiB /  4096MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+

hdelan:
Some test results:

# CUDA
[gtx1050.txt](https://github.com/oneapi-src/oneMKL/files/14759693/gtx1050.txt)
Some failures due to precision also present on `develop` branch.

# HIP
[gfx90a_oneMKL_test.txt](https://github.com/oneapi-src/oneMKL/files/14759695/gfx90a_oneMKL_test.txt)
Test failures in HIP are also present on the `develop` branch:
[gfx90a_oneMKL_test_develop_branch.txt](https://github.com/oneapi-src/oneMKL/files/14759697/gfx90a_oneMKL_test_develop_branch.txt)

I am not sure how to build/run the FFT tests. Are there some build/test instructions that I can follow?

hdelan:
~~In terms of building with icpx 2024.0.0 for CUDA. I am getting a segfault at linking with `develop` branch.~~

Fixed. `LD_LIBRARY_PATH` problems -_-

I can successfully build this branch with icpx 2024.0.2 for CUDA

Rbiessy:
Thanks a lot @hdelan ! The instructions are [here](https://github.com/oneapi-src/oneMKL/blob/develop/docs/building_the_project.rst) but need to be improved.

The short answer is that you should just need to add `-DENABLE_CUFFT_BACKEND=True -DENABLE_ROCFFT_BACKEND=True` to also test the DFT domain with the native CUDA and HIP backends.
If you are explicitly setting `-DTARGET_DOMAINS` in your CMake command you will also need to append `dft` to the list, otherwise it will be enabled by default.
If you don't want to build and test the other domains again you can use `-DTARGET_DOMAINS=dft`.

hdelan:
Thanks @Rbiessy !

Building rocFFT is broken for me but this PR does not touch that code. Building with cuFFT is OK. Here is updated tests for all oneMKL for CUDA including cuBLAS, cuFFT, cuRAND, cuSOLVER:

 
[gtx1050.txt](https://github.com/oneapi-src/oneMKL/files/14772072/gtx1050.txt)

Rbiessy (approved):

hjabird (approved):

ericlars:
Thanks! LGTM

ericlars (approved):

Rbiessy:
Thanks for the review. Let me know @lhuot or @mmeterel if you need more time, otherwise I will go ahead and merge this on Monday.

PR #426, opened 2023-12-13
[RNG] Refactored the curand backend

andreyfe1:
Refactoring includes following:
* SYCL 2020 API changes.
* Removing unnecessary wait calls for SYCL buffer and USM API both.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? [log.txt](https://github.com/oneapi-src/oneMKL/files/13659464/log.txt)

### DIFF
--- a/src/rng/backends/curand/curand_helper.hpp
+++ b/src/rng/backends/curand/curand_helper.hpp
@@ -211,22 +211,22 @@ static inline void range_transform_fp(sycl::queue& queue, T a, T b, std::int64_t
                                       sycl::buffer<T, 1>& r) {
     queue.submit([&](sycl::handler& cgh) {
         auto acc = r.template get_access<sycl::access::mode::read_write>(cgh);
-        cgh.parallel_for(sycl::range<1>(n),
-                         [=](sycl::id<1> id) { acc[id] = acc[id] * (b - a) + a; });
+        cgh.parallel_for(n, [=](sycl::id<1> id) { acc[id] = acc[id] * (b - a) + a; });
     });
 }
+
 template <typename T>
-static inline sycl::event range_transform_fp(sycl::queue& queue, T a, T b, std::int64_t n, T* r) {
-    return queue.submit([&](sycl::handler& cgh) {
-        cgh.parallel_for(sycl::range<1>(n), [=](sycl::id<1> id) { r[id] = r[id] * (b - a) + a; });
-    });
+static inline sycl::event range_transform_fp(sycl::queue& queue, T a, T b, std::int64_t n, T* r,
+                                             sycl::event dependency) {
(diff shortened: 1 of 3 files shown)


aelizaro (approved):
Changes looks great to me, thank you @andreyfe1!

egrabovskaya (approved):

JackAKirk:
Hi @andreyfe1 

I just have a question about the removal of the `wait_and_throw`s in this PR, In particular I was wondering if there was a reason that you left in the wait_and_throw for the `oneapi::mkl::rng::uniform_method::standard` generate function?

According to curand docs these funcs are asyncronous: see https://docs.nvidia.com/cuda/curand/host-api-overview.html#host-api-example

relevant excerpt:

"Note that curandGenerate() in step 4 above launches a kernel and returns asynchronously. If you launch another kernel in a different stream, and that kernel needs to use the results of curandGenerate(), you must either call cudaThreadSynchronize() or use the stream management/event management routines, to ensure that the random generation kernel has finished execution before the new kernel is launched. "

However we know that actually some cusolver calls that are meant to be asyncronous actually aren't. I'm just curious whether you tested this and found something similar?

Otherwise I may add back stream synchronization to support the standard host_task 2020 case that does not wait for asyncronous native calls.

Thanks.

andreyfe1:
Hi @JackAKirk,
The reason for removal was that `wait_and_throw()` was an extra synchronization. For users it's an extra call during profiling.

>if there was a reason that you left in the wait_and_throw for the `oneapi::mkl::rng::uniform_method::standard` generate function?

Do you mean [this call](https://github.com/oneapi-src/oneMKL/blob/develop/src/rng/backends/curand/mrg32k3a.cpp#L346)? This is because we need to allocate a [temporary memory](https://github.com/oneapi-src/oneMKL/blob/develop/src/rng/backends/curand/mrg32k3a.cpp#L335) and we need to synchronize until we release the memory to not have the dangling reference. Unfortunately this call is synchronous.

JackAKirk:
> Hi @JackAKirk, The reason for removal was that `wait_and_throw()` was an extra synchronization. For users it's an extra call during profiling.
> 
> > if there was a reason that you left in the wait_and_throw for the `oneapi::mkl::rng::uniform_method::standard` generate function?
> 
> Do you mean [this call](https://github.com/oneapi-src/oneMKL/blob/develop/src/rng/backends/curand/mrg32k3a.cpp#L346)? This is because we need to allocate a [temporary memory](https://github.com/oneapi-src/oneMKL/blob/develop/src/rng/backends/curand/mrg32k3a.cpp#L335) and we need to synchronize until we release the memory to not have the dangling reference. Unfortunately this call is synchronous.

Yeah, thanks for the info!

PR #428, opened 2023-12-14
[BLAS][rocBLAS] Update rocBLAS backend

jle-quel:
# Description

This PR updates the rocBLAS backend's functions for LEVEL1, LEVEL2, LEVEL3, EXTENSION and BATCH which used to be unimplemented.

Most ColumnMajor, RowMajor, Buffer and USM functions which used to be marked as `throw unimplemented` are now implemented and does not result in skipped test anymore.

The functions that are still marked as implemented (due to missing a mapping to rocBLAS) are:

- LEVEL1
  - `axpby`
  - `dot(float, float, double)`
- LEVEL2
- LEVEL3
- EXTENSIONS
  - `gemm_bias`
  - `gemmt`
  - `imatcopy`
- BATCH
  - `imatcopy_batch`

~With the exception of `row_major::gemv_batch` from BATCH which will soon be implemented.~
Implemented with the latest commit [Update unimplemented function 'row_major::gemv_batch'](https://github.com/oneapi-src/oneMKL/pull/428/commits/135b76a78ddd1ca0c0a335def798cb66a4a9187f)

This PR also updated some minor parts of already implemented functions:
- Modify the call to `depends_on`. This method can take a vector, so there is no need to loop around it.
- Modify the function argument names so that they are all coherent.
- Switch the call to `wait` to `wait_and_throw`.
- ~Modify the `throw unimplemented` messages to be more accurate and up to date.~ (Restore during PR's review)

# Checklist

## All Submissions

[example_blas_gemm_usm.log.txt](https://github.com/oneapi-src/oneMKL/files/13676802/example_blas_gemm_usm.log.txt)
[test_main_blas_ct.log.txt](https://github.com/oneapi-src/oneMKL/files/13676803/test_main_blas_ct.log.txt)
[test_main_blas_rt.log.txt](https://github.com/oneapi-src/oneMKL/files/13676804/test_main_blas_rt.log.txt)


jinz2014:
I submitted a pull request for bf16 gemm and then found your pull requests for supporting many functions. 

Is it right rocblas supports bf16 gemm ?

https://rocm.docs.amd.com/projects/rocBLAS/en/docs-5.7.1/API_Reference_Guide.html#rocblas-gemm-ex-batched-strided-batched

jle-quel:
Nice catch @jinz2014 

`level3::gemm` does not support bf16; However, the `extension::gemm_ex` does support it.

It also made me realize that `gemm_ex` should be moved out of the `rocblas_level3.cpp` and into the `rocblas_extensions.cpp`.

I'll commit this ASAP.

Would you mind also adding the support for row_major gemm_ex? The code is already present for it, within this PR.

Do you prefer to merge your PR on its own, or merge it as part of this rocBLAS update PR?

jinz2014:
Please merge my PR as part of your rocBLAS update. I will close my PR.

jinz2014:
@andrewtbarker  Could you please add reviewers for the PR ?  Thanks.

src/blas/backends/rocblas/rocblas_batch.cpp
```diff
@@ -188,380 +233,566 @@ GEMM_STRIDED_BATCH_LAUNCHER(std::complex<double>, rocblas_zgemm_strided_batched)
...
+OMATCOPY_STRIDED_BATCH_LAUNCHER(std::complex<float>, rocblas_cgeam_strided_batched)
+OMATCOPY_STRIDED_BATCH_LAUNCHER(std::complex<double>, rocblas_zgeam_strided_batched)
+
+#undef OMATCOPY_STRIDED_BATCH_LAUNCHER
+
 void imatcopy_batch(sycl::queue &queue, transpose trans, int64_t m, int64_t n, float alpha,
                     sycl::buffer<float, 1> &ab, int64_t lda, int64_t ldb, int64_t stride,
                     int64_t batch_size) {
-    throw unimplemented("blas", "imatcopy_batch", "for column_major layout");
+    throw unimplemented("blas", "column_major imatcopy_batch",
```
mmeterel: In order to have consistency across different backend, can we remove these changes in the exception message? Or update the message across all backends accordingly.

src/blas/backends/rocblas/rocblas_batch.cpp
```diff
@@ -732,344 +976,505 @@ TRSM_BATCH_LAUNCHER_USM(std::complex<double>, rocblas_ztrsm_batched)
...
+                queue.submit([&](sycl::handler &cgh) {
+                    auto acc = y.template get_access<sycl::access::mode::read_write>(cgh);
+                    cgh.parallel_for(sycl::range{ (std::size_t)batch_size, (std::size_t)n },
+                                     [=](sycl::item<2> it) {
+                                         auto index = it.get_id(0) * stridey + it.get_id(1) * incy;
+                                         acc[index] = std::conj(acc[index]);
+                                     });
+                });
+            }
+        }
```
mmeterel: Can you explain what is the purpose and need here? Why do we need to take the conjugate of vectors exclusively?

src/blas/backends/rocblas/rocblas_extensions.cpp
```diff
@@ -36,79 +37,157 @@ void gemm_bias(sycl::queue &queue, transpose transa, transpose transb, offset of
...
+    throw unimplemented("blas", "column_major gemmt", "gemm_bias unsupported by the rocBLAS API");
 }
 
 void gemmt(sycl::queue &queue, uplo upper_lower, transpose transa, transpose transb, int64_t n,
            int64_t k, std::complex<double> alpha, sycl::buffer<std::complex<double>, 1> &a,
            int64_t lda, sycl::buffer<std::complex<double>, 1> &b, int64_t ldb,
            std::complex<double> beta, sycl::buffer<std::complex<double>, 1> &c, int64_t ldc) {
-    throw unimplemented("blas", "gemmt", "for column_major layout");
-}
+    throw unimplemented("blas", "column_major gemmt", "gemm_bias unsupported by the rocBLAS API");
```
mmeterel: same comment as before  for this all such changes.

src/blas/backends/rocblas/rocblas_extensions.cpp
```diff
@@ -36,79 +37,157 @@ void gemm_bias(sycl::queue &queue, transpose transa, transpose transb, offset of
...
+}
+
+template <typename Func, typename T_A, typename T_B, typename T_C, typename T_S,
+          typename DATATYPE_A, typename DATATYPE_B, typename DATATYPE_C, typename COMPUTETYPE>
+inline void gemm_ex(Func func, DATATYPE_A DT_A, DATATYPE_B DT_B, DATATYPE_C DT_C, COMPUTETYPE CT,
```
mmeterel: In order to align with [spec ](https://github.com/oneapi-src/oneAPI-spec/blob/main/source/elements/oneMKL/source/domains/blas/gemm.rst) these extensions should be part GEMM API and needs to be placed in level3. I understand, their names are `_ex` for rocblas, however, we should follow oneMKL Spec here.

src/blas/backends/rocblas/rocblas_extensions.cpp
```diff
@@ -36,79 +37,157 @@ void gemm_bias(sycl::queue &queue, transpose transa, transpose transb, offset of
...
-}
+#undef GEMM_EX_LAUNCHER
 
-void omatcopy(sycl::queue &queue, transpose trans, int64_t m, int64_t n, std::complex<double> alpha,
-              sycl::buffer<std::complex<double>, 1> &a, int64_t lda,
-              sycl::buffer<std::complex<double>, 1> &b, int64_t ldb) {
-    throw unimplemented("blas", "omatcopy", "for column_major layout");
-}
+template <typename Func, typename T>
+inline void omatcopy(Func func, sycl::queue &queue, transpose trans, int64_t m, int64_t n,
```
mmeterel, line 93: Agreed, this one belongs here.

src/blas/backends/rocblas/rocblas_level1.cpp
```diff
@@ -69,10 +69,12 @@ inline void asum(Func func, sycl::queue &queue, int64_t n, sycl::buffer<T1, 1> &
               sycl::buffer<TYPE2, 1> &result) {                                             \
         asum(ROCBLAS_ROUTINE, queue, n, x, incx, result);                                   \
     }
+
```
mmeterel, line 72: Are these added manually or by clang-format? If manual, I would prefer avoiding them since it not only adds inconsistency with other backend but also complicates review.

src/blas/backends/rocblas/rocblas_level3.cpp
```diff
@@ -71,62 +74,13 @@ GEMM_LAUNCHER(std::complex<double>, rocblas_zgemm)
 
 #undef GEMM_LAUNCHER
 
-void gemm(sycl::queue &queue, transpose transa, transpose transb, int64_t m, int64_t n, int64_t k,
```
mmeterel, line 74: As mentioned before, lets keep all flavors of gemm in level3.

mmeterel (commented):
@jle-quel Thanks a lot for this high quality PR. It is adding features that do not even exist in vendor libraries. 
I have a few cosmetic comments and some concerns about the structure and consistency across entire project.

src/blas/backends/rocblas/rocblas_batch.cpp
jle-quel: I can revert these changes since modifying the message across all backends will be out of scope for the description of this PR.

src/blas/backends/rocblas/rocblas_level1.cpp
jle-quel: These newlines were added manually since most of the rocBLAS functions have these.
I added these newlines for the ones that were missing, for consistency across the different levels of rocBLAS.

Do you still want me to restore it?

src/blas/backends/rocblas/rocblas_batch.cpp
jle-quel: ## Can you explain what is the purpose and need here?

The reason needed for `x` to be conjugated is that we are dealing in this case with row-major layout.

Emulating the transpose operation can be done by switching the "propriety" arguments of the vectors/matrices, thus modifying the way the inputs are accessed as opposed to physically transposing the inputs.

However for complex conjugations transpose, for elements in vector/matrices, this operation must be done
explicitly since there is no way to "emulate" the effect of the conjugate transpose operation.

## Why do we need to take the conjugate of vectors exclusively?

From my readings during the work of this PR, I realized that for the TRSM function, as an example, which does not need to have the conjugate emulation, is because triangular matrices do not change upon transposition.

Proprieties like the one mentioned for TRSM are the reasons manual conjugation is needed or not needed.

Also, the source of truth for the oneMKL's tests is the CBLAS's lapack reference implementation, and this is how they implemented them

Hope this helps, let me know otherwise

src/blas/backends/rocblas/rocblas_batch.cpp
jle-quel: However, I thought that this could be the entry point for having more explicit exception messages, explaining the reason for it to be unimplemented.

src/blas/backends/rocblas/rocblas_extensions.cpp
jle-quel: Done

src/blas/backends/rocblas/rocblas_batch.cpp
mmeterel: Sure, lets revert them for the sake of consistency.

src/blas/backends/rocblas/rocblas_batch.cpp
mmeterel: Thanks for the explanation. Makes sense.

src/blas/backends/rocblas/rocblas_level1.cpp
mmeterel: We can keep them. Thanks.

src/blas/backends/rocblas/rocblas_batch.cpp
jle-quel: Done

muhammad-tanvir-1211:
I see all the buffer tests failing for the `rocblas` backend with `PI_ERROR_INVALID_OPERATION`.

Logs:
[PR_428.txt](https://github.com/oneapi-src/oneMKL/files/13918449/PR_428.txt)

This PR is blocked on https://github.com/oneapi-src/unified-runtime/pull/1226 and https://github.com/intel/llvm/pull/12297

jle-quel:
> I see all the buffer tests failing for the `rocblas` backend with `PI_ERROR_INVALID_OPERATION`.
> 
> Logs: [PR_428.txt](https://github.com/oneapi-src/oneMKL/files/13918449/PR_428.txt)
> 
> This PR is blocked on [oneapi-src/unified-runtime#1226](https://github.com/oneapi-src/unified-runtime/pull/1226) and [intel/llvm#12297](https://github.com/intel/llvm/pull/12297)

The reason that the buffer tests fails (`PI_ERROR_INVALID_OPERATION` ) is not because of this PR.

It is because the "Accessor interop requires urMemGetNativeHandle to return a valid mem handle. Multi dev ctx breaks this ..."

This means that if you go to a previous commit, the results will be the same.
Defacto this PR is not blocked; this PR lives independently from the PRs that you provided.
However, to test it, you either have to pull the modifications made from the PRs that you provided or wait for them to be merged (and this logic does not only apply to this PR, but to any "buffers function"-"accessor interop" from oneMKL)

muhammad-tanvir-1211:
> > I see all the buffer tests failing for the `rocblas` backend with `PI_ERROR_INVALID_OPERATION`.
> > Logs: [PR_428.txt](https://github.com/oneapi-src/oneMKL/files/13918449/PR_428.txt)
> > This PR is blocked on [oneapi-src/unified-runtime#1226](https://github.com/oneapi-src/unified-runtime/pull/1226) and [intel/llvm#12297](https://github.com/intel/llvm/pull/12297)
> 
> The reason that the buffer tests fails (`PI_ERROR_INVALID_OPERATION` ) is not because of this PR.
> 
> It is because the "Accessor interop requires urMemGetNativeHandle to return a valid mem handle. Multi dev ctx breaks this ..."
> 
> This means that if you go to a previous commit, the results will be the same. Defacto this PR is not blocked; this PR lives independently from the PRs that you provided. However, to test it, you either have to pull the modifications made from the PRs that you provided or wait for them to be merged (and this logic does not only apply to this PR, but to any "buffers function"-"accessor interop" from oneMKL)

Yes, that is what I meant actually, but I should have worded my comment better. Thank you for the correction. 
The two PRs mentioned in my comment will hopefully help resolve the failing tests, and the failures are not directly because of the changes made in this PR.

src/blas/backends/rocblas/rocblas_level2.cpp
```diff
@@ -1742,15 +1752,70 @@ TRSV_LAUNCHER_USM(std::complex<double>, rocblas_ztrsv)
...
+
+        if (m > 0) {
+            queue.submit([&](sycl::handler &cgh) {
+                const auto abs_incx = std::abs(incx);
+                auto acc = x.template get_access<sycl::access::mode::read_write>(cgh);
+                cgh.parallel_for(sycl::range{ (std::size_t)m }, [=](sycl::id<1> id) {
+                    const auto index = id * abs_incx;
+                    acc[index] = std::conj(acc[index]);
+                });
+            });
```
andrewtbarker: I wonder if it makes sense to define an inline function (or a lambda or maybe even a preprocessor macro) to wrap this, something like:
```c++
template <typename fp>
static inline conj_vector(sycl::queue& queue, sycl::buffer<fp>& x,
                          int64_t m, int64_t incx) {
}
```
then the contents of the function would be lines 1773-1780 above. Then you can reduce a lot of the code duplication below by reusing this function.

src/blas/backends/rocblas/rocblas_level2.cpp
```diff
@@ -2193,11 +2781,73 @@ TRSV_LAUNCHER(std::complex<double>, rocblas_ztrsv)
...
+
+    auto new_trans = trans == oneapi::mkl::transpose::nontrans ? oneapi::mkl::transpose::trans
+                                                               : oneapi::mkl::transpose::nontrans;
+
+    if (trans == oneapi::mkl::transpose::conjtrans) {
+        alpha = std::conj(alpha);
+        beta = std::conj(beta);
+
+        if (m > 0) {
+            done = queue.submit([&](sycl::handler &cgh) {
```
andrewtbarker: I suggest a corresponding inline function here to reduce code duplication in the USM code path - it is very similar but will need a `const std::vector<sycl::event> &dependencies` argument (or maybe you can get away with a single `sycl::event` dependency).

What we actually should do (but probably won't) is implement a highly optimized [lacgv](https://netlib.org/lapack/explore-html/d9/d50/group__lacgv.html) API and just use that.

andrewtbarker (commented):
This is a nice PR. I have a suggestion to reduce code duplication in the level 2 BLAS interfaces, but generally I'm very happy with this.

src/blas/backends/rocblas/rocblas_level2.cpp
jle-quel: Done

src/blas/backends/rocblas/rocblas_level2.cpp
jle-quel: I agree with the `lacgv` API; I committed a simple inline function (like you proposed in your other feedback) to remove the duplicate code for this PR.
As part of another PR, another approach could be proposed.
What do you think?

andrewtbarker (approved):
This is very nice, approved!

src/blas/backends/rocblas/rocblas_level2.cpp
andrewtbarker: Yes, what you've done is perfect for this PR, `lacgv` would be a much larger discussion impacting all the backends.

mmeterel (approved):
Thanks for updates. Looks good.

PR #430, opened 2023-12-15
Allow multiple CUDA backends to be enabled at once

### LINKED ISSUE #127: Building tests with both cuBLAS and cuRAND enabled hits compilation issues
# Summary

Trying to build the tests with both cuBLAS and cuRAND backends enabled hit compilation issues.

# Version

Using the latest version: 19c43b0f68dcaba5bc518d5becdfa0097bebfa7c

# Environment

This should be pretty straightforward to reproduce in any environment but let me know if you need more details for this section.

# Steps to reproduce

The issue happens if all of these are provided:
* `-DENABLE_CUBLAS_BACKEND=True`
* `-DENABLE_CURAND_BACKEND=True`
* `-DBUILD_FUNCTIONAL_TESTS=True`

# Observed behavior

This was discussed in #91 and https://github.com/oneapi-src/oneMKL/pull/126#discussion_r704654552, however it looks like that ticket fixed the issue for building one of them but not both of them at the same time.

`TARGET_DOMAINS` also doesn't seem to work properly with cuBLAS and cuRAND, it always gets reset here:
* https://github.com/oneapi-src/oneMKL/blob/develop/CMakeLists.txt#L108

Because it's checking the specified target domains against a domain list that is only populated when using the MKLCPU or MKLGPU backends.

I tried hacking that part of the CMake to let a `blas;rng` target domain list go through, however this still seemed to fail when building the tests.

# Expected behavior

I see two main possible outcomes for this:

* Mark this as unsupported and make CMake throw out an error if someone is trying to build this configuration
* Fix the build issue/CMake so that this combination is supported

andrewtbarker:
# Description

For compile-time linking our tests currently assume only one CUDA backend (eg cublas, curand, cusolver) is enabled at once. This PR fixes this limitation.

Fixes #127

The substantive changes are only in `tests/unit_tests/include/test_helper.hpp`, all the other changes are just text replacement of some macro names.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? [test_issue127.txt](https://github.com/oneapi-src/oneMKL/files/13688469/test_issue127.txt)

## Bug fixes

- [ ] Have you added relevant regression tests? (not applicable)
      GitHub issue or in this PR)?

### DIFF
--- a/tests/unit_tests/blas/batch/axpy_batch_stride.cpp
+++ b/tests/unit_tests/blas/batch/axpy_batch_stride.cpp
@@ -120,13 +120,14 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t incx, int64_t incy, fp
 #else
         switch (layout) {
             case oneapi::mkl::layout::col_major:
-                TEST_RUN_CT_SELECT(main_queue, oneapi::mkl::blas::column_major::axpy_batch, n,
-                                   alpha, x_buffer, incx, stride_x, y_buffer, incy, stride_y,
-                                   batch_size);
+                TEST_RUN_BLAS_CT_SELECT(main_queue, oneapi::mkl::blas::column_major::axpy_batch, n,
+                                        alpha, x_buffer, incx, stride_x, y_buffer, incy, stride_y,
+                                        batch_size);
                 break;
             case oneapi::mkl::layout::row_major:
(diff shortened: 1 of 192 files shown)


mmeterel (approved):
Thanks for fixing this long standing issue.

jinz2014:
I browsed your changes and learned that different kinds of tests (lapack, blas, ct) need different macros. Thank you.

mkrainiuk (approved):

PR #432, opened 2023-12-18
[BLAS][HIP] Enable bf16 column-major GEMM in the ROCm backend

jinz2014:
Thank you for your reviews.


jinz2014:
Close the PR because it will be part of #428.

PR #448, opened 2024-02-20
[rocBLAS] Fix issues with rocBLAS 4.0.0

nilsfriess:
# Description

rocBLAS 4.0.0 [removed support for the legacy BLAS inplace trmm](https://github.com/ROCm/rocBLAS/blob/develop/CHANGELOG.md#removals), so calls to trmm cause a compilation error. The old functionality is provided by just duplicating the last two arguments of trmm (see also [here](https://rocblas.readthedocs.io/en/master/API_Reference_Guide.html#announced-in-rocblas-3-0)).

To fix this, I am defining a macro `ROCBLAS_NO_LEGACY_TRMM` depending on the rocBLAS version which is then used in an `#ifdef` to decide which version of the function to call. 

I also had to change slightly the way the path to `librocblas.so` is constructed in `FindrocBLAS.cmake`. I'm not sure if this is also related to the rocBLAS version or if this is a general issue on some systems, so I can also put that in a separated PR if you prefer.  

# Checklist

## All Submissions

  I don't have an AMD gpu to test, I'm just assessing the state of compiling oneMKL with AdaptiveCpp to see if #307 needs any updates and can the maybe be merged. 

### DIFF
--- a/src/blas/backends/rocblas/rocblas_level3.cpp
+++ b/src/blas/backends/rocblas/rocblas_level3.cpp
@@ -381,10 +381,19 @@ inline void trmm(Func func, sycl::queue &queue, side left_right, uplo upper_lowe
             auto a_ = sc.get_mem<rocDataType *>(a_acc);
             auto b_ = sc.get_mem<rocDataType *>(b_acc);
             rocblas_status err;
+
+// rocblas version 4.0.0 removed the legacy BLAS trmm implementation
+#ifdef ROCBLAS_NO_LEGACY_TRMM
             ROCBLAS_ERROR_FUNC_SYNC(func, err, handle, get_rocblas_side_mode(left_right),
                                     get_rocblas_fill_mode(upper_lower),
                                     get_rocblas_operation(trans), get_rocblas_diag_type(unit_diag),
+                                    m, n, (rocDataType *)&alpha, a_, lda, b_, ldb, b_, ldb);
+#else
+	    ROCBLAS_ERROR_FUNC_SYNC(func, err, handle, get_rocblas_side_mode(left_right),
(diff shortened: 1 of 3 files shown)


mmeterel:
@nilsfriess Thanks for the PR. I think this needs to be tested. You can contact me or infra team ( @toxicscum ) to access AMD GPUs.

src/blas/backends/rocblas/rocblas_level3.cpp
```diff
@@ -381,10 +381,19 @@ inline void trmm(Func func, sycl::queue &queue, side left_right, uplo upper_lowe
             auto a_ = sc.get_mem<rocDataType *>(a_acc);
             auto b_ = sc.get_mem<rocDataType *>(b_acc);
             rocblas_status err;
+
+// rocblas version 4.0.0 removed the legacy BLAS trmm implementation
+#ifdef ROCBLAS_NO_LEGACY_TRMM
```
npmiller, line 386:
```suggestion
#if ROCBLAS_VERSION_MAJOR >= 3
```
We can use `ROCBLAS_VERSION_MAJOR`, that way we don't need to add a new define, also since it's deprecated in rocBLAS 3 it gives warning so we could also use the new interface for rocBLAS 3.x.

I ran into the same issue and ended up with a similar patch before finding this PR. I didn't have any `ROCBLAS_LIB_PATH` issues with rocBLAS 4.x though.

Rbiessy:
Thank you for the PR, we have recently removed the `Findroc*.cmake` files in https://github.com/oneapi-src/oneMKL/pull/490 which should avoid the issues with `ROCBLAS_LIB_PATH`. Can you rebase your PR with the tip of develop?

PR #450, opened 2024-02-22
Update SYCL support discovery to check for `-fsycl` flag.

### LINKED ISSUE #449: oneMKL's CMake Does Not Detect SYCL Support when Compiled with `intel/llvm`
# Summary
oneMKL uses [`CXX_COMPILER_NAME` to detect SYCL support](https://github.com/oneapi-src/oneMKL/blob/develop/cmake/mkl/MKLConfig.cmake#L211).  This means that if a user compiles their own copy of [`intel/llvm`](https://github.com/intel/llvm), SYCL support is not detected and the `MKL::MKL_SYCL` library will not be available.  (The `CXX_COMPILER_NAME` is by default `Clang`.)

Users can force set `SYCL_COMPILER` with `-DSYCL_COMPILER=ON`, which will make the `MKL::MKL_SYCL` library available.  However, with `CLANG_COMPILER` also set, the `intel_thread` backend is also made available.  This elicits a compiler warning:

```bash
bbrock@ortce-skl22:~/src/distributed-ranges/build/examples/shp$ make -j
Consolidate compiler generated dependencies of target fmt
Warning: Using MKL::MKL_SYCL* targets with intel_thread may have potential composability problems on CPU device with other SYCL kernels.
[  5%] Building CXX object _deps/cpp-format-build/CMakeFiles/fmt.dir/src/os.cc.o
```

Which seems to indicate this is still a sub-optimal solution.

# Version
oneMKL `develop` branch.

# Environment
* Running on a system with two SPRs and 8 PVC GPUs.
* Both production oneMKL `2024.0` and oneMKL develop.
* Ubuntu 22.04 LTS
* Fresh [`intel/llvm`](https://github.com/intel/llvm)

[CMake output](https://gist.github.com/BenBrock/65bb26fae33ed11e7ecad9df59c6cf17)

# Steps to reproduce
* Compile fresh [`intel/llvm`](https://github.com/intel/llvm)
* Compile a project, including oneMKL via CMake (in my case [distributed ranges](https://github.com/oneapi-src/distributed-ranges))

# Observed behavior
[Build fails](https://gist.github.com/BenBrock/65bb26fae33ed11e7ecad9df59c6cf17) due to `MKL::MKL_DPCPP`/`MKL::MKL_SYCL` library not being available.

# Expected behavior
I expect SYCL support to be detected by oneMKL and the build to succeed.

BenBrock:
# Description
oneMKL's CMake currently checks `CXX_COMPILER_NAME` and uses that to determine whether SYCL is supported.  This prevents SYCL support from being detected if the user compiles with their own build of `intel/llvm` (see issue #449).

This is a first stab at modifying `MKLConfig.cmake` to check for `-fsycl` support directly instead of inferring from the compiler name.  It is based on h[ow oneDPL checks for SYCL support](https://github.com/oneapi-src/oneDPL/blob/main/CMakeLists.txt#L79), which works for both release builds of IntelLLVM and custom builds of `intel/llvm`.

I have also modified the CMake to avoid setting `CLANG_COMPILER` if `SYCL_COMPILER` is set, as this is the behavior when using IntelLLVM.  I assume open-source DPC++ builds should behave similarly.

Fixes #449

# Checklist

## All Submissions

## New interfaces

it was accepted? # (RFC)

## New features

## Bug fixes

      GitHub issue or in this PR)?

### DIFF
--- a/cmake/mkl/MKLConfig.cmake
+++ b/cmake/mkl/MKLConfig.cmake
@@ -208,12 +208,28 @@ if(CMAKE_Fortran_COMPILER)
 endif()
 
 # Determine Compiler Family
-if(CXX_COMPILER_NAME STREQUAL "dpcpp" OR CXX_COMPILER_NAME STREQUAL "dpcpp.exe"
-    OR CXX_COMPILER_NAME STREQUAL "icpx" OR CXX_COMPILER_NAME STREQUAL "icx.exe")
-  set(SYCL_COMPILER ON)
+
+include(CMakePackageConfigHelpers)
+include(CheckCXXCompilerFlag)
+include(CheckIncludeFileCXX)
+include(GNUInstallDirs)
+
+# Check SYCL support by the compiler
+check_cxx_compiler_flag("-fsycl" _fsycl_option)
+if (_fsycl_option)
+  CHECK_INCLUDE_FILE_CXX("sycl/sycl.hpp" _sycl_header "-fsycl")
+  if (NOT _sycl_header)
+    CHECK_INCLUDE_FILE_CXX("CL/sycl.hpp" _sycl_header_old "-fsycl")
+  endif()
+  if (_sycl_header OR _sycl_header_old)
+    set(SYCL_COMPILER ON)
+  endif()
 endif()
-if(C_COMPILER_NAME MATCHES "^clang" OR CXX_COMPILER_NAME MATCHES "^clang")
-  set(CLANG_COMPILER ON)
+
+if(NOT DEFINED SYCL_COMPILER OR SYCL_COMPILER MATCHES OFF)
+  if(C_COMPILER_NAME MATCHES "^clang" OR CXX_COMPILER_NAME MATCHES "^clang")
+    set(CLANG_COMPILER ON)
+  endif()
 endif()
 if(CMAKE_C_COMPILER_ID STREQUAL "PGI" OR CMAKE_CXX_COMPILER_ID STREQUAL "PGI" OR CMAKE_Fortran_COMPILER_ID STREQUAL "PGI"
     OR CMAKE_C_COMPILER_ID STREQUAL "NVHPC" OR CMAKE_CXX_COMPILER_ID STREQUAL "NVHPC"


cmake/mkl/MKLConfig.cmake
```diff
@@ -208,12 +208,28 @@ if(CMAKE_Fortran_COMPILER)
...
+    CHECK_INCLUDE_FILE_CXX("CL/sycl.hpp" _sycl_header_old "-fsycl")
+  endif()
+  if (_sycl_header OR _sycl_header_old)
+    set(SYCL_COMPILER ON)
+  endif()
 endif()
-if(C_COMPILER_NAME MATCHES "^clang" OR CXX_COMPILER_NAME MATCHES "^clang")
-  set(CLANG_COMPILER ON)
+
+if(NOT DEFINED SYCL_COMPILER OR SYCL_COMPILER MATCHES OFF)
```
mkrainiuk, line 229: This approach could not work correctly for case when MKLConfig is used for building non-SYCL app with clang compiler that supports SYCL, but I believe this fix is enough for oneMKL project build. We can update this part later with more general fix from Intel oneMKL

mkrainiuk (approved):
Thank you for the fix! I have one minor comment but overall looks good to me.

cmake/mkl/MKLConfig.cmake
```diff
@@ -208,12 +208,28 @@ if(CMAKE_Fortran_COMPILER)
...
 
 # Determine Compiler Family
-if(CXX_COMPILER_NAME STREQUAL "dpcpp" OR CXX_COMPILER_NAME STREQUAL "dpcpp.exe"
-    OR CXX_COMPILER_NAME STREQUAL "icpx" OR CXX_COMPILER_NAME STREQUAL "icx.exe")
-  set(SYCL_COMPILER ON)
+
+include(CMakePackageConfigHelpers)
+include(CheckCXXCompilerFlag)
+include(CheckIncludeFileCXX)
+include(GNUInstallDirs)
```
dnhsieh-intel, line 215:
```suggestion
include(CheckCXXCompilerFlag)
include(CheckIncludeFileCXX)
```
I think we can delete the first line `include(CMakePackageConfigHelpers)` and the fourth line `include(GNUInstallDirs)`. `CheckCXXCompilerFlag` is needed for `check_cxx_compiler_flag()`, and `CheckIncludeFileCXX` is needed for `CHECK_INCLUDE_FILE_CXX()`.

cmake/mkl/MKLConfig.cmake
```diff
@@ -208,12 +208,28 @@ if(CMAKE_Fortran_COMPILER)
 endif()
 
 # Determine Compiler Family
-if(CXX_COMPILER_NAME STREQUAL "dpcpp" OR CXX_COMPILER_NAME STREQUAL "dpcpp.exe"
-    OR CXX_COMPILER_NAME STREQUAL "icpx" OR CXX_COMPILER_NAME STREQUAL "icx.exe")
-  set(SYCL_COMPILER ON)
+
+include(CMakePackageConfigHelpers)
+include(CheckCXXCompilerFlag)
```
dnhsieh-intel, line 213: Could you put this block in a condition please?
```
if("CXX" IN_LIST CURR_LANGS)
  include(CheckCXXCompilerFlag)
  include(CheckIncludeFileCXX)

  check_cxx_compiler_flag("-fsycl" _fsycl_option)
  ...
      if (_sycl_header OR _sycl_header_old)
      set(SYCL_COMPILER ON)
    endif()
  endif()
endif()
```
Although only CXX is used in oneMKL interfaces, oneMKL also needs to support C and Fortran. There is a possibility that users would use `MKLConfig.cmake` here with oneMKL.

cmake/mkl/MKLConfig.cmake
dnhsieh-intel, line 229: Did you mean
```suggestion
if(NOT SYCL_COMPILER)
```

cmake/mkl/MKLConfig.cmake
```diff
@@ -208,12 +208,28 @@ if(CMAKE_Fortran_COMPILER)
...
+include(CheckIncludeFileCXX)
+include(GNUInstallDirs)
+
+# Check SYCL support by the compiler
+check_cxx_compiler_flag("-fsycl" _fsycl_option)
+if (_fsycl_option)
+  CHECK_INCLUDE_FILE_CXX("sycl/sycl.hpp" _sycl_header "-fsycl")
+  if (NOT _sycl_header)
+    CHECK_INCLUDE_FILE_CXX("CL/sycl.hpp" _sycl_header_old "-fsycl")
```
dnhsieh-intel, line 222: `SYCL_SYCL_HPP` and `CL_SYCL_HPP` might be more expressive than `_sycl_header` and `_sycl_header_old`. It would be great if you also use `COMPILER_FSYCL_SUPPORT` instead of `_fsycl_option`. Please note that the convention in `MKLConfig.cmake` is to use all caps for variable names.

In addition, please consider removing the space after `if` to match the style in the file:
```
if(<condition>)
```

dnhsieh-intel (commented):
Thank you for the PR! I have some suggestions.

PR #463, opened 2024-03-25
Update .gitignore to ignore all build dirs

hdelan:
Changing build/ to build*/ means that all build dirs can be ignored, so I can maintain separate `build/` `build_lapack/` `build_blas/` etc. directories that will all be successfully ignored by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,4 @@
 .git/
 
 # Build
-build/
+build*/


Rbiessy (approved):

hjabird (approved):

mkrainiuk (approved):
Good idea

Issue #446, opened 2024-02-15
Missing gemm_batch data types


AidanBeltonS:
# Summary
I believe there are some missing gemm_batch implementations, looking at the oneMKL docs it seems this should support. A `gemm_batch` with, two half matrices as input, a float matrix out, and float scaling. My reference: https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2023-0/gemm-batch.html 
I run into issues of this overload not being found. Is my documentation correct, or have I misunderstood something?

# Version
oneMKL hash: 7d2044e202dbc67ff4eee598a4392edcd163deaf

# Environment
oneMKL works with multiple HW and backend libraries and also depends on the
compiler and build environment. Include
the following information to help reproduce the issue:
* HW: A100 GPU
* Backend: cuBlas
* OS: Ubuntu 20.04
* Compiler version: DPC++ 2024.0.2

# Steps to reproduce
Compile with for NVidia GPUs: `icpx -fsycl -fsycl-targets=nvptx64-nvidia-cuda reproducer_onemkl_batch.cpp -lonemkl`
or for Intel GPUs: `icpx -fsycl reproducer_onemkl_batch.cpp -lonemkl`

```
#include <sycl/sycl.hpp>
#include <oneapi/mkl.hpp>

template <class Ta, class Tb, class Tc, class Ts>
void run_gemm(sycl::queue q) {
    // Construct some arbitrary data, error is in compilation, so it does not have to be correct.
    const Ta *a[4] = {nullptr};
    const Tb *b[4] = {nullptr};
    Tc *c[4] = {nullptr};

    int64_t batch_size = 4;

    oneapi::mkl::transpose a_trans = oneapi::mkl::transpose::trans;
    oneapi::mkl::transpose b_trans = oneapi::mkl::transpose::nontrans;

    int64_t m = 10;
    int64_t n = 10;
    int64_t k = 10;

    int64_t lda = 10;
    int64_t ldb = 10;
    int64_t ldc = 10;

    int64_t group_size = 1;

    Ts alpha = 1;
    Ts beta = 0;
    oneapi::mkl::transpose *trans =
        reinterpret_cast<oneapi::mkl::transpose *>( 
            std::malloc(sizeof(oneapi::mkl::transpose) * 2 * batch_size));
    for (int batch = 0; batch < batch_size; ++batch) {
      trans[batch + batch_size * 0] = a_trans;
[...]

PR #445, opened 2024-02-14
[BLAS] fix iamin/iamax header to correspond to oneMKL backend

andrewtbarker:
# Description

The local headers in `src/blas/backends/mkl_common/mkl_blas_backend.hxx` correspond to deprecated symbols in the oneMKL library backend. When those symbols are removed the oneMKL interfaces will no longer build with the oneMKL backend. This PR fixes the headers to avoid this.

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Attach a log. 
[test.txt](https://github.com/oneapi-src/oneMKL/files/14273891/test.txt)
- [x] Have you formatted the code using clang-format?

Testing was done with the 2024.0 version of the oneMKL library for the backend. One test (`blas/EXAMPLE/RT/gemm_usm/gpu`) fails but it is unrelated to this PR.

PR #466, opened 2024-04-03
Add new batch_gemm types

AidanBeltonS:
# Description

This adds new data types for the gemm_batch operation, to better be in line with the oneMKL spec. The types added are <half, half, float, float>, <int8, int8, float, float>, and <int8, int8, int32, float>.

New testing is added for these data types. Tests where the scalar type does not match the input type require a higher tolerance as the reference calculation is being performed at a much higher precision. 

Test logs:
[rocblas_test_log.txt](https://github.com/oneapi-src/oneMKL/files/14853370/rocblas_test_log.txt)
[cublas_test_log.txt](https://github.com/oneapi-src/oneMKL/files/14853373/cublas_test_log.txt)

I have been unable to test the mkl backends as I was running into some problems regarding duplicate definitions between the mkl headers and the openBlas/CBlas headers.

Fixes # (GitHub issue)
https://github.com/oneapi-src/oneMKL/issues/446

# Checklist

## All Submissions

- [x]Have you formatted the code using clang-format?

## New interfaces

it was accepted? # (RFC)

## New features


AidanBeltonS:
@Rbiessy, cc

mmeterel:
@AidanBeltonS  Thanks for the PR. Before going through the review in more detail, what is your plan for this issue? Why openBLAS come into picture here?  I would prefer to have all applicable backends working before adding these new APIs. 

"I have been unable to test the mkl backends as I was running into some problems regarding duplicate definitions between the mkl headers and the openBlas/CBlas headers."

src/blas/backends/cublas/cublas_helper.hpp
```diff
@@ -231,6 +231,56 @@ inline cublasSideMode_t get_cublas_side_mode(oneapi::mkl::side lr) {
...
+
+template <>
+inline cudaDataType_t get_cublas_datatype<std::int32_t>() {
+    return CUDA_R_32I;
+}
+
+template <>
+inline cudaDataType_t get_cublas_datatype<std::uint32_t>() {
+    return CUDA_R_8U;
+}
```
hjabird: This doesn't look right.

src/blas/backends/portblas/portblas_batch.cxx
```diff
@@ -700,6 +727,33 @@ sycl::event gemm_batch(sycl::queue &queue, oneapi::mkl::transpose *transa,
     throw unimplemented("blas", "gemm_batch", " for USM");
 }
 
+sycl::event gemm_batch(sycl::queue &queue, oneapi::mkl::transpose *transa,
+                       oneapi::mkl::transpose *transb, std::int64_t *m, std::int64_t *n,
+                       std::int64_t *k, float *alpha, const sycl::half **a, std::int64_t *lda,
+                       const sycl::half **b, std::int64_t *ldb, float *beta, float **c,
+                       std::int64_t *ldc, std::int64_t group_count, std::int64_t *group_size,
+                       const std::vector<sycl::event> &dependencies) {
+    throw unimplemented("blas", "gemm_batch", " for USM");
```
hjabird, line 736: I'm not sure which portBLAS version gets used in oneMKL right now, but isn't USM supported @s-Nick @pgorlani ?

src/blas/backends/rocblas/rocblas_helper.hpp
```diff
@@ -205,6 +205,66 @@ inline rocblas_side get_rocblas_side_mode(oneapi::mkl::side lr) {
...
 }
 
+template <typename T>
+inline rocblas_datatype get_rocblas_datatype() {
+    static_assert(false);
+}
+
+template <>
+inline rocblas_datatype get_rocblas_datatype<rocblas_half>() {
+    return rocblas_datatype_f16_r;
```
hjabird, line 216: Was there a reason for these not to be `constexpr`? I don't think the code needs to be changed since these are being used in a very limited scope, but I'm curious.

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
```diff
@@ -99,19 +99,28 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t batch_size) {
...
         rand_matrix(C.data() + stride_c * i, layout, oneapi::mkl::transpose::nontrans, m, n, ldc);
     }
 
-    C_ref = C;
+    for (size_t i = 0; i < A.size(); ++i)
+        A_ref[i] = A[i];
+    for (size_t i = 0; i < B.size(); ++i)
+        B_ref[i] = B[i];
+    for (size_t i = 0; i < C.size(); ++i)
+        C_ref[i] = C[i];
```
hjabird: I know that the existing code doesn't always conform to it, but there is a [style guide](https://github.com/oneapi-src/oneMKL/blob/develop/CONTRIBUTING.md#coding-style). There are a few places where it calls for putting `{}` around the body of your for loops.

hjabird (commented):
Since the reference cblas implementation doesn't support some of the operations that are being added (as I understand it), is the new functionality actually tested?

Rbiessy:
> @AidanBeltonS Thanks for the PR. Before going through the review in more detail, what is your plan for this issue? Why openBLAS come into picture here? I would prefer to have all applicable backends working before adding these new APIs.
> 
> "I have been unable to test the mkl backends as I was running into some problems regarding duplicate definitions between the mkl headers and the openBlas/CBlas headers."

Hey @mmeterel, I checked with Aidan about the issue with the MKL backends. The duplicate definitions seemed to be an issue with the setup or build commands used. We ran into another issue with undefined references with `iamax` and `iamin` functions using buffers using 2024.1 oneAPI base toolkit. Just a few example:
```
/usr/bin/ld: lib/libonemkl_blas_mklcpu.so.0: undefined reference to `oneapi::mkl::blas::row_major::iamax(sycl::_V1::queue&, long, sycl::_V1::buffer<std::complex<double>, 1, sycl::_V1::detail::aligned_allocator<std::complex<double> >, void>&, long, sycl::_V1::buffer<long, 1, sycl::_V1::detail::aligned_allocator<long>, void>&)'
/usr/bin/ld: lib/libonemkl_blas_mklcpu.so.0: undefined reference to `oneapi::mkl::blas::column_major::iamin(sycl::_V1::queue&, long, sycl::_V1::buffer<double, 1, sycl::_V1::detail::aligned_allocator<double>, void>&, long, sycl::_V1::buffer<long, 1, sycl::_V1::detail::aligned_allocator<long>, void>&)'
/usr/bin/ld: lib/libonemkl_blas_mklcpu.so.0: undefined reference to `oneapi::mkl::blas::column_major::iamin(sycl::_V1::queue&, long, sycl::_V1::buffer<std::complex<double>, 1, sycl::_V1::detail::aligned_allocator<std::complex<double> >, void>&, long, sycl::_V1::buffer<long, 1, sycl::_V1::detail::aligned_allocator<long>, void>&)'
/usr/bin/ld: lib/libonemkl_blas_mklcpu.so.0: undefined reference to `oneapi::mkl::blas::column_major::iamax(sycl::_V1::queue&, long, sycl::_V1::buffer<float, 1, sycl::_V1::detail::aligned_allocator<float>, void>&, long, sycl::_V1::buffer<long, 1, sycl::_V1::detail::aligned_allocator<long>, void>&)'
```

Looking at `libmkl_sycl_blas.so.4` in 2024.1 these functions expect an `index_base` as a last argument but is not there in oneMKL:
```
$ readelf -Wa /path/to/mkl/latest/lib/libmkl_sycl_blas.so.4 | c++filt -t | grep "row_major::iamax(sycl::_V1::queue&, long, sycl::_V1::buffer<std::complex<double>"
  1302: 0000000002b573e0     9 FUNC    GLOBAL DEFAULT   11 oneapi::mkl::blas::row_major::iamax(sycl::_V1::queue&, long, sycl::_V1::buffer<std::complex<double>, 1, sycl::_V1::detail::aligned_allocator<std::complex<double> >, void>&, long, sycl::_V1::buffer<int, 1, sycl::_V1::detail::aligned_allocator<int>, void>&, oneapi::mkl::index_base)
  8510: 0000000002b573d0     9 FUNC    GLOBAL DEFAULT   11 oneapi::mkl::blas::row_major::iamax(sycl::_V1::queue&, long, sycl::_V1::buffer<std::complex<double>, 1, sycl::_V1::detail::aligned_allocator<std::complex<double> >, void>&, long, sycl::_V1::buffer<long, 1, sycl::_V1::detail::aligned_allocator<long>, void>&, oneapi::mkl::index_base)
```

We can use 2024.0 for the tests for now. Aidan is running more tests.

src/blas/backends/portblas/portblas_batch.cxx
s-Nick: portBLAS is cloned by oneMKL when building it and it always takes last commit in master. In any case, portBLAS does not have this  `gemm_batch` API yet

hjabird: Thanks Nicolò. In that case, perhaps having "for USM" is misleading here - it makes it sound like it'll work if only the buffer interface is used.

src/blas/backends/portblas/portblas_batch.cxx
```diff
@@ -213,6 +213,33 @@ void gemm_batch(sycl::queue &queue, oneapi::mkl::transpose transa, oneapi::mkl::
     throw unimplemented("blas", "gemm_batch", " for complex");
 }
 
+void gemm_batch(sycl::queue &queue, oneapi::mkl::transpose transa, oneapi::mkl::transpose transb,
+                std::int64_t m, std::int64_t n, std::int64_t k, float alpha,
+                sycl::buffer<sycl::half, 1> &a, std::int64_t lda, std::int64_t stride_a,
+                sycl::buffer<sycl::half, 1> &b, std::int64_t ldb, std::int64_t stride_b, float beta,
+                sycl::buffer<float, 1> &c, std::int64_t ldc, std::int64_t stride_c,
+                std::int64_t batch_size) {
+    throw unimplemented("blas", "gemm_batch", " for unsupporeted dtype");
```
Rbiessy: `unsupporeted` -> `unsupported` (here and below)

src/blas/backends/portblas/portblas_batch.cxx
Rbiessy: To be fair this is already done this way in many other places for the portBLAS backend. I don't mind either way.

mmeterel:
> > @AidanBeltonS Thanks for the PR. Before going through the review in more detail, what is your plan for this issue? Why openBLAS come into picture here? I would prefer to have all applicable backends working before adding these new APIs.
> > "I have been unable to test the mkl backends as I was running into some problems regarding duplicate definitions between the mkl headers and the openBlas/CBlas headers."
> 
> Hey @mmeterel, I checked with Aidan about the issue with the MKL backends. The duplicate definitions seemed to be an issue with the setup or build commands used. We ran into another issue with undefined references with `iamax` and `iamin` functions using buffers using 2024.1 oneAPI base toolkit. Just a few example:
> 
> ```
> /usr/bin/ld: lib/libonemkl_blas_mklcpu.so.0: undefined reference to `oneapi::mkl::blas::row_major::iamax(sycl::_V1::queue&, long, sycl::_V1::buffer<std::complex<double>, 1, sycl::_V1::detail::aligned_allocator<std::complex<double> >, void>&, long, sycl::_V1::buffer<long, 1, sycl::_V1::detail::aligned_allocator<long>, void>&)'
> /usr/bin/ld: lib/libonemkl_blas_mklcpu.so.0: undefined reference to `oneapi::mkl::blas::column_major::iamin(sycl::_V1::queue&, long, sycl::_V1::buffer<double, 1, sycl::_V1::detail::aligned_allocator<double>, void>&, long, sycl::_V1::buffer<long, 1, sycl::_V1::detail::aligned_allocator<long>, void>&)'
> /usr/bin/ld: lib/libonemkl_blas_mklcpu.so.0: undefined reference to `oneapi::mkl::blas::column_major::iamin(sycl::_V1::queue&, long, sycl::_V1::buffer<std::complex<double>, 1, sycl::_V1::detail::aligned_allocator<std::complex<double> >, void>&, long, sycl::_V1::buffer<long, 1, sycl::_V1::detail::aligned_allocator<long>, void>&)'
> /usr/bin/ld: lib/libonemkl_blas_mklcpu.so.0: undefined reference to `oneapi::mkl::blas::column_major::iamax(sycl::_V1::queue&, long, sycl::_V1::buffer<float, 1, sycl::_V1::detail::aligned_allocator<float>, void>&, long, sycl::_V1::buffer<long, 1, sycl::_V1::detail::aligned_allocator<long>, void>&)'
> ```
> 
> Looking at `libmkl_sycl_blas.so.4` in 2024.1 these functions expect an `index_base` as a last argument but is not there in oneMKL:
> 
> ```
> $ readelf -Wa /path/to/mkl/latest/lib/libmkl_sycl_blas.so.4 | c++filt -t | grep "row_major::iamax(sycl::_V1::queue&, long, sycl::_V1::buffer<std::complex<double>"
>   1302: 0000000002b573e0     9 FUNC    GLOBAL DEFAULT   11 oneapi::mkl::blas::row_major::iamax(sycl::_V1::queue&, long, sycl::_V1::buffer<std::complex<double>, 1, sycl::_V1::detail::aligned_allocator<std::complex<double> >, void>&, long, sycl::_V1::buffer<int, 1, sycl::_V1::detail::aligned_allocator<int>, void>&, oneapi::mkl::index_base)
>   8510: 0000000002b573d0     9 FUNC    GLOBAL DEFAULT   11 oneapi::mkl::blas::row_major::iamax(sycl::_V1::queue&, long, sycl::_V1::buffer<std::complex<double>, 1, sycl::_V1::detail::aligned_allocator<std::complex<double> >, void>&, long, sycl::_V1::buffer<long, 1, sycl::_V1::detail::aligned_allocator<long>, void>&, oneapi::mkl::index_base)
> ```
> 
> We can use 2024.0 for the tests for now. Aidan is running more tests.

@Rbiessy @AidanBeltonS AFAIK, there should not be any issues with missing symbols with 2024.1. This version has been in CI for a while now. I would suspect it can be a rebase issue on your branch. We should make it functional with 2024.1.

mmeterel:
@andrewtbarker Will you be able to help with this review?

andrewtbarker:
> @andrewtbarker Will you be able to help with this review?

Sure, I will take a look.

src/blas/backends/cublas/cublas_batch.cpp
```diff
@@ -140,16 +140,21 @@ void dgmm_batch(sycl::queue &queue, side left_right, int64_t m, int64_t n,
...
+    using cuTypeA = typename CudaEquivalentType<Ta>::Type;
+    using cuTypeB = typename CudaEquivalentType<Tb>::Type;
+    using cuTypeC = typename CudaEquivalentType<Tc>::Type;
+    using cuTypeS = typename CudaEquivalentType<Ts>::Type;
     overflow_check(m, n, k, lda, ldb, ldc, stride_a, stride_b, stride_c, batch_size);
+
+    cublasGemmAlgo_t cublas_gemm_algo = CUBLAS_GEMM_DEFAULT;
     queue.submit([&](sycl::handler &cgh) {
-        if (!verify_support<sycl::half, T>(queue, sycl::aspect::fp16)) {
+        if (!verify_support<sycl::half, Ta, Tb, Tc, Ts>(queue, sycl::aspect::fp16)) {
```
andrewtbarker, line 157: What role is `sycl::half` playing here? I'm not sure the `verify` or the error message in the `throw` actually matches the datatypes the user is asking for.

src/blas/backends/cublas/cublas_batch.cpp
```diff
@@ -553,17 +578,23 @@ sycl::event dgmm_batch(sycl::queue &queue, side *left_right, int64_t *m, int64_t
...
+    using cuTypeA = typename CudaEquivalentType<Ta>::Type;
+    using cuTypeB = typename CudaEquivalentType<Tb>::Type;
+    using cuTypeC = typename CudaEquivalentType<Tc>::Type;
+    using cuTypeS = typename CudaEquivalentType<Ts>::Type;
     overflow_check(m, n, k, lda, ldb, ldc, stride_a, stride_b, stride_c, batch_size);
+
+    cublasGemmAlgo_t cublas_gemm_algo = CUBLAS_GEMM_DEFAULT;
     auto done = queue.submit([&](sycl::handler &cgh) {
-        if (!verify_support<sycl::half, T>(queue, sycl::aspect::fp16)) {
+        if (!verify_support<sycl::half, Ta, Tb, Tc, Ts>(queue, sycl::aspect::fp16)) {
```
andrewtbarker, line 600: Again, why is `sycl::half` here and in the error message?

src/blas/backends/cublas/cublas_batch.cpp
```diff
@@ -573,50 +604,71 @@ inline sycl::event gemm_batch(const char *func_name, Func func, sycl::queue &que
...
+    using cuTypeC = typename CudaEquivalentType<Tc>::Type;
+    using cuTypeS = typename CudaEquivalentType<Ts>::Type;
     for (int64_t i = 0; i < group_count; i++) {
         overflow_check(m[i], n[i], k[i], lda[i], ldb[i], ldc[i], group_size[i]);
     }
+
+    cublasGemmAlgo_t cublas_gemm_algo = CUBLAS_GEMM_DEFAULT;
     auto done = queue.submit([&](sycl::handler &cgh) {
-        if (!verify_support<sycl::half, T>(queue, sycl::aspect::fp16)) {
+        if (!verify_support<sycl::half, Ta, Tb, Tc, Ts>(queue, sycl::aspect::fp16)) {
```
andrewtbarker, line 677: sycl::half?

src/blas/backends/cublas/cublas_batch.cpp
```diff
@@ -629,36 +681,53 @@ inline sycl::event gemm_batch(const char *func_name, Func func, sycl::queue &que
...
+
+#undef GEMM_BATCH_LAUNCHER_USM
+
+#define GEMM_BATCH_LAUNCHER_USM(TYPE_A, TYPE_B, TYPE_C, TYPE_S)                                    \
+    sycl::event gemm_batch(sycl::queue &queue, transpose *transa, transpose *transb, int64_t *m,   \
+                           int64_t *n, int64_t *k, TYPE_S *alpha, const TYPE_A **a, int64_t *lda,  \
+                           const TYPE_B **b, int64_t *ldb, TYPE_S *beta, TYPE_C **c, int64_t *ldc, \
+                           int64_t group_count, int64_t *group_size,                               \
+                           const std::vector<sycl::event> &dependencies) {                         \
+        throw unimplemented("blas", "gemm_batch", "for unimplmented dtypes");                      \
```
andrewtbarker: Can we add some kind of inline function to give the actual name(s) of the types? Something like

```c++
template <typename T>
inline const std::string& dtype_string();
template <>
inline const std::string& dtype_string<float>() {return "float";}
// ...
```

It would be a little painful but would provide nicer experience for the users, and could be used also for netlib backends etc.

src/blas/backends/rocblas/rocblas_batch.cpp
```diff
@@ -227,47 +227,76 @@ DGMM_STRIDED_BATCH_LAUNCHER(std::complex<double>, rocblas_zdgmm_strided_batched)
...
+
+#undef GEMM_STRIDED_BATCH_LAUNCHER
+
+#define GEMM_STRIDED_BATCH_LAUNCHER(TYPE_A, TYPE_B, TYPE_C, TYPE_S)                               \
+    void gemm_batch(sycl::queue &queue, transpose transa, transpose transb, int64_t m, int64_t n, \
+                    int64_t k, TYPE_S alpha, sycl::buffer<TYPE_A, 1> &a, int64_t lda,             \
+                    int64_t stridea, sycl::buffer<TYPE_B, 1> &b, int64_t ldb, int64_t strideb,    \
+                    TYPE_S beta, sycl::buffer<TYPE_C, 1> &c, int64_t ldc, int64_t stridec,        \
+                    int64_t batch_size) {                                                         \
+        throw unimplemented("blas", "gemm_batch", "for data type combination");                   \
```
andrewtbarker: As for cublas it would be nice to be able to provide actual type information here.

src/blas/backends/rocblas/rocblas_batch.cpp
```diff
@@ -816,63 +845,97 @@ DGMM_BATCH_LAUNCHER_USM(std::complex<double>, rocblas_zdgmm_batched)
...
-                              int64_t group_count, int64_t *group_size,
-                              const std::vector<sycl::event> &dependencies) {
-    using rocDataType = typename RocEquivalentType<T>::Type;
+#define GEMM_STRIDED_BATCH_LAUNCHER_USM(TYPE_A, TYPE_B, TYPE_C, TYPE_S)                        \
+    sycl::event gemm_batch(sycl::queue &queue, transpose transa, transpose transb, int64_t m,  \
+                           int64_t n, int64_t k, TYPE_S alpha, const TYPE_A *a, int64_t lda,   \
+                           int64_t stridea, const TYPE_B *b, int64_t ldb, int64_t strideb,     \
+                           TYPE_S beta, TYPE_C *c, int64_t ldc, int64_t stridec,               \
+                           int64_t batch_size, const std::vector<sycl::event> &dependencies) { \
+        throw unimplemented("blas", "gemm_batch", "for data type combination");                \
```
andrewtbarker: It would be nice to have actual type info here.

src/blas/function_table.hpp
```diff
@@ -869,6 +869,26 @@ typedef struct {
                                                   std::int64_t stride_b, sycl::half beta,
                                                   sycl::buffer<sycl::half, 1> &c, std::int64_t ldc,
                                                   std::int64_t stride_c, std::int64_t batch_size);
+    void (*column_major_hsgemm_batch_strided_sycl)(
```
andrewtbarker: I think the name should be `column_major_gemm_f16f16f32_batch_strided_sycl` to be consistent with the non-batch name [here](https://github.com/oneapi-src/oneMKL/blob/develop/src/blas/function_table.hpp#L646).

src/blas/function_table.hpp
```diff
@@ -869,6 +869,26 @@ typedef struct {
...
+        sycl::queue &queue, oneapi::mkl::transpose transa, oneapi::mkl::transpose transb,
+        std::int64_t m, std::int64_t n, std::int64_t k, float alpha, sycl::buffer<sycl::half, 1> &a,
+        std::int64_t lda, std::int64_t stride_a, sycl::buffer<sycl::half, 1> &b, std::int64_t ldb,
+        std::int64_t stride_b, float beta, sycl::buffer<float, 1> &c, std::int64_t ldc,
+        std::int64_t stride_c, std::int64_t batch_size);
+    void (*column_major_isgemm_batch_strided_sycl)(
```
andrewtbarker:
```suggestion
    void (*column_major_gemm_s8s8f32_batch_strided_sycl)(
```

src/blas/function_table.hpp
```diff
@@ -869,6 +869,26 @@ typedef struct {
...
+        sycl::queue &queue, oneapi::mkl::transpose transa, oneapi::mkl::transpose transb,
+        std::int64_t m, std::int64_t n, std::int64_t k, float alpha,
+        sycl::buffer<std::int8_t, 1> &a, std::int64_t lda, std::int64_t stride_a,
+        sycl::buffer<std::int8_t, 1> &b, std::int64_t ldb, std::int64_t stride_b, float beta,
+        sycl::buffer<float, 1> &c, std::int64_t ldc, std::int64_t stride_c,
+        std::int64_t batch_size);
+    void (*column_major_iigemm_batch_strided_sycl)(
```
andrewtbarker:
```suggestion
    void (*column_major_gemm_s8s8s32_batch_strided_sycl)(
```
consistent with [this](https://github.com/oneapi-src/oneMKL/blob/develop/src/blas/function_table.hpp#L928)

src/blas/function_table.hpp
```diff
@@ -2180,6 +2200,24 @@ typedef struct {
         std::int64_t *lda, const sycl::half **b, std::int64_t *ldb, sycl::half *beta,
         sycl::half **c, std::int64_t *ldc, std::int64_t group_count, std::int64_t *group_size,
         const std::vector<sycl::event> &dependencies);
+    sycl::event (*column_major_hsgemm_batch_group_usm_sycl)(
```
andrewtbarker:
```suggestion
    sycl::event (*column_major_gemm_f16f16f32_batch_group_usm_sycl)(
```

src/blas/function_table.hpp
```diff
@@ -2180,6 +2200,24 @@ typedef struct {
...
+        sycl::queue &queue, oneapi::mkl::transpose *transa, oneapi::mkl::transpose *transb,
+        std::int64_t *m, std::int64_t *n, std::int64_t *k, float *alpha, const sycl::half **a,
+        std::int64_t *lda, const sycl::half **b, std::int64_t *ldb, float *beta, float **c,
+        std::int64_t *ldc, std::int64_t group_count, std::int64_t *group_size,
+        const std::vector<sycl::event> &dependencies);
+    sycl::event (*column_major_isgemm_batch_group_usm_sycl)(
```
andrewtbarker:
```suggestion
    sycl::event (*column_major_gemm_s8s8f32_batch_group_usm_sycl)(
```

src/blas/function_table.hpp
```diff
@@ -2180,6 +2200,24 @@ typedef struct {
...
+        sycl::queue &queue, oneapi::mkl::transpose *transa, oneapi::mkl::transpose *transb,
+        std::int64_t *m, std::int64_t *n, std::int64_t *k, float *alpha, const std::int8_t **a,
+        std::int64_t *lda, const std::int8_t **b, std::int64_t *ldb, float *beta, float **c,
+        std::int64_t *ldc, std::int64_t group_count, std::int64_t *group_size,
+        const std::vector<sycl::event> &dependencies);
+    sycl::event (*column_major_iigemm_batch_group_usm_sycl)(
```
andrewtbarker:
```suggestion
    sycl::event (*column_major_gemm_s8s8s32_batch_group_usm_sycl)(
```

src/blas/function_table.hpp
```diff
@@ -2213,6 +2251,24 @@ typedef struct {
         std::int64_t stride_b, sycl::half beta, sycl::half *c, std::int64_t ldc,
         std::int64_t stride_c, std::int64_t batch_size,
         const std::vector<sycl::event> &dependencies);
+    sycl::event (*column_major_hsgemm_batch_strided_usm_sycl)(
```
andrewtbarker: Same comments throughout this file, for row-major etc.

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
```diff
@@ -63,10 +63,10 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t batch_size) {
     lda = std::max(m, k);
     ldb = std::max(n, k);
     ldc = std::max(m, n);
-    alpha = rand_scalar<fp>();
-    beta = rand_scalar<fp>();
+    alpha = rand_scalar<Ts>();
+    beta = rand_scalar<Ts>();
 
-    if ((std::is_same<fp, float>::value) || (std::is_same<fp, double>::value)) {
+    if ((std::is_same<Ts, float>::value) || (std::is_same<Ts, double>::value)) {
```
andrewtbarker: I don't think this is doing what you want. If `Ts` is half, for example, you may get conjtrans in line 76. Probably it makes sense to reverse the sense of the if and check for `complex<float>` or `complex<double>` in the if and everything else in the else.

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
```diff
@@ -200,11 +211,14 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t batch_size) {
     }
 
     // Compare the results of reference implementation and DPC++ implementation.
+    constexpr int tol_scalar = std::is_same_v<Ta, Ts> ? 10 : 40;
```
andrewtbarker: When exactly do you need this lower tolerance? I would have expected `is_same_v<Tc, Ts>` because internal accumulation is done in `Tc` precision but I may be wrong.

tests/unit_tests/blas/batch/gemm_batch_stride_usm.cpp
```diff
@@ -83,9 +83,9 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t batch_size) {
     lda = std::max(m, k);
     ldb = std::max(n, k);
     ldc = std::max(m, n);
-    alpha = rand_scalar<fp>();
-    beta = rand_scalar<fp>();
-    if ((std::is_same<fp, float>::value) || (std::is_same<fp, double>::value)) {
+    alpha = rand_scalar<Ts>();
+    beta = rand_scalar<Ts>();
+    if ((std::is_same<Ts, float>::value) || (std::is_same<Ts, double>::value)) {
```
andrewtbarker: Same comment as in buffer test, this if should be modified / reversed.

tests/unit_tests/blas/batch/gemm_batch_usm.cpp
```diff
@@ -104,9 +104,9 @@ int test(device *dev, oneapi::mkl::layout layout, int64_t group_count) {
         lda[i] = std::max(m[i], k[i]);
         ldb[i] = std::max(n[i], k[i]);
         ldc[i] = std::max(m[i], n[i]);
-        alpha[i] = rand_scalar<fp>();
-        beta[i] = rand_scalar<fp>();
-        if ((std::is_same<fp, float>::value) || (std::is_same<fp, double>::value)) {
+        alpha[i] = rand_scalar<Ts>();
+        beta[i] = rand_scalar<Ts>();
+        if ((std::is_same<Ts, float>::value) || (std::is_same<Ts, double>::value)) {
```
andrewtbarker: Same comment, do not allow `conjtrans` for (real) half, int8, int32, etc.

andrewtbarker (commented):
Thanks for the PR, there is a lot of good work here. Most of my comments are just about style and naming consistency.

andrewtbarker:
> @Rbiessy @AidanBeltonS AFAIK, there should not be any issues with missing symbols with 2024.1. This version has been in CI for a while now. I would suspect it can be a rebase issue on your branch. We should make it functional with 2024.1.

Yes, this should have been fixed in #445 . If not we should fix it.

mmeterel:
Have you tested the PR with hipSYCL/AdaptiveSYCL? Can you please add the logs?

src/blas/backends/rocblas/rocblas_helper.hpp
hjabird: Short answer: just the way the code was written for current requirements.

src/blas/backends/cublas/cublas_batch.cpp
AidanBeltonS: The half is treated differently to the subsequent dtypes. It is considered the `verify_type` which is compared to the list of typenames passed after it.
If the aspect fails and the verify_type is equal to any of the other dtypes then it returns failse.
This is the behaviour we want, as if any of the types in the gemm are a half we want to fail if we do not support halfs.
Impl: https://github.com/oneapi-src/oneMKL/blob/develop/src/include/runtime_support_helper.hpp#L33

So in this case if any of the data types that used in the gemm are a half, then the aspect is queried and fails as a result. I think the error message does still convey the problem to the use. I.e. the device does not support fp16.

src/blas/backends/cublas/cublas_batch.cpp
AidanBeltonS: See above

src/blas/backends/cublas/cublas_batch.cpp
AidanBeltonS: See above

src/blas/backends/portblas/portblas_batch.cxx
AidanBeltonS: I think this message could be changed but this PR may not be the place to do it, as it should be changed in all the other places

src/blas/backends/cublas/cublas_helper.hpp
AidanBeltonS: Fixed

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
AidanBeltonS: Iv placed `{}` where I think the style guide recommends, let me know if you see a place iv missed

src/blas/backends/portblas/portblas_batch.cxx
AidanBeltonS: Done

src/blas/backends/cublas/cublas_batch.cpp
AidanBeltonS: I have added this feature

src/blas/backends/rocblas/rocblas_batch.cpp
AidanBeltonS: See above

src/blas/backends/rocblas/rocblas_batch.cpp
AidanBeltonS: See above

src/blas/function_table.hpp
AidanBeltonS: Done

src/blas/function_table.hpp
AidanBeltonS: Done

src/blas/function_table.hpp
AidanBeltonS: Done

src/blas/function_table.hpp
AidanBeltonS: Done

src/blas/function_table.hpp
AidanBeltonS: Done

src/blas/function_table.hpp
AidanBeltonS: Done

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
AidanBeltonS: I see your point. I have flipped the check and applied it to all the tests

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
AidanBeltonS: Although the computation is happening in fp32 according the compute type, I am still seeing differences in the output. They are small, I suspect it could be down the implementation and hardware differences. It is a larger discussion about how precise these computations must be to the reference, which is doing something slightly different due to the difference in input type.

AidanBeltonS:
No I have not tested HIPsycl. I have attached the other backend tests below. Netlib and portblas are passing fine. MKL has some failing tests due to tolerating which I am investigating further. It seems it deviates more from the reference implementation in some cases.
[mkl_test_log.txt](https://github.com/oneapi-src/oneMKL/files/14907941/mkl_test_log.txt)
[netlib_test_log.txt](https://github.com/oneapi-src/oneMKL/files/14907946/netlib_test_log.txt)
[port_blas_test_logs.txt](https://github.com/oneapi-src/oneMKL/files/14907949/port_blas_test_logs.txt)

MKL tests error:
[mkl_test_log.txt](https://github.com/oneapi-src/oneMKL/files/14907982/mkl_test_log.txt)

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
andrewtbarker: Agreed, this tolerance is fine for this PR.

src/blas/backends/cublas/cublas_batch.cpp
andrewtbarker: Thanks for the explanation, I was confused about `verify_support`, this makes sense now.

andrewtbarker:
> MKL has some failing tests due to tolerating which I am investigating further.

It looks like `dotc` and `dotu` have segfaults in your tests. Initially I think this is unlikely to be due to your PR but have you looked at this at all?

AidanBeltonS:
> > MKL has some failing tests due to tolerating which I am investigating further.
> 
> It looks like `dotc` and `dotu` have segfaults in your tests. Initially I think this is unlikely to be due to your PR but have you looked at this at all?

The failures it Dot are due to error
```
[ RUN      ] DotTestSuite/DotTests.RealDoubleSinglePrecision/Row_Major_Intel_R__Data_Center_GPU_Max_1100
relative error = 1.83849e-08 absolute error = 1.24863e-07 limit = 3.01315e-13
Difference in result: DPC++ 6.79159 vs. Reference 6.79159
/home/aidanbelton/source/oneMKL/tests/unit_tests/blas/level1/dot.cpp:157: Failure
Expected equality of these values:
  res
    Which is: 0
  1
[  FAILED  ] DotTestSuite/DotTests.RealDoubleSinglePrecision/Row_Major_Intel_R__Data_Center_GPU_Max_1100, where GetParam() = (0x560f5e0, 1-byte object <00>) (1 ms)
```

DotU is an odd one, it does not appear to be related to my changes however
```
[ RUN      ] DotuTestSuite/DotuTests.ComplexSinglePrecision/Row_Major_Intel_R__Data_Center_GPU_Max_1100
Caught synchronous SYCL exception during DOTU:
The program was built for 1 devices
Build program log for 'Intel(R) Data Center GPU Max 1100':
 -11 (PI_ERROR_BUILD_PROGRAM_FAILURE) -11 (PI_ERROR_BUILD_PROGRAM_FAILURE)
OpenCL status: sycl:7
unknown file: Failure
C++ exception with description "Enqueue process failed. -59 (PI_ERROR_INVALID_OPERATION)" thrown in the test body.
[  FAILED  ] DotuTestSuite/DotuTests.ComplexSinglePrecision/Row_Major_Intel_R__Data_Center_GPU_Max_1100, where GetParam() = (0x560f5e0, 1-byte object <00>) (0 ms)
```

AidanBeltonS:
I have resolved all but one issue with GemmBatch's tests. The CPU MKL implementation has significant amounts of error compared to the GPU. I believe there may be a fundamental difference in the precision of the calculation for the CPU. One possible fix would be to increase the tolerance significantly just for the CPU. Im not a fan of this approach as it is a bit of a brute force solution. Does anyone have any recommendations on how they would like to see this handled?
```
[ RUN      ] GemmBatchUsmTestSuite/GemmBatchUsmTests.RealIntRealScalarPrecision/Column_Major_Intel_R__Xeon_R__Gold_5418Y
relative error = 0.000911658 absolute error = 0.00168478 limit = 0.000333786
Difference in entry (58,119): DPC++ 1.84973 vs. Reference 1.84804
relative error = 0.000812301 absolute error = 0.00121021 limit = 0.000333786
Difference in entry (0,124): DPC++ 1.49107 vs. Reference 1.48986
relative error = 0.000534697 absolute error = 0.000857353 limit = 0.000333786
Difference in entry (17,144): DPC++ 1.60258 vs. Reference 1.60344
relative error = 0.000527185 absolute error = 0.00049144 limit = 0.000333786
Difference in entry (52,186): DPC++ -0.932689 vs. Reference -0.932197
/home/aidanbelton/source/oneMKL/tests/unit_tests/blas/batch/gemm_batch_usm.cpp:408: Failure
Expected equality of these values:
  res
    Which is: 0
  1
[  FAILED  ] GemmBatchUsmTestSuite/GemmBatchUsmTests.RealIntRealScalarPrecision/Column_Major_Intel_R__Xeon_R__Gold_5418Y, where GetParam() = (0x56845d0, 1-byte object <01>) (331 ms)
```

mmeterel:
> No I have not tested HIPsycl. I have attached the other backend tests below. Netlib and portblas are passing fine. MKL has some failing tests due to tolerating which I am investigating further. It seems it deviates more from the reference implementation in some cases. [mkl_test_log.txt](https://github.com/oneapi-src/oneMKL/files/14907941/mkl_test_log.txt) [netlib_test_log.txt](https://github.com/oneapi-src/oneMKL/files/14907946/netlib_test_log.txt) [port_blas_test_logs.txt](https://github.com/oneapi-src/oneMKL/files/14907949/port_blas_test_logs.txt)
> 
> MKL tests error: [mkl_test_log.txt](https://github.com/oneapi-src/oneMKL/files/14907982/mkl_test_log.txt)

Can you please test hipSYCL backend as well?

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
```diff
@@ -213,29 +282,49 @@ class GemmBatchStrideTests
         : public ::testing::TestWithParam<std::tuple<sycl::device *, oneapi::mkl::layout>> {};
 
 TEST_P(GemmBatchStrideTests, RealHalfPrecision) {
-    EXPECT_TRUEORSKIP(test<sycl::half>(std::get<0>(GetParam()), std::get<1>(GetParam()), 5));
+    EXPECT_TRUEORSKIP((test<sycl::half, sycl::half, sycl::half, sycl::half>(
+        std::get<0>(GetParam()), std::get<1>(GetParam()), 5)));
+}
+
+TEST_P(GemmBatchStrideTests, RealHalfRealScalarPrecision) {
```
andrewtbarker:
```suggestion
TEST_P(GemmBatchStrideTests, HalfHalfFloatPrecision) {
```
To be consistent with the test names in `gemm_usm.cpp`. More generally I think these test names should be more specific (see also `../extensions/gemm_bias.cpp`) and `Scalar` is not very helpful.

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
```diff
@@ -213,29 +282,49 @@ class GemmBatchStrideTests
...
+    EXPECT_TRUEORSKIP((test<sycl::half, sycl::half, float, float>(std::get<0>(GetParam()),
+                                                                  std::get<1>(GetParam()), 5)));
+}
+
+TEST_P(GemmBatchStrideTests, RealIntRealScalarPrecision) {
```
andrewtbarker:
```suggestion
TEST_P(GemmBatchStrideTests, Int8Int8SinglePrecision) {
```
more like what is in `gemm_bias` tests. (Single could also be Float32, not sure what is best).

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
```diff
@@ -213,29 +282,49 @@ class GemmBatchStrideTests
...
+    EXPECT_TRUEORSKIP((test<std::int8_t, std::int8_t, float, float>(std::get<0>(GetParam()),
+                                                                    std::get<1>(GetParam()), 5)));
+}
+
+TEST_P(GemmBatchStrideTests, RealIntPrecision) {
```
andrewtbarker: `Int8Int8Int32Precision`

andrewtbarker (commented):
Some suggestions to make interpreting failed test results easier - I flagged a few places but there are similar issues in most of the new tests.

andrewtbarker:
> Does anyone have any recommendations on how they would like to see this handled?

If, as we suspect, the CPU backend is doing accumulation in double while the GPU backend does it in float, one option would be changing what reference gemm from `tests/unit_tests/blas/include/reference_blas_templates.hpp` we call (might need to add a reference gemm in that file).

andrewtbarker:
What is the status here? As I see it we have three outstanding items:
1. AdaptiveCpp testing
2. Test names (my most recent review, minor)
3. Failure in `RealIntRealScalarPrecision`

(1) may be a larger issue with CI that in my opinion can be dealt with separately in another PR. (2) is minor and should be easy to fix. I hope (3) is also minor but I'm not sure, is there any progress understanding it?

Rbiessy:
Hi @andrewtbarker, I have updated the status by email as it was easier to discuss issues with testing AdaptiveCpp on the CI. In short there are a few issues @AidanBeltonS will need to look at once he is back from Holiday next week!

AidanBeltonS:
> What is the status here? As I see it we have three outstanding items:
> 
>     1. AdaptiveCpp testing
> 
>     2. Test names (my most recent review, minor)
> 
>     3. Failure in `RealIntRealScalarPrecision`
> 
> 
> (1) may be a larger issue with CI that in my opinion can be dealt with separately in another PR. (2) is minor and should be easy to fix. I hope (3) is also minor but I'm not sure, is there any progress understanding it?

I have addressed items 2. and 3. 
To resolve 3 I am scaling the tolerance by the possible input range from int8 matricies. i.e. 256
I have yet to test this with AdaptiveCpp, Ill start looking at that shortly

src/include/error_helper.hpp
```diff
@@ -0,0 +1,46 @@
...
+*******************************************************************************/
+
+#ifndef _ONEMKL_ERROR_HELPER_HPP_
+#define _ONEMKL_ERROR_HELPER_HPP_
+
+#if __has_include(<sycl/sycl.hpp>)
+#include <sycl/sycl.hpp>
+#else
+#include <CL/sycl.hpp>
+#endif
```
Rbiessy: This doesn't need SYCL, including `<string>` is enough.
Also I think `dtype_string.hpp` would be a better name for the file.

andrewtbarker (approved):
I think we're good to go. Thanks for sticking with this one!

src/include/error_helper.hpp
AidanBeltonS: I have renamed the file and removed the SYCL dependency

Rbiessy (dismissed):

AidanBeltonS:
I have disabled the (int8, int8, float, float) dtype combinations on the CuBlas backend due to failing tests on AdaptiveCPP. This is because I do not currently have bandwidth to solve this problem. The root goal of this PR was to get the (half, half, float, float) combination for llama.cpp, and that is working.

I think it would be easier for another person to pick up this work in the future, as the interface, testing, and structure that is needed already exists.

andrewtbarker:
> I think it would be easier for another person to pick up this work in the future, as the interface, testing, and structure that is needed already exists.

Can you file an issue documenting this so we can track it?

AidanBeltonS:
> > I think it would be easier for another person to pick up this work in the future, as the interface, testing, and structure that is needed already exists.
> 
> Can you file an issue documenting this so we can track it?

Sure, I have created #506

AidanBeltonS:
I have also disabled the in8, float combination for MKLCPU/GPU as I found similar precision issues. 
#506 will be updated to reflect this

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
```diff
@@ -48,12 +48,65 @@ extern std::vector<sycl::device *> devices;
...
+
+    return good;
+}
+
+template <typename T>
+bool check_mat(acc_type_t<T> &M, vec_type_t<T> &M_ref, oneapi::mkl::layout layout, int m, int n,
+               int ld, int error_mag, std::ostream &out) {
+    if constexpr (std::is_same<T, int32_t>::value)
+        return check_equal_matrix_int(M, M_ref, layout, m, n, ld, error_mag, out);
+    return check_equal_matrix<acc_type_t<T>>(M, M_ref, layout, m, n, ld, error_mag, out);
```
Rbiessy: Thanks, I'm running the CI to make sure this is not giving us any warning. It's worth adding a comment to explain the issues with specializing `check_equal_matrix`.

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
```diff
@@ -48,12 +48,65 @@ extern std::vector<sycl::device *> devices;
...
+template <class T>
+struct vec_type {
+    typedef vector<T, allocator_helper<T, 64>> type;
+};
+
+template <class T>
+using vec_type_t = typename vec_type<T>::type;
+
+// Check for int32_t and Ts=float as small differences in the reference become large after rounding
+bool check_equal_matrix_int(acc_type_t<int32_t> &M, vec_type_t<int32_t> &M_ref,
```
Rbiessy: This seems to be duplicated on 3 files. Is it possible to move `check_equal_matrix_int`, `check_mat` and the helper types in `test_common.hpp`?
I think just `check_equal_matrix_int` will need to be `inline` to make sure it is not duplicated in multiple translation units.

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
```diff
@@ -48,12 +48,65 @@ extern std::vector<sycl::device *> devices;
...
+    return (std::abs(x - x_ref) <= 1);
+}
+
+template <class T>
+struct acc_type {
+    typedef host_accessor<T, 1, access_mode::read> type;
+};
+
+template <class T>
+using acc_type_t = typename acc_type<T>::type;
```
Rbiessy: Is there a benefit of the `acc_type` struct over the suggestion below?
```
using acc_type_t = host_accessor<T, 1, access_mode::read>;
```
Same comment for `vec_type` below.

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
AidanBeltonS: This has been refactored so the comment would not make much sense anymore

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
AidanBeltonS: I have moved this into `test_common.hpp` under `check_almost_equal_matrix`

tests/unit_tests/blas/batch/gemm_batch_stride.cpp
AidanBeltonS: vec and acc types are no longer used after refactor

tests/unit_tests/blas/include/test_common.hpp
```diff
@@ -655,4 +655,57 @@ bool check_equal_trsm_matrix(acc1 &M, acc2 &M_ref, oneapi::mkl::layout layout, i
...
+// Helper for checking if a matrix/vector/accessor structure returns an integral type
+template <typename T>
+constexpr bool is_matrix_type_integral() {
+    return std::is_integral_v<typename std::result_of<access_index<T>(int)>::type>;
+}
+
+template <typename fp>
+typename std::enable_if<std::is_integral<fp>::value, bool>::type check_almost_equal_int(
+    fp x, fp x_ref, int error_mag) {
+    return (std::abs(x - x_ref) <= 1);
```
Rbiessy:
```suggestion
    return (std::abs(x - x_ref) <= error_mag);
```

AidanBeltonS: Done

Rbiessy (approved):

Rbiessy:
I have confirmed the tests pass with AdaptiveCpp on AMD and Nvidia HW.

PR #468, opened 2024-04-04
[Sparse BLAS] remove deprecation warning when using oneMKL 2024.1 with sparse::trsv() on x86 CPU and Intel GPU backends

spencerpatty:
# Description

oneMKL 2024.1 deprecated sparse::trsv() without an alpha scaling factor and introduced a new sparse::trsv() with alpha. We update trsv backend calls to use the new one from oneMKL 2024.1 with constant fixed alpha == 1 to avoid the deprecation warnings.

# Checklist

## All Submissions

## Bug fixes

- [x] Have you added relevant regression tests?  (N/A)
      GitHub issue or in this PR)?  (N/A)

### DIFF
--- a/src/sparse_blas/backends/mkl_common/mkl_helper.hpp
+++ b/src/sparse_blas/backends/mkl_common/mkl_helper.hpp
@@ -20,6 +20,7 @@
 // MKLCPU and MKLGPU backends include
 // This include defines its own oneapi::mkl::sparse namespace with some of the types that are used here: matrix_handle_t, index_base, transpose, uolo, diag.
 #include <oneapi/mkl/spblas.hpp>
+#include "mkl_version.h"
 
 // Includes are set up so that oneapi::mkl::sparse namespace refers to the MKLCPU and MKLGPU backends namespace (oneMKL product)
 // in this file.
--- a/src/sparse_blas/backends/mkl_common/mkl_helper.hpp
+++ b/src/sparse_blas/backends/mkl_common/mkl_helper.hpp
@@ -20,7 +20,6 @@
 // MKLCPU and MKLGPU backends include
 // This include defines its own oneapi::mkl::sparse namespace with some of the types that are used here: matrix_handle_t, index_base, transpose, uolo, diag.
 #include <oneapi/mkl/spblas.hpp>
-#include "mkl_version.h"
 
 // Includes are set up so that oneapi::mkl::sparse namespace refers to the MKLCPU and MKLGPU backends namespace (oneMKL product)
 // in this file.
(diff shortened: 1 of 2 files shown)


Rbiessy (approved):
Thanks for the PR. I am wondering for how long do we need to support previous MKL releases?
Previous releases are not tested AFAIK so I am not sure it is worth complexifying the code with macros.

spencerpatty:
> Thanks for the PR. I am wondering for how long do we need to support previous MKL releases? Previous releases are not tested AFAIK so I am not sure it is worth complexifying the code with macros.

To me, it makes sense to support the latest MKL minor release, but also the latest MKL Major release as well, but seeing as we always use the latest compiler, maybe it also just makes sense to support the latest MKL release and not distinguish between minor and major releases ?  I also agree that having macros to complexify the code is not so good, luckily here it is a simple change, but it could be much worse ...  If we agree to not worry about minor/major, I can quickly just drop the else part of this fix.

Rbiessy:
I would avoid the macro in this case. They don't seem to be used in the project so far.

src/sparse_blas/backends/mkl_common/mkl_helper.hpp
```diff
@@ -20,6 +20,7 @@
 // MKLCPU and MKLGPU backends include
 // This include defines its own oneapi::mkl::sparse namespace with some of the types that are used here: matrix_handle_t, index_base, transpose, uolo, diag.
 #include <oneapi/mkl/spblas.hpp>
+#include "mkl_version.h"
```
spencerpatty: this doesn't hurt, but isn't strictly necessary now that we have removed the MACROs.  OK to keep ?

mmeterel:
> > Thanks for the PR. I am wondering for how long do we need to support previous MKL releases? Previous releases are not tested AFAIK so I am not sure it is worth complexifying the code with macros.
> 
> To me, it makes sense to support the latest MKL minor release, but also the latest MKL Major release as well, but seeing as we always use the latest compiler, maybe it also just makes sense to support the latest MKL release and not distinguish between minor and major releases ? I also agree that having macros to complexify the code is not so good, luckily here it is a simple change, but it could be much worse ... If we agree to not worry about minor/major, I can quickly just drop the else part of this fix.

Typically, we claim support for only `latest` oneMKL product - which is in the `latest` base tool kit (not distinguished between minor/major releases).

src/sparse_blas/backends/mkl_common/mkl_helper.hpp
Rbiessy: It's best to remove.
```suggestion

```

Rbiessy (approved):

Rbiessy:
Note this is also addressed by https://github.com/oneapi-src/oneMKL/pull/500

gajanan-choudhary:
Closing this PR as it is no longer relevant given #500 now merged in to match the overhauled spec APIs in https://github.com/uxlfoundation/oneAPI-spec/pull/522

PR #469, opened 2024-04-09
[README] Remove Open DPC++ compiler from oneMKL CPU supported configurations

dnhsieh-intel:
# Description

This PR removes oneAPI DPC++ Compiler (Open DPC++) from oneMKL CPU supported configurations.

Since the timeline of Open DPC++ is faster than oneMKL releases, the latest Open DPC++ is currently incompatible with the latest oneMKL release, and it is expected that the incompatibility will happen every year starting from breaking changes of Open DPC++ around April until the next oneMKL major release around the end of year. It is unreasonable to freeze the Open DPC++ supported version for such a long time. Although we might use a compatible version for oneMKL and the latest version for other libraries, there is an overhead to track a compatible (and usually old) version with probably little value added. Hence the proposal is to remove Open DPC++ from oneMKL CPU supported configurations. Note that this PR does not update the CMake files so that users can still use Open DPC++ with oneMKL CPU backend if they choose to.

# Checklist

## All Submissions

- Do all unit tests pass locally? N/A, only updating README.
- Have you formatted the code using clang-format? N/A, only updating README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ Supported compilers include:
             <td rowspan=9 align="center">BLAS</td>
             <td rowspan=3 align="center">x86 CPU</td>
             <td align="center">Intel(R) oneMKL</td>
-            <td align="center">Intel DPC++</br>Open DPC++</br>AdaptiveCpp</td>
+            <td align="center">Intel DPC++</br>AdaptiveCpp</td>
             <td align="center">Dynamic, Static</td>
         </tr>
         <tr>
@@ -225,7 +225,7 @@ Supported compilers include:
             <td rowspan=4 align="center">LAPACK</td>
             <td align="center">x86 CPU</td>
             <td align="center">Intel(R) oneMKL</td>
-            <td align="center">Intel DPC++</br>Open DPC++</td>
+            <td align="center">Intel DPC++</td>
             <td align="center">Dynamic, Static</td>
         </tr>
         <tr>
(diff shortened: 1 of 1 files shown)


dnhsieh-intel:
@hjabird Please let us know if Codeplay has concerns about this change. Thanks!

mmeterel (approved):
Thanks for the PR. Could you please open a ticket to infra to remove this config from CI (if you haven't already)?

dnhsieh-intel:
>Could you please open a ticket to infra to remove this config from CI (if you haven't already)?

I have a draft PR to remove it from CI and will open it once we have confirmation from Codeplay. But we still need to bother infra to resume testing the latest version... I will update the ticket originally for 2025.0.

Rbiessy (approved):
I don't have any concerns with removing these mentions. This is a step in the right direction I would say. Thanks.

hjabird (approved):

hjabird:
LGTM!

Any idea what broke exactly? I'd have thought MKLCPU uses just the C interface, so changes in the open-source DPC++ wouldn't break it.

dnhsieh-intel:
>Any idea what broke exactly? I'd have thought MKLCPU uses just the C interface, so changes in the open-source DPC++ wouldn't break it.

The latest oneMKL depends on `libsycl.so.7`, but the open-source DPC++ has already moved to `libsycl.so.8`.

PR #470, opened 2024-04-09
[RNG] Fixed some bugs for device API

andreyfe1:
# Description

It doesn't make sense to keep constructors with `std::initializer_list` for `mcg31m1` and `mcg59` random generators since one seed value is enough. Size of the state for `mcg31m1` is 31 bits, `uint32_t` is enough to store the initial state. For `mcg59` size is 59 bits, `uint64_t` is enough. So, let's keep only `std::uint32_t seed` for `mcg31m1` and `std::uint64_t seed` for `mcg59`.

PR for the Spec is here: https://github.com/oneapi-src/oneAPI-spec/pull/525.

# Checklist

- [x] Do all unit tests pass locally? Attach a log. [log.txt](https://github.com/oneapi-src/oneMKL/files/14919838/log.txt)
- [x] What version of oneAPI spec the interface is targeted?
        See https://github.com/oneapi-src/oneAPI-spec/pull/525

### DIFF
--- a/examples/rng/device/uniform.cpp
+++ b/examples/rng/device/uniform.cpp
@@ -46,7 +46,7 @@ bool isDoubleSupported(sycl::device my_dev) {
 }
 
 // example parameters
-constexpr int seed = 777;
+constexpr std::uint64_t seed = 777;
 constexpr std::size_t n = 1024;
 constexpr int n_print = 10;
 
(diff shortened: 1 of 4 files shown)


aelizaro (approved):
Looks good to me, thank you!

egrabovskaya (approved):

PR #474, opened 2024-04-18
[BLAS::portBLAS backend] Add CUDA_TARGET

pgorlani:
# Description

This patch adds the `CUDA_TARGET` CMake variable for setting a specific device architecture in case of `NVIDIA_GPU` tuning target.

This avoids the need of device code translation performed at runtime from the DPCPP default CUDA target architecture (`sm_50`) to the one of the actual device.     

# Checklist

## All Submissions

- [*] Do all unit tests pass locally? Attached [log](https://github.com/oneapi-src/oneMKL/files/15024376/ctest-add-cuda-target.txt).
- [*] Have you formatted the code using clang-format?

### DIFF
--- a/src/blas/backends/portblas/CMakeLists.txt
+++ b/src/blas/backends/portblas/CMakeLists.txt
@@ -78,6 +78,12 @@ if(PORTBLAS_TUNING_TARGET)
         -fsycl-targets=nvptx64-nvidia-cuda -fsycl-unnamed-lambda)
       target_link_options(ONEMKL::SYCL::SYCL INTERFACE
         -fsycl-targets=nvptx64-nvidia-cuda)
+      if(DEFINED CUDA_TARGET)
+        target_compile_options(ONEMKL::SYCL::SYCL INTERFACE
+          -Xsycl-target-backend --cuda-gpu-arch=${CUDA_TARGET})
+        target_link_options(ONEMKL::SYCL::SYCL INTERFACE
+          -Xsycl-target-backend --cuda-gpu-arch=${CUDA_TARGET})
+      endif()
     else()
       message(WARNING "Compiler is not supported."
       " Unable to automatically set the required flags for the target '${PORTBLAS_TUNING_TARGET}'."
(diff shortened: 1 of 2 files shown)


Rbiessy (approved):
Thanks Paolo. That looks good to me.

I was wondering if we should define and use `CUDA_TARGET` in the same places that we use `HIP_TARGETS` but given that `CUDA_TARGET` is optional and seems to be only useful to portBLAS so far I think it makes sense. What do you think @mkrainiuk ?

src/blas/backends/portblas/CMakeLists.txt
```diff
@@ -78,6 +78,12 @@ if(PORTBLAS_TUNING_TARGET)
         -fsycl-targets=nvptx64-nvidia-cuda -fsycl-unnamed-lambda)
       target_link_options(ONEMKL::SYCL::SYCL INTERFACE
         -fsycl-targets=nvptx64-nvidia-cuda)
+      if(DEFINED CUDA_TARGET)
```
mkrainiuk, line 81: Does it make sense to verify the value of `CUDA_TARGET` or it's better to rely on compiler to generate compile time error in case of unrecognized/unsupported target?

mkrainiuk (commented):
> I was wondering if we should define and use CUDA_TARGET in the same places that we use HIP_TARGETS but given that CUDA_TARGET is optional and seems to be only useful to portBLAS so far I think it makes sense. What do you think @mkrainiuk ?

Agree, we can move `CUDA_TARGET`(S?) to the same place later when it will be used by more than one backend.
For now I think we need a note in the documentation that it's supported only for portBLAS.

src/blas/backends/portblas/CMakeLists.txt
pgorlani: Thanks for your review, @mkrainiuk.

Setting `-DCUDA_TARGET=sm_99` produces
```
icpx: error: unsupported CUDA gpu architecture: sm_99
```
I think that the error emitted by the compiler is quite clear. Moreover, the supported architectures depend also on the CUDA toolkit version.

Rbiessy:
> For now I think we need a note in the documentation that it's supported only for portBLAS.

@mkrainiuk The documentation that Paolo added in `docs/building_the_project.rst` is already in the portBLAS section. Are you suggesting it should be more even more explicit?

mkrainiuk:
> @mkrainiuk The documentation that Paolo added in `docs/building_the_project.rst` is already in the portBLAS section. Are you suggesting it should be more even more explicit?

No, this is enough to be part of portBLAS documentation.

mkrainiuk (approved):

Issue #459, opened 2024-03-13
[Build][Docs] Improve build documentation


hjabird:
# Summary

oneMKL can be difficult to build and use. The reasons for this are:
* The build documentation is fragmented and at times outdated
* The compiler ecosystem and compatibility is a little confusing

# Problem statement

The current build documentation has the following problems:
* It suggests oneMKL supports Conan. Conan is no longer supported since https://github.com/oneapi-src/oneMKL/discussions/267
  * Consequently, you have scroll some way to find the CMake instructions.
* In most cases, building oneMKL backends is a case of setting `-DENABLE_XYZ_BACKEND=ON`. This isn't clearly communicated by the documentation, that makes building look far more complicated.
* The docs suggests setting `TARGET_DOMAINS` even thought it is set automatically. 
* At times the documentation is worded in a way that suggests that backends cannot be enabled simultaneously, even thought I think they can.
* In most cases, building a backend with hipSYCL (which is now AdaptiveCpp) is similar to building with clang++ or icpx, but we have duplication of information instead.

We also support a range of compilers. 
* I believe that some backends build with ICPX and Codeplay's plugins for Nvidia and AMD GPUs, although this is undocumented. We should consider supporting and documenting these. 
  * [Codeplay's oneAPI for Nvidia GPUs](https://developer.codeplay.com/products/oneapi/nvidia/2024.0.2/guides/), [Codeplay's oneAPI for AMD GPUs](https://developer.codeplay.com/products/oneapi/amd/2024.0.2/guides/get-started-guide-amd)
* In cases where a backend is dependent on a particular compiler, we should consider trying to generate useful CMake errors.

# Preferred solution

* Consider rewriting `building_the_project.rst` to address the above problems.
  * Some aspects (e.g. additional options for building with AdaptiveCpp) could potentially move to new files.
* Generate more errors in CMake to reflect compiler support for different backends.

cdgarland:
Hi @hjabird We are looking for community support for this project. Are you (or any others) willing to pitch in and help improve the documentation?

Rbiessy:
Hi Craig, yes we will try to help with some of these issues when time permits. It would be useful to have some feedback before we start spending time on it.

[...]

PR #482, opened 2024-05-10
workflow for publishing docs

rscohn2:
# Description

Build/publish documentation automatically when merging to develop. To enable this workflow, someone with admin rights needs to click settings/pages. And then change the "source" to "GitHub Actions"

Partially addresses https://github.com/oneapi-src/oneMKL/issues/459. I also wasted time following the conan instructions, then found that conan support was dropped. I also saw that someone had updated the documentation to reflect the change. But I never saw that because the new documentation was not published. The published documentation has not been updated for 2 years!


.github/workflows/pr.yml
```diff
@@ -0,0 +1,50 @@
+name: "PR Tests"
```
hjabird: It might be worth making this super-clear that this is for CPU only - otherwise we might end up with a false sense of security that all is good when the tests pass for code that isn't being compiled.

rscohn2: I intended for the unit tests workflow to be in #483 . I will remove from this PR.

requirements.txt
```diff
@@ -0,0 +1 @@
+Sphinx
```
mkrainiuk, line 1: Is it possible to move this requirement document to `doc`? So that it will be clear that it's requirements only for Documentation, not for the full oneMKL build.

dnhsieh-intel:
Would it be sufficient to use ["publishing from a branch"](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-from-a-branch)? We might also want to restrict publishing only when there are changes in the `docs` directory.

rscohn2:
It's a different implementation to publish from a branch. Publishing from a branch is a little more manual, but also more flexible. I can make the changes, but I would rather do it after https://github.com/oneapi-src/oneMKL/pull/483 is merged. It will work better if there is a single workflow file so I can structure checks, unit tests, and doc publishing as CI jobs that depend on each other. With that in mind, I am closing this PR. It will help me to wrap this up if we can get https://github.com/oneapi-src/oneMKL/pull/483 merged and I can add new CI jobs there.

PR #483, opened 2024-05-11
GitHub workflow for running unit tests for MKL backend on CPU

rscohn2:
# Description

GitHub Actions workflow to run unit tests for MKL/CPU. 

I disabled tests that require GPU by excluding tests where the name ends in `gpu`. RNG and DFT domains finish in 10 minutes. I am still waiting for BLAS/LAPACK to finish. We could have nightly testing that runs everything and PR testing that runs a smaller set. Any thoughts on how to reduce the testing for PR's?

@mkrainiuk

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -49,10 +49,10 @@ jobs:
     - name: Configure/Build for a domain
       run: |
         source /opt/intel/oneapi/setvars.sh
-        cmake -DREF_BLAS_ROOT=${PWD}/lapack/install -DREF_LAPACK_ROOT=${PWD}/lapack/install -DTARGET_DOMAINS=${{ matrix.domain }} -B build
+        cmake -DREF_BLAS_ROOT=${PWD}/lapack/install -DREF_LAPACK_ROOT=${PWD}/lapack/install -DENABLE_MKLGPU_BACKEND=off -DTARGET_DOMAINS=${{ matrix.domain }} -B build
         cmake --build build ${PARALLEL}
     - name: Run tests
       run: |
         source /opt/intel/oneapi/setvars.sh
         # Run tests, skip GPU tests
-        ctest --test-dir build ${PARALLEL} -E 'gpu$'
+        ctest --test-dir build ${PARALLEL} -E 'ct$'
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -15,7 +15,7 @@ env:
 
 jobs:
   mkl:
-    runs-on: ubuntu-latest
+    runs-on: intel-ubuntu-latest
     # One runner for each domain
     strategy:
       matrix:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -15,7 +15,7 @@ env:
 
 jobs:
   mkl:
-    runs-on: [ubuntu-latest, intel-ubuntu-latest]
+    runs-on: intel-ubuntu-latest
     # One runner for each domain
     strategy:
       matrix:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -15,7 +15,7 @@ env:
 
 jobs:
   mkl:
-    runs-on: intel-ubuntu-latest
+    runs-on: [ubuntu-latest, intel-ubuntu-latest]
     # One runner for each domain
     strategy:
       matrix:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -49,10 +49,10 @@ jobs:
     - name: Configure/Build for a domain
       run: |
         source /opt/intel/oneapi/setvars.sh
-        cmake -DREF_BLAS_ROOT=${PWD}/lapack/install -DREF_LAPACK_ROOT=${PWD}/lapack/install -DENABLE_MKLGPU_BACKEND=off -DTARGET_DOMAINS=${{ matrix.domain }} -B build
+        cmake -DREF_BLAS_ROOT=${PWD}/lapack/install -DREF_LAPACK_ROOT=${PWD}/lapack/install -DENABLE_MKLGPU_BACKEND=off -DTARGET_DOMAINS=${{ matrix.domain }} -DCMAKE_VERBOSE_MAKEFILE=on -B build
         cmake --build build ${PARALLEL}
     - name: Run tests
       run: |
         source /opt/intel/oneapi/setvars.sh
         # Run tests, skip GPU tests
-        ctest --test-dir build ${PARALLEL} -E 'ct$'
+        ctest --test-dir build -j 1 -E 'ct$'
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,57 @@
+name: "PR Tests"
+permissions: read-all
+
+# Trigger for PR an merge to develop branch
+on:
+  push:
+    branches: develop
+  pull_request:
+  workflow_dispatch:
+
+env:
+  CTEST_OUTPUT_ON_FAILURE: 1
+  LAPACK_VERSION: 3.12.0
+  PARALLEL: -j 2
+
+jobs:
+  mkl:
+    runs-on: ubuntu-latest
+    # One runner for each domain
+    strategy:
+      matrix:
+        domain: [blas, dft, lapack, rng]
+    steps:
+    - uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # v4.1.5
+    - name: Restore netlib from cache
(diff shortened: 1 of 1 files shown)


rscohn2:
lapack finishes in 30 minutes and blas takes 1.5 hours. All the time is building the tests. Combining multiple tests into a single binary might make testing faster without reducing coverage.

rscohn2:
@hjabird: Added CPU to job name to address: https://github.com/oneapi-src/oneMKL/pull/482#discussion_r1598245390

.github/workflows/pr.yml
```diff
@@ -0,0 +1,58 @@
...
+on:
+  push:
+    branches: develop
+  pull_request:
+  workflow_dispatch:
+
+env:
+  CTEST_OUTPUT_ON_FAILURE: 1
+  LAPACK_VERSION: 3.12.0
+  PARALLEL: -j 2
```
sknepper, line 14: What sort of system will this run on? If we can increase the number of parallel jobs, that should significantly decrease the build time of tests.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,58 @@
...
+        curl -sL https://github.com/Reference-LAPACK/lapack/archive/refs/tags/v${LAPACK_VERSION}.tar.gz | tar zx
+        SHARED_OPT="lapack-${LAPACK_VERSION} -DBUILD_SHARED_LIBS=on -DCBLAS=on -DLAPACKE=on -DCMAKE_INSTALL_PREFIX=${PWD}/lapack/install"
+        # 32 bit int
+        cmake ${SHARED_OPT} -B lapack/build32 
+        cmake --build lapack/build32 ${PARALLEL} --target install
+        # 64 bit int
+        cmake ${SHARED_OPT} -DBUILD_INDEX64=on -B lapack/build64
+        cmake --build lapack/build64 ${PARALLEL} --target install
+    - name: Install oneapi
+      uses: rscohn2/setup-oneapi@2ad0cf6b74bc2426bdcee825cf88f9db719dd727 # v0.1.0
```
sknepper, line 44: It looks like we'll need to update this workflow for every public release of mkl/icx. Will this uses line also need to be updated each time the version of mkl/icx changes? Is there a way to somehow signify "latest" that doesn't require regularly updating this workflow? Thanks!

.github/workflows/pr.yml
rscohn2: github provided runners have 1 core, 2 threads. You can have up to 20 concurrent jobs. There is a way to pay for runners with more cores. I don't think it would be expensive. Someone is looking into getting a dedicated node in devcloud that we could use. Then 1 runner could use 120 cores and 4 GPU's. This would solve most of our problems so I am hoping we could live with slow performance until that happens.

.github/workflows/pr.yml
rscohn2: `rscohn2/setup-oneapi@v0`
would pick up the latest. I use hashes because I believe the openssf scanning tool flags anything that is not a hash.  After we move the repo to onemkl-project org, we could decide to allow versions instead of hashes. Pinning the CI to a specific compiler/MKL makes CI more stable/reproducible so I recommend leaving it as is.

mkrainiuk: Should we use faceless account for such automation?

.github/workflows/pr.yml
```diff
@@ -0,0 +1,58 @@
...
+  workflow_dispatch:
+
+env:
+  CTEST_OUTPUT_ON_FAILURE: 1
+  LAPACK_VERSION: 3.12.0
+  PARALLEL: -j 2
+
+jobs:
+  mkl:
+    runs-on: intel-ubuntu-latest
```
dnhsieh-intel, line 18: We run CPU tests on Red Hat 9 internally. If there are no issues, we could use this as an extended testing.

.github/workflows/pr.yml
rscohn2: My id appears there because that is the URL for the plugin that is used to install oneapi components:
https://github.com/rscohn2/setup-oneapi
The plugin simplifies the install and reduces the install time from 5 minutes to 30 seconds.

.github/workflows/pr.yml
rscohn2: This is public testing for PR's. Non-intel contributors and reviewers will be able to see and correct problems before they are merged. It is limited by compute resources available: cpu only, single core runners, etc. The intel testing is needed to provide extensive coverage. but requires intel people to look at the results.

.github/workflows/pr.yml
mkrainiuk: I understand why we need it, but as far as I know using personal id could be a problem for security scans, but I'm not sure if it's a case here.

rscohn2: I took a couple repos through the openssf badge process which includes openssf best practices. There were several requirements related to github actions, but I don't think there is anything about which org a GitHub Action comes from. The rules for private repos with access to intel resources are different, but this is a public repo. In any case, if there is a problem in the future, I will fix it.

mkrainiuk:
> Any thoughts on how to reduce the testing for PR's? 

I guess we can also skip building GPU part with `-DENABLE_MKLGPU_BACKEND=False` for now since it's not tested anyway, so it will save some build time, and we can run only runtime dispatching tests and ignore compiler time dispatching tests with `-E /CT/`, it should give us significant testing time reduction.

.github/workflows/pr.yml
dnhsieh-intel: I understand the scope and purpose of this public CI. I was thinking it might be a little confusing that we list only Red Hat 9 in [the supported operating systems](https://github.com/oneapi-src/oneMKL?tab=readme-ov-file#supported-operating-systems) for CPU but run tests on Ubuntu.

rscohn2: Thanks for clarifying.

rscohn2:
> > Any thoughts on how to reduce the testing for PR's?
> 
> I guess we can also skip building GPU part with `-DENABLE_MKLGPU_BACKEND=False` for now since it's not tested anyway, so it will save some build time, and we can run only runtime dispatching tests and ignore compiler time dispatching tests with `-E /CT/`, it should give us significant testing time reduction.

I made the change. It runs about 25% faster, which makes up for the intel hosted runners being 25% slower than github hosted runners. When you move to onemkl-project org, we can use github hosted runners without contending with other projects.

mkrainiuk (approved):

.github/workflows/pr.yml
```diff
@@ -0,0 +1,58 @@
...
+    - name: Install oneapi
+      uses: rscohn2/setup-oneapi@2ad0cf6b74bc2426bdcee825cf88f9db719dd727 # v0.1.0
+      with:
+        components: |
+          icx@2024.1.0
+          mkl@2024.1.0
+    - name: Configure/Build for a domain
+      run: |
+        source /opt/intel/oneapi/setvars.sh
+        cmake -DREF_BLAS_ROOT=${PWD}/lapack/install -DREF_LAPACK_ROOT=${PWD}/lapack/install -DENABLE_MKLGPU_BACKEND=off -DTARGET_DOMAINS=${{ matrix.domain }} -B build
```
dnhsieh-intel: A verbose build output could help debugging.
```suggestion
        cmake -DREF_BLAS_ROOT=${PWD}/lapack/install -DREF_LAPACK_ROOT=${PWD}/lapack/install -DENABLE_MKLGPU_BACKEND=off -DTARGET_DOMAINS=${{ matrix.domain }} -DCMAKE_VERBOSE_MAKEFILE=on -B build
```

dnhsieh-intel (approved):
Thanks for the PR! I have one suggestion.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,58 @@
...
+        cmake --build build ${PARALLEL}
+    - name: Run tests
+      run: |
+        source /opt/intel/oneapi/setvars.sh
+        # Run tests, skip GPU tests
+        ctest --test-dir build ${PARALLEL} -E 'ct$'
```
dnhsieh-intel: And I would suggest using `-j 1` for testing:
```suggestion
        ctest --test-dir build -j 1 -E 'ct$'
```
If we run in parallel, I am worried that the log would be hard to read when there are failures.

.github/workflows/pr.yml
rscohn2: > What sort of system will this run on? If we can increase the number of parallel jobs, that should significantly decrease the build time of tests.

The new github org: `onemkl-project` has runners for a 24 core system. It reduces the total time to 30 minutes. Still working on PVC and more cores.
https://github.com/onemkl-project/ci-test/actions/runs/9140509510

PR #490, opened 2024-05-14
[CMake] Remove FindRoc* files

hjabird:
# Description

This PR removes the files Findroc* to remove complexity. Additionally, it resolves an issue where HIP is not installed at the default location given by `HIP_PATH`. If the (as far as I now) undocumented `HIP_PATH` is not set, the build system (make, ninja) produces an error when the `.so` file is not found in the expected location.

The version in this PR seems to find the ROCm maths libraries by default and reduce the amount of CMake that we need to maintain.

The need for the files being deleted was not clear to me. Deleting them seemed like the most simple fix for the above issue, but we could also check the shared library exists and generate an error in the CMake instead. 

# Checklist

## All Submissions

[testrng.txt](https://github.com/oneapi-src/oneMKL/files/15482373/testrng.txt)
[lapackres.txt](https://github.com/oneapi-src/oneMKL/files/15482375/lapackres.txt)
[rocblas_test.log](https://github.com/oneapi-src/oneMKL/files/15482379/rocblas_test.log)


Rbiessy:
Did you test to build this with CMake 1.13, the minimum version required by the root CMakeLists.txt? It occurred to me just now that older version of CMake may not be able to find these packages.
We know it works for `rocfft`, hopefully the other domains will work too.

hjabird:
> Did you test to build this with CMake 1.13, the minimum version required by the root CMakeLists.txt? It occurred to me just now that older version of CMake may not be able to find these packages. We know it works for `rocfft`, hopefully the other domains will work too.

It didn't occur to me to do this. I've now built the library (tests disabled), and it builds with this patch with 3.13.

Rbiessy (approved):

mkrainiuk (approved):

src/rng/backends/rocrand/CMakeLists.txt
```diff
@@ -54,7 +54,9 @@
 
 set(LIB_NAME onemkl_rng_rocrand)
 set(LIB_OBJ ${LIB_NAME}_obj)
-find_package(rocRAND REQUIRED)
+find_package(hip REQUIRED)
+find_package(rocrand REQUIRED)
```
dnhsieh-intel, line 58: Two naive questions:
1. Why do we find `hip` additionally here but not for `rocblas` and `rocsolver`?
2. Should we also fix a version for `rocrand` or drop the fixed versions for `rocblas` and `rocsolver`? We might need to be more relaxed if we are going to impose a version restriction.

hjabird: Thank you for your review! I think both of these things need changing.

1. It looks like this was an oversight on my part. In the original scripts, rocsolver also links against `hip::host`, and rocblas uses its include directory. Presumably this still worked because the roc* packages depend on HIP, but my current solution is stictly incorrect. I'll fix this.
2. I think I should drop the version restrictions. I didn't have a fixed rocrand version because AMD's release notes didn't match the version of rocRAND that was installed on the system I was using - I had no idea what was the right.

hjabird: Done in [8e04fce](https://github.com/oneapi-src/oneMKL/pull/490/commits/8e04fce1724a382d3d80e9642bcfe6cdaca3dee8)

Rbiessy (approved):

dnhsieh-intel (approved):

PR #495, opened 2024-05-21
testing for portblas portfft

rscohn2:
# Description

Add Unit tests for portfft, portblas to CI

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -19,8 +19,20 @@ jobs:
     # One runner for each domain
     strategy:
       matrix:
-        domain: [blas, dft, lapack, rng]
-    name: MKL ${{ matrix.domain }} CPU
+        include:
+        - config: portBLAS
+          options: -DENABLE_PORTBLAS_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=blas
+        - config: portFFT
+          options: -DENABLE_PORTFFT_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=dft
+        - config: MKL BLAS
+          options: -DTARGET_DOMAINS=blas
+        - config: MKL DFT
+          options: -DTARGET_DOMAINS=dft
+        - config: MKL LAPACK
+          options: -DTARGET_DOMAINS=lapack
+        - config: MKL RNG
+          options: -DTARGET_DOMAINS=rng
+    name: ${{ matrix.config }} CPU
     steps:
     - uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # v4.1.5
     - name: Restore netlib from cache
@@ -49,10 +61,9 @@ jobs:
     - name: Configure/Build for a domain
       run: |
         source /opt/intel/oneapi/setvars.sh
-        cmake -DREF_BLAS_ROOT=${PWD}/lapack/install -DREF_LAPACK_ROOT=${PWD}/lapack/install -DENABLE_MKLGPU_BACKEND=off -DTARGET_DOMAINS=${{ matrix.domain }} -DCMAKE_VERBOSE_MAKEFILE=on -B build
+        cmake -DREF_BLAS_ROOT=${PWD}/lapack/install -DREF_LAPACK_ROOT=${PWD}/lapack/install -DENABLE_MKLGPU_BACKEND=off -DCMAKE_VERBOSE_MAKEFILE=on ${{ matrix.options }} -B build
         cmake --build build ${PARALLEL}
     - name: Run tests
       run: |
         source /opt/intel/oneapi/setvars.sh
-        # Run tests, skip GPU tests
-        ctest --test-dir build -j 1 -E 'ct$'
+        ctest --test-dir build
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -21,9 +21,9 @@ jobs:
       matrix:
         include:
         - config: portBLAS
-          options: -DENABLE_PORTBLAS_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=blas
+          options: -DENABLE_PORTBLAS_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=blas -DCMAKE_CXX_FLAGS="-fsycl -fsycl-targets=spir64"
         - config: portFFT
-          options: -DENABLE_PORTFFT_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=dft
+          options: -DENABLE_PORTFFT_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=dft -DCMAKE_CXX_FLAGS="-fsycl -fsycl-targets=spir64"
         - config: MKL BLAS
           options: -DTARGET_DOMAINS=blas
         - config: MKL DFT
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -25,7 +25,7 @@ jobs:
           # tests: '.*Nrm2Tests.RealSinglePrecision.*'
           tests: '.*'
         - config: portFFT
-          options: -DENABLE_PORTFFT_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=dft -DCMAKE_CXX_FLAGS="-fsycl -fsycl-targets=spir64_x86_64"
+          options: -DENABLE_PORTFFT_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=dft -DCMAKE_CXX_FLAGS="-fsycl -fsycl-targets=spir64"
           tests: 'DFT/CT/.*ComputeTests_in_place_COMPLEX.COMPLEX_SINGLE_in_place_buffer.sizes_8_batches_1*'
         - config: MKL BLAS
           options: -DTARGET_DOMAINS=blas
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -21,17 +21,23 @@ jobs:
       matrix:
         include:
         - config: portBLAS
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -19,8 +19,20 @@ jobs:
     # One runner for each domain
     strategy:
       matrix:
-        domain: [blas, dft, lapack, rng]
-    name: MKL ${{ matrix.domain }} CPU
+        include:
+        - config: portBLAS
+          options: -DENABLE_PORTBLAS_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=blas -DCMAKE_CXX_FLAGS="-fsycl -fsycl-targets=spir64"
```
Rbiessy:
```suggestion
          options: -DENABLE_PORTBLAS_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=blas
```

The sycl flags shouldn't be needed, `-fsycl` is added automatically in oneMKL and `-fsycl-targets=spir64` is the default.

.github/workflows/pr.yml
```diff
@@ -19,8 +19,20 @@ jobs:
...
+        - config: portFFT
+          options: -DENABLE_PORTFFT_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=dft -DCMAKE_CXX_FLAGS="-fsycl -fsycl-targets=spir64"
```
Rbiessy, line 27:
```suggestion
          options: -DENABLE_PORTFFT_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=dft
```
Same comment than above.

hjabird (commented):
With Romain's comments, LGTM.

Rbiessy:
We'll look into ways to reduce the time it takes to run the tests. We may need to cache the SYCL kernels across the tests.

.github/workflows/pr.yml
rscohn2: I had to do this because it defaults to spirv+cuda: https://github.com/oneapi-src/oneMKL/actions/runs/9180847269/job/25246216751#step:6:128:~:text=per_kernel%20%2DWall%20%2DWextra%20%2DWshadow%20%2DWconversion%20%2DWpedantic%20%2Dfsycl%20%2Dfsycl%2Dunnamed%2Dlambda-,%2Dfsycl%2Dtargets%3Dnvptx64%2Dnvidia%2Dcuda%2Cspir64,-%2DMD%20%2DMT%20bin/dft/backends/portfft/CMakeFiles/onemkl_dft_portfft_obj.dir/descriptor

and the portfft recommended fixing it by setting `CMAKE_CXX_FLAGS`:
https://oneapi-src.github.io/oneMKL/building_the_project.html#building-for-portblas:~:text=The%20portFFT%20library%20is%20compiled%20using%20the%20same%20%2Dfsycl%2Dtargets

If you set CMAKE_CXX_FLAGS for the target, then you also need to set `-fsycl`

portblas talks about `PORTBLAS_TUNING_TARGET`:  https://oneapi-src.github.io/oneMKL/building_the_project.html#building-for-portblas:~:text=%2DDPORTBLAS_TUNING_TARGET%3D%3Ctarget%3E

It would be better if it defaulted to only `spirv` and the documentation was clearer about how to change the target.

What is the simplest way to get it to only target CPU?

rscohn2:
> We'll look into ways to reduce the time it takes to run the tests. We may need to cache the SYCL kernels across the tests.

The build times are comparable to MKL backend. MKL tests run in < 5 minutes. CI failed with a timeout because both portblas and portfft tests take > 4 hours to run. Are you saying that driver compile time is the cause? I can add a regex to select a subset of the tests to run, with a target of running for 5 minutes. What do you recommend?

.github/workflows/pr.yml
Rbiessy: For portBLAS the tuning target is optional. We don't need to set it for CPU or for quick tests. From my local tests the `CMAKE_CXX_FLAGS` can be removed.

I can see we have an issue with the portFFT backend. We wanted the default to also run on Nvidia but it seems we only tested it on machines where CUDA was installed so using `nvptx64-nvidia-cuda,spir64` by default worked. We'll need to improve this in the future.

.github/workflows/pr.yml
Rbiessy: Resolving this comment as we need to fix the portFFT backend to not use `nvptx64-nvidia-cuda` if the CUDA runtime is not installed.

Rbiessy:
> The build times are comparable to MKL backend. MKL tests run in < 5 minutes. CI failed with a timeout because both portblas and portfft tests take > 4 hours to run. Are you saying that driver compile time is the cause? I can add a regex to select a subset of the tests to run, with a target of running for 5 minutes. What do you recommend?

Yes the JIT compilation may be the issue, we'll need to confirm. I think in the current situation we would need to disable too many tests so I would suggest to wait until we know more. Running all the tests under 5 minutes seems very optimistic for portFFT or portBLAS though.

.github/workflows/pr.yml
rscohn2: Will remove for portblas

rscohn2:
I changed it so we build all the unit tests for portfft/portblas but only run a handful of tests. Everything passes and port* finishes before MKL. It is not what we want for the long term, but we can enable more as we get faster machines and hopefully find a way to reduce the JIT time. Are you OK with committing what is there now?

dnhsieh-intel:
An indirectly related question: I noticed that previous workflows were run on [my forked repo](https://github.com/dnhsieh-intel/oneMKL/actions). Do you think a condition like [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository) can avoid the runs on forked repos?

rscohn2:
> Do you think a condition like [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository) can avoid the runs on forked repos?

That will work, but you have to put the condition in every job for every workflow. I don't think there is a file-level way to do it. An alternative is to disable github actions for your fork. It is in settings/actions/general, then look near the top.

dnhsieh-intel:
I see. I hope GitHub can provide a more maintainable way for this. At the same time, we may need to communicate this with contributors, but this is out of scope of this PR.

Rbiessy:
@rscohn2 from our local testing we have confirmed that using AOT compilation of the kernels speeds up the testing a lot. This is best to use if we know we will run on a specific device.
For the portBLAS backend it turns out this can be done by setting the cmake flag `-DPORTBLAS_TUNING_TARGET=INTEL_CPU`. This sets the SYCL target to `spir64_x86_64`. With this we were able to compile in 32 minutes and run all the tests in 8 minutes on a i9-12900K. I think it would take a bit longer with the current GitHub runner though.

For the portFFT backend you might as well directly set the target to `spir64_x86_64` and enable more tests if that's ok.

We'll try to improve the documentation for this. We need to find a meaningful default behavior while still letting the user tune for more specific use-cases.

rscohn2:
The github runner has 2 cores/4 threads, but I think we will have more cores available soon. For the 8 minute test, were they parallel or serial? @dnhsieh-intel prefers serial tests because the logs are easier to read.

Rbiessy:
We just ran `ctest` so the tests should have been run serially when we measured 8 minutes.

rscohn2:
portfft tests are not passing with aot

https://github.com/oneapi-src/oneMKL/actions/runs/9208889271/job/25332103601#:~:text=48-,unknown%20file%3A%20Failure,%5B%20%20FAILED%20%20%5D,-ComputeTestSuite/ComputeTests_in_place_COMPLEX.COMPLEX_SINGLE_in_place_buffer

Rbiessy:
Right, that makes sense actually portFFT is using spec constants. Let's revert the target to `spir64` for portFFT then.

rscohn2:
With AOT, portblas takes 1.5 hours to compile and 5 minutes to run. Runners only have 2 cores, but we are getting runners with more cores so parallel build will be faster. 2nd longest is mkl blas, which takes 1 hour with most time spent in compile.
portfft is back to not using aot and running a couple tests. Everything passes and runs in an acceptable amount of time.

.github/workflows/pr.yml
```diff
@@ -14,13 +14,32 @@ env:
...
+    runs-on: ubuntu-latest
     # One runner for each domain
     strategy:
       matrix:
-        domain: [blas, dft, lapack, rng]
-    name: MKL ${{ matrix.domain }} CPU
+        include:
+        - config: portBLAS
+          options: -DENABLE_PORTBLAS_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=blas -DPORTBLAS_TUNING_TARGET=INTEL_CPU
+          # tests: '.*Nrm2Tests.RealSinglePrecision.*'
```
Rbiessy:
```suggestion
```

Rbiessy (approved):
Just a minor comment, looks good to me otherwise, thanks

.github/workflows/pr.yml
```diff
@@ -14,13 +14,31 @@ env:
...
-    name: MKL ${{ matrix.domain }} CPU
+        include:
+        - config: portBLAS
+          options: -DENABLE_PORTBLAS_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=blas -DPORTBLAS_TUNING_TARGET=INTEL_CPU
+          tests: '.*'
+        - config: portFFT
+          options: -DENABLE_PORTFFT_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=dft -DCMAKE_CXX_FLAGS="-fsycl -fsycl-targets=spir64"
+          tests: 'DFT/CT/.*ComputeTests_in_place_COMPLEX.COMPLEX_SINGLE_in_place_buffer.sizes_8_batches_1*'
+        - config: MKL BLAS
+          options: -DTARGET_DOMAINS=blas
```
dnhsieh-intel: This is a minor improvement to avoid CMake warnings. Since now we separate domain options, we may move `-DREF_BLAS_ROOT` here and `-DREF_LAPACK_ROOT` in `MKL LAPACK`.

dnhsieh-intel (approved):

.github/workflows/pr.yml
rscohn2: moved...

PR #496, opened 2024-05-23
[SparseBlas]Initial design of Sparse Coo data

abhilash1910:
# Description

## Motivation 

Currently from one of the priority app CUDA - SYCL migration perspective  (CUDA-SYCL Task Force) (bitsandbytes) , there is a requirement to support COO format. Since currently the scope is through CSR, the code migration is routed through CSR pathway. However for an efficient migration of CUDA applications invoking COO, I propose a design for COO format.  I have taken reference from the existing CSR implementation as most of the function args are similar.  The current application requires ```cusparseCreateCoo``` api, and I had taken reference from this api invokation:

`
cusparseCreateCoo(cusparseSpMatDescr_t* spMatDescr, 
                  int64_t               rows,
                  int64_t               cols,
                  int64_t               nnz,
                  void*                 cooRowInd,
                  void*                 cooColInd,
                  void*                 cooValues,
                  cusparseIndexType_t   cooIdxType,
                  cusparseIndexBase_t   idxBase,
                  cudaDataType          valueType)`
                  

It would entail development of some COO auxiliary methods , getters ,aos and const variants in general.  Since currently COO is not supported on iGPU yet , it would be nice to have this feature for future application migrations. 
Tagging @Rbiessy  from discussion.


Rbiessy:
Hello @abhilash1910, we have updated the specification for the sparse domain recently: https://github.com/uxlfoundation/oneAPI-spec/pull/522. Among other changes this adds support for the COO format.
I am already working on updating the MKL backends to match the new API. I expect the PR will be in review sometime next week so I would rather not change the existing implementation. FYI I will start working on adding support for cuSPARSE after that.

Rbiessy:
I think this PR can be closed in favor of https://github.com/oneapi-src/oneMKL/pull/500. Let me know @abhilash1910 whether this would work for you.

abhilash1910:
Thanks @Rbiessy , I will review shortly , in the meantime closing this.

PR #502, opened 2024-05-28
[DFT] Add static_assert to check types passed into compute_foward and compute_backward

### LINKED ISSUE #499: [DFT] compute_forward (forward_ip_rr) function template in forward.hpp overrides compute_forward (forward_op_cc) if in and out types are the same
# Summary
forward.hpp - "In-place transform, using config_param::COMPLEX_STORAGE=config_value::REAL_REAL data format" compute_forward function template overlaps with the "Out-of-place transform" where input_type == output_type  (for example, Compex to Complex DFT).

# Version
develop branch  v0.4 tag (#34)

# Environment
oneMKL DFT -  implementing a new GPU backend

# Observed behavior

When I set up my descriptor like this:
 desc.set_value(oneapi::mkl::dft::config_param::PLACEMENT, oneapi::mkl::dft::detail::config_value::NOT_INPLACE);
 desc.set_value(oneapi::mkl::dft::config_param::COMPLEX_STORAGE, oneapi::mkl::dft::detail::config_value::COMPLEX_COMPLEX);

 and call a compute_forward() with complex in and out parameters the in-place REAL_REAL function template overlaps and is incorrectly chosen instead of the out-of-place COMPLEX_COMPLEX transform.

forward.hpp:
```cpp
//In-place transform, using config_param::COMPLEX_STORAGE=config_value::REAL_REAL data format <- THIS GETS CHOSEN
void compute_forward(descriptor_type &desc, sycl::buffer<data_type, 1> &inout_re,
                     sycl::buffer<data_type, 1> &inout_im)  {
    static_assert(detail::valid_compute_arg<descriptor_type, data_type>::value,
                  "unexpected type for data_type");

    using scalar_type = typename detail::descriptor_info<descriptor_type>::scalar_type;
    auto type_corrected_inout_re = inout_re.template reinterpret<scalar_type, 1>(
        detail::reinterpret_range<data_type, scalar_type>(inout_re.size()));
    auto type_corrected_inout_im = inout_im.template reinterpret<scalar_type, 1>(
        detail::reinterpret_range<data_type, scalar_type>(inout_im.size()));
    get_commit(desc)->forward_ip_rr(desc, type_corrected_inout_re, type_corrected_inout_im);
}

//Out-of-place transform <- THIS SHOULD HAVE BEEN CHOSEN
template <typename descriptor_type, typename input_type, typename output_type>
void compute_forward(descriptor_type &desc, sycl::buffer<input_type, 1> &in,
                     sycl::buffer<output_type, 1> &out)
```

I can't use explicit template specialisation in my  DFT backend (forward.cpp) file because the oneapi::mkl::dft REAL_REAL function above reinterprets the "in" and "out" data from complex<float> (or complex<double>)  to scalar_type so my backend only sees compute_forward(desc, scalar_type, scalar_type).

# Expected behavior

REAL_REAL variant only gets selected if inout_re and inout_im are of float or double type (not complex type).

or, alternatively:

REAL_REAL variant only gets selected if:
 desc.set_value(oneapi::mkl::dft::config_param::PLACEMENT, oneapi::mkl::dft::detail::config_value::INPLACE);
 desc.set_value(oneapi::mkl::dft::config_param::COMPLEX_STORAGE, oneapi::mkl::dft::detail::config_value::REAL_REAL);

hjabird:
# Description

This adds static_asserts more rigidly check that input/output argument types are as expected.

Eg.
```c++
    auto in_usm = sycl::malloc_shared<std::complex<float>>(N, sycl_queue);
    auto out_usm = sycl::malloc_shared<std::complex<float>>(N, sycl_queue);
    oneapi::mkl::dft::descriptor<oneapi::mkl::dft::precision::SINGLE,
                                 oneapi::mkl::dft::domain::COMPLEX>
        desc(static_cast<std::int64_t>(N));
    desc.set_value(oneapi::mkl::dft::config_param::PLACEMENT,
                   oneapi::mkl::dft::config_value::NOT_INPLACE);
    desc.commit(sycl_queue);
    auto compute_event = oneapi::mkl::dft::compute_forward(desc, in_usm, out_usm);
```
now fails because the user has inadverently used called the compute_forward for in-place real-real DFTs. The error would help the user identify that they ought to be using
```c++
  oneapi::mkl::dft::compute_forward<decltype(desc), std::complex<float>, std::complex<float>>(desc, in_usm, out_usm);
```

Partially resolves https://github.com/oneapi-src/oneMKL/issues/499

# Checklist

## All Submissions

[testres.txt](https://github.com/oneapi-src/oneMKL/files/15472131/testres.txt)

### DIFF
--- a/include/oneapi/mkl/dft/backward.hpp
+++ b/include/oneapi/mkl/dft/backward.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright 2023 Intel Corporation
+* Copyright 2023-2024 Intel Corporation
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -49,6 +49,7 @@ void compute_backward(descriptor_type &desc, sycl::buffer<data_type, 1> &inout_r
                       sycl::buffer<data_type, 1> &inout_im) {
     static_assert(detail::valid_compute_arg<descriptor_type, data_type>::value,
                   "unexpected type for data_type");
+    static_assert(!detail::is_complex_arg<data_type>::value, "expected real type for data_type");
 
     using scalar_type = typename detail::descriptor_info<descriptor_type>::scalar_type;
     auto type_corrected_inout_re = inout_re.template reinterpret<scalar_type, 1>(
@@ -69,6 +70,13 @@ void compute_backward(descriptor_type &desc, sycl::buffer<input_type, 1> &in,
 
     using fwd_type = typename detail::descriptor_info<descriptor_type>::forward_type;
     using bwd_type = typename detail::descriptor_info<descriptor_type>::backward_type;
+
+    // If the DFT is COMPLEX, the input and output types are expected to be complex.
+    static_assert(
+        !std::is_same_v<fwd_type, bwd_type> || (detail::is_complex_arg<input_type>::value &&
+                                                detail::is_complex_arg<output_type>::value),
+        "expected std::complex input_type and output_type");
+
     auto type_corrected_in = in.template reinterpret<bwd_type, 1>(
         detail::reinterpret_range<input_type, bwd_type>(in.size()));
     auto type_corrected_out = out.template reinterpret<fwd_type, 1>(
@@ -85,6 +93,9 @@ void compute_backward(descriptor_type &desc, sycl::buffer<input_type, 1> &in_re,
                   "unexpected type for input_type");
     static_assert(detail::valid_compute_arg<descriptor_type, output_type>::value,
                   "unexpected type for output_type");
+    static_assert(
+        !detail::is_complex_arg<input_type>::value && !detail::is_complex_arg<output_type>::value,
+        "expected input_type and output_type to be real");
 
     using scalar_type = typename detail::descriptor_info<descriptor_type>::scalar_type;
     auto type_corrected_in_re = in_re.template reinterpret<scalar_type, 1>(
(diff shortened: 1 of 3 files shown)


Rbiessy (approved):

lhuot:
The case you describe in the PR description seems like one of the most common use-case. Is there any way we can get the correct template being picked up without having to explicitly pass the additional template argument? I think it would be okay to have to explicitly pass the template argument for the REAL_REAL descriptor but this seems really inconvenient for "standard" complex out of place FFTs.

hjabird:
> The case you describe in the PR description seems like one of the most common use-case. Is there any way we can get the correct template being picked up without having to explicitly pass the additional template argument? I think it would be okay to have to explicitly pass the template argument for the REAL_REAL descriptor but this seems really inconvenient for "standard" complex out of place FFTs.

I think the obvious way to do it would be to use `enable_if`. However, this would make the function signatures look much less like the [spec](https://spec.oneapi.io/versions/latest/elements/oneMKL/source/domains/dft/compute_forward.html).

lhuot:
> > The case you describe in the PR description seems like one of the most common use-case. Is there any way we can get the correct template being picked up without having to explicitly pass the additional template argument? I think it would be okay to have to explicitly pass the template argument for the REAL_REAL descriptor but this seems really inconvenient for "standard" complex out of place FFTs.
> 
> I think the obvious way to do it would be to use `enable_if`. However, this would make the function signatures look much less like the [spec](https://spec.oneapi.io/versions/latest/elements/oneMKL/source/domains/dft/compute_forward.html).

I think we found a gap in the spec that should be resolved, if implementing the spec is forcing the user to specify the template argument on the most commonly used FFT then there is a problem in my opinion. My suggestion would be to address to the best we can and by staying as close to the spec as possible this issue here and open an issue for the specification. For example, keeping the same entry point but adding a runtime check to re-dispatch the proper function as suggested by @raphael-egan. @Rbiessy @hjabird what do you think?

hjabird:
I've created a version using SFINAE at https://github.com/hjabird/oneMKL/tree/dft_overload_resolution_with_enableif. If this is what we're looking for, I'll close this PR and open a new PR for that.

The benefit is that the desired overload correctly gets chosen.

The downside is that if you don't have the correct types as inputs, you'll get long error messages.

hjabird:
It is likely that this PR will be closed in favour of https://github.com/oneapi-src/oneMKL/pull/503.

hjabird:
Closing since #503 is the preferred solution.

PR #504, opened 2024-05-31
Selective CI based on changes to a domain

rscohn2:
# Description

Selective CI based on changes to a domain. Requested by @sknepper.

I tested that the filter works by making changes to RNG and DFT, src and tests directories. I undid the changes to prepare to merge.

It does not skip the first few steps because they finish almost immediately. It skips the expensive steps starting with oneapi install.


.github/workflows/pr.yml
```diff
@@ -58,18 +57,31 @@ jobs:
         # 64 bit int
         cmake ${SHARED_OPT} -DBUILD_INDEX64=on -B lapack/build64
         cmake --build lapack/build64 ${PARALLEL} --target install
+    - name: Check for changes to this domain
+      id: changes
+      uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
+      with:
+        filters: |
+          src:
+          - 'src/${{ matrix.domain}}/**'
```
dnhsieh-intel: nitpicking
```suggestion
          - 'src/${{ matrix.domain }}/**'
```

.github/workflows/pr.yml
```diff
@@ -58,18 +57,31 @@ jobs:
...
+          - 'tests/unit_tests/${{ matrix.domain }}/**'
+          - CMakeLists.txt
+          - '.github/workflows/**'
```
dnhsieh-intel: I would also add `examples` and `include` directories because we test examples and headers could affect the build.

sknepper: I agree. Thanks, @rscohn2 !

rscohn2: I am looking for a more maintainable way to specify the files that will trigger a test.

PR #508, opened 2024-06-05
Selective CI basecd on changes to a domain

rscohn2:
# Description

Selective CI based on changes to a domain.

Tested in a fork with changes to domain directories

It does not skip the first few steps because they finish almost immediately. It skips the expensive steps starting with oneapi install.

Continuation of https://github.com/oneapi-src/oneMKL/pull/504. Resolves review comments. Using javascript directly instead of the 3rd party action. The action was not being maintained by the author.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -21,23 +21,28 @@ jobs:
       matrix:
         include:
         - config: portBLAS
-          options: -DTARGET_DOMAINS=blas -DREF_BLAS_ROOT=${PWD}/lapack/install -DENABLE_PORTBLAS_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DPORTBLAS_TUNING_TARGET=INTEL_CPU
-          tests: '.*'
+          domain: blas
+          exclude: lapack|dft|rng
+          build_options: -DREF_BLAS_ROOT=${PWD}/lapack/install -DENABLE_PORTBLAS_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DPORTBLAS_TUNING_TARGET=INTEL_CPU
         - config: portFFT
-          options: -DENABLE_PORTFFT_BACKEND=ON -DENABLE_MKLCPU_BACKEND=OFF -DTARGET_DOMAINS=dft -DCMAKE_CXX_FLAGS="-fsycl -fsycl-targets=spir64"
-          tests: 'DFT/CT/.*ComputeTests_in_place_COMPLEX.COMPLEX_SINGLE_in_place_buffer.sizes_8_batches_1*'
+          domain: dft
+          exclude: blas|lapack|rng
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -58,18 +57,44 @@ jobs:
...
+      uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+      with:
+        script: |
+          function matchesPattern(filePaths) {
+            console.log('files:', filePaths)
+            // These directories contain domain specific code
+            const dirs = '(tests/unit_tests|examples|src|include/oneapi/mkl)'
+            const domains = '(blas|lapack|rng|dft)'
+            const domain = '${{ matrix.domain }}'
+            // matches changes to the domain of interest or non domain-specific code
```
Rbiessy, line 71: I don't fully understand the syntax here so just double-checking: does this mean that if there is a non domain-specific change, every domain will be tested?

rscohn2: > does this mean that if there is a non domain-specific change, every domain will be tested?

Yes.

```
${dirs}/${domain}
```
Is the domain-specific code that should trigger a test. e.g. tests/unit_tests/blas
```
?!${dirs}/${domains}
```
Is all the non-domain specific code. The `?!` is a negation and `${dirs}/${domains}` is all the domain specific code for all domains. Testing is triggered if either of above matches.

Javascript is the underlying language for the GitHub Actions, and using a javascript regular expression makes it easy to test the code outside of the CI system. Simple whitelisting is likely to be wrong or out of date.

Rbiessy: Thank you, this looks good to me.

sknepper (approved):
LGTM, thanks, @rscohn2 !
And thanks to @dnhsieh-intel for the suggestion on the previous PR to add the examples and include directories!

.github/workflows/pr.yml
```diff
@@ -58,18 +57,44 @@ jobs:
...
+            const re = new RegExp(`^(${dirs}/${domain}|(?!${dirs}/${domains}))`);
```
dnhsieh-intel, line 72: Would it be a match if the changes are in `docs`, `scripts`, `third-party-programs` directories, or `*.md` files?

rscohn2: Yes. It would match. Now that the  logic is javascript and a regular expression, you can make the filter as selective as you want. It is straightforward to exclude files that are certain not to affect the build like `*.md`. If the team wants to fine tune the selection logic, I would first move the javascript to a separate file so it could more easily be tested outside the CI system. 

I would rather do it in a separate PR. I want to propose adding some static checkers and I want to get that started with a PR because it might take a while to settle on something that the team is happy with.

dnhsieh-intel: Thank you!

dnhsieh-intel (approved):

PR #514, opened 2024-06-13
[DFT][MKLGPU] Use FWD/BWD_STRIDES

### LINKED ISSUE #487: [DFT][MKLGPU] Use FWD_ and BWD_STRIDES for MKLGPU 2024.1
# Summary
The MKLGPU library should use updated oneMKL interface for forward and bwd strides for the MKLGPU backend.

# Problem statement
The oneMKL specification was revised to use FWD_ and BWD_STRIDES instead of INPUT_ and OUTPUT_STRIDES. This has been implemented in Intel's oneMKL for 2024.1. Whilst oneMKL interface library still works, MKLGPU prints deprecation notices. See https://www.intel.com/content/www/us/en/developer/articles/release-notes/onemkl-release-notes.html.

# Preferred solution
When using Intel oneMKL 2024.1 or later, use the updated API for strides.

hjabird:
# Description

* Intel(R) oneMKL 2024.1 deprecate the INPUT/OUTPUT_STRIDES API
* Using the old API causes error messages to be printed.
* This PR uses the new FWD/BWD_STRIDES API, but maintains support for older Intel(R) oneMKL.
* Fixes a warning
  * `sources/src/dft/backends/mklgpu/descriptor.cpp:30:29: warning: a declarative nested name specifier cannot name an alias template [-Walias-template-in-declaration-name]`
  * This warning seems to need nightly clang rather than icpx.
* It additionally fixes one-line bug in cuFFT (albeit one that doesn't seem to cause any issues).

Fixes https://github.com/oneapi-src/oneMKL/issues/487

# Checklist

## All Submissions

  - [testres.txt](https://github.com/user-attachments/files/15827902/testres.txt)
  - Tests only attached for MKLGPU. Other domains have the minor modification to stop the compiler warning. Ctest has been run and passes.

### DIFF
--- a/src/dft/backends/mklgpu/commit.cpp
+++ b/src/dft/backends/mklgpu/commit.cpp
@@ -37,6 +37,12 @@
 // MKLGPU header
 #include "oneapi/mkl/dfti.hpp"
 
+// MKL 2024.1 deprecates input/output strides.
+#include "mkl_version.h"
+namespace oneapi::mkl::dft::mklgpu::detail {
+constexpr bool mklgpu_use_forward_backward_strides_api = INTEL_MKL_VERSION >= 20240001;
+}
+
 /**
 Note that in this file, the Intel oneMKL closed-source library's interface mirrors the interface
 of this OneMKL open-source library. Consequently, the types under dft::TYPE are closed-source oneMKL types,
@@ -53,14 +59,22 @@ private:
     // Equivalent MKLGPU precision and domain from OneMKL's precision / domain.
     static constexpr dft::precision mklgpu_prec = to_mklgpu(prec);
     static constexpr dft::domain mklgpu_dom = to_mklgpu(dom);
+
+    // A pair of descriptors are needed because of the [[deprecated]]IN/OUTPUT_STRIDES vs F/BWD_STRIDES API.
+    // Of the pair [0] is fwd DFT, [1] is backward DFT. If possible, the pointers refer to the same desciptor.
+    // Both pointers must be valid.
     using mklgpu_descriptor_t = dft::descriptor<mklgpu_prec, mklgpu_dom>;
+    using descriptor_shptr_t = std::shared_ptr<mklgpu_descriptor_t>;
+    using handle_t = std::pair<descriptor_shptr_t, descriptor_shptr_t>;
+
     using scalar_type = typename dft::detail::commit_impl<prec, dom>::scalar_type;
 
 public:
     mklgpu_commit(sycl::queue queue, const dft::detail::dft_values<prec, dom>& config_values)
             : oneapi::mkl::dft::detail::commit_impl<prec, dom>(queue, backend::mklgpu,
                                                                config_values),
-              handle(config_values.dimensions) {
+              handle(std::make_shared<mklgpu_descriptor_t>(config_values.dimensions), nullptr) {
+        handle.second = handle.first; // Make sure the bwd pointer is valid.
         // MKLGPU does not throw an informative exception for the following:
         if constexpr (prec == dft::detail::precision::DOUBLE) {
             if (!queue.get_device().has(sycl::aspect::fp64)) {
@@ -75,13 +89,31 @@ public:
             oneapi::mkl::dft::detail::external_workspace_helper<prec, dom>(
                 config_values.workspace_placement ==
                 oneapi::mkl::dft::detail::config_value::WORKSPACE_EXTERNAL);
-        set_value(handle, config_values);
+        // Generate forward DFT descriptor.
+        set_value(*handle.first, config_values, true);
         try {
-            handle.commit(this->get_queue());
+            handle.first->commit(this->get_queue());
         }
         catch (const std::exception& mkl_exception) {
             // Catching the real Intel oneMKL exception causes headaches with naming.
             throw mkl::exception("dft/backends/mklgpu", "commit", mkl_exception.what());
+        }
+
+        // Generate backward DFT descriptor only if required.
+        if (config_values.input_strides == config_values.output_strides) {
+            // Required if second != first before a recommit.
+            handle.second = handle.first;
+        }
+        else {
+            handle.second = std::make_shared<mklgpu_descriptor_t>(config_values.dimensions);
+            set_value(*handle.second, config_values, false);
+            try {
+                handle.second->commit(this->get_queue());
+            }
+            catch (const std::exception& mkl_exception) {
+                // Catching the real Intel oneMKL exception causes headaches with naming.
+                throw mkl::exception("dft/backends/mklgpu", "commit", mkl_exception.what());
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/dft/backends/mklgpu/commit.cpp
```diff
@@ -53,14 +57,22 @@ class mklgpu_commit final : public dft::detail::commit_impl<prec, dom> {
     // Equivalent MKLGPU precision and domain from OneMKL's precision / domain.
     static constexpr dft::precision mklgpu_prec = to_mklgpu(prec);
     static constexpr dft::domain mklgpu_dom = to_mklgpu(dom);
+
+    // A pair of descriptors are needed because of the [[deprecated]]IN/OUTPUT_STRIDES vs F/BWD_STRIDES API.
+    // Of the pair [0] is fwd DFT, [1] is backward DFT. If possible, the pointers refer to the same desciptor.
+    // Both pointers must be valid.
```
Rbiessy, line 65: I didn't realize using F/BWD_STRIDES would add this much complexity. Would we be able to keep only one descriptor if we required oneAPI 2024.1?
In previous PRs we have discussed that we typically only need to support the latest release: https://github.com/oneapi-src/oneMKL/pull/468#issuecomment-2040116467
I'm concerned of adding complexity for a feature that users may not need.

hjabird: Unfortunately, I couldn't see a nice way to avoid needing two descriptors whilst mixing the INPUT/OUTPUT & FWD/BWD APIs. The only way I could see of doing this is by continuing to use the I/O API and hiding the deprecation warnings somehow, but I don't know if that's possible.

src/dft/backends/mklgpu/commit.cpp
```diff
@@ -75,13 +87,37 @@ class mklgpu_commit final : public dft::detail::commit_impl<prec, dom> {
...
+            handle.first->commit(this->get_queue());
         }
         catch (const std::exception& mkl_exception) {
             // Catching the real Intel oneMKL exception causes headaches with naming.
-            throw mkl::exception("dft/backends/mklgpu", "commit", mkl_exception.what());
+            fwd_exception = mkl::exception("dft/backends/mklgpu", "commit", mkl_exception.what());
+        }
+
+        // Generate backward DFT descriptor only if required.
+        if (config_values.input_strides == config_values.output_strides) {
```
Rbiessy: I haven't worked on the DFT domain in a long time. How come we can be sure that we won't need the backward descriptor if the input and output strides are equal?

src/dft/backends/mklgpu/commit.cpp
```diff
@@ -158,9 +214,10 @@ class mklgpu_commit final : public dft::detail::commit_impl<prec, dom> {
 
     // This is called by the workspace_helper, and is not part of the user API.
     virtual std::int64_t get_workspace_external_bytes_impl() override {
-        std::size_t workspaceSize = 0;
-        handle.get_value(dft::config_param::WORKSPACE_BYTES, &workspaceSize);
-        return static_cast<std::int64_t>(workspaceSize);
+        std::size_t workspaceSizeFwd = 0, workspaceSizeBwd;
```
Rbiessy: Nitpicking but it's weird that one variable is initialized and not the other in this case.

src/dft/backends/mklgpu/commit.cpp
Rbiessy: Right, I am still wondering if we should completely remove support for `INPUT/OUTPUT STRIDES` and require oneAPI 2024.1 once oneMKL Interface supports `FWD/BWD STRIDES`. I think it would allow to have just one descriptor for MKLGPU and portFFT backends but I don't know the impact on the other backends. I understand this would be a breaking change. I'm curious to know the opinion of @lhuot on the matter.
Note that we could just hide the deprecation warnings with pragma like [here](https://github.com/oneapi-src/oneMKL/blob/1988d4e3bcea2ae8dbba1835dc0adb7d1047b5a6/tests/unit_tests/dft/include/test_common.hpp#L100) if we just want a quick fix. It would be better to find a long term solution.

src/dft/backends/mklgpu/commit.cpp
hjabird: We need two descriptors because we have to swap the in/output strides between the forward and backward descriptors. If the strides are the same between the two, then swapping them doesn't achieve anything, so we can use a single descriptor.

hjabird:
> Note that we could just hide the deprecation warnings with pragma like [here](https://github.com/oneapi-src/oneMKL/blob/1988d4e3bcea2ae8dbba1835dc0adb7d1047b5a6/tests/unit_tests/dft/include/test_common.hpp#L100) if we just want a quick fix. It would be better to find a long term solution.

MKL prints an warning to std::cerr at runtime, so I don't think this is the fix we're looking for.

src/dft/backends/mklgpu/commit.cpp
lhuot: So you are only updating the oneMKL backends to use FWD/BWD_STRIDES, you are not changing the external API of the oneMKL Interfaces library to add FWD/BWD_STRIDES (those are now part of the oneMKL specification)?

src/dft/backends/mklgpu/commit.cpp
Rbiessy: Of course

src/dft/backends/mklgpu/commit.cpp
hjabird: This PR only changes the MKLGPU backend to use FWD/BWD_STRIDES. We will make the changes to support FWD/BWD_STRIDES in the oneMKL Interface Library interface in another PR.

src/dft/backends/mklgpu/commit.cpp
hjabird: Rebase + do this in [9ae3a61](https://github.com/oneapi-src/oneMKL/pull/514/commits/9ae3a61876b05e631ce796ce7b453e73935d4663).

lhuot:
> Right, I am still wondering if we should completely remove support for INPUT/OUTPUT STRIDES and require oneAPI 2024.1 once oneMKL Interface supports FWD/BWD STRIDES. I think it would allow to have just one descriptor for MKLGPU and portFFT backends but I don't know the impact on the other backends. I understand this would be a breaking change. I'm curious to know the opinion of @lhuot on the matter.

We should support the latest release as documented in the README so imo it is okay to require oneMKL 2024.1 and cleanup the code to use only fwd/bwd_stride when using the SYCL API from the Intel oneMKL library.

src/dft/backends/mklgpu/commit.cpp
Rbiessy: We've agreed this PR is the best solution for now so resolving this comment.

Rbiessy (approved):

hjabird:
I'll remove the support for 2024.0.

In the meantime, it looks like I've somehow broken my implementation, so please hold off reviewing!

hjabird:
I've updated the PR to remove support for earlier than 2024.1, and also to sort out an issue with exception handling. Test results are here: [testres.txt](https://github.com/user-attachments/files/15891652/testres.txt)

*Q: How come it was working earlier and then you said it was broken?*
A: I previously fixed this issue, giving the results earlier, but failed push the relavent commit. Then I deleted it locally and and had to redo the fix all over.

hjabird:
@lhuot are you happy with this PR in its revised form?

lhuot (approved):
LGTM, thanks for the patch!

PR #515, opened 2024-06-17
Fix trademark issues

nmishra31:
**Background:**
There were trademark issues found in these README files, and this PR changes all these issues so that there are no trademark problems.


README.md
```diff
@@ -151,10 +151,10 @@ Header-based and backend-independent Device API can be called within ```sycl ker
 
 ### Supported Configurations:
 
-Supported domains include: BLAS, LAPACK, RNG, DFT, SPARSE_BLAS
+Supported domains include: BLAS, LAPACK, RNG, DFT, SPARSE_BLASS
```
hjabird: SPARSE_BLASS doesn't look right!

examples/README.md
```diff
@@ -46,7 +46,7 @@ $ ./bin/example_blas_gemm_usm
 ########################################################################
 
 Running BLAS GEMM USM example on CPU device.
-Device name is: Intel(R) Core(TM) i7-6770HQ CPU @ 2.60GHz
+Device name is: Intel(R) Core(TM) i7-6770HQ Processor @ 2.60GHz
```
Rbiessy: This is meant to be the output of some examples running. The examples may output "CPU" depending on the device used. If this is an issue should we consider not copying the example outputs here?

README.md
nmishra31: Thanks for the catch, fixed the typo in [f1decce](https://github.com/oneapi-src/oneMKL/pull/515/commits/f1decce5cee5c4bba59a62645d1763aa50d62e68)

examples/README.md
nmishra31: This is a good point, and after taking a look I agree on not copying the outputs. I've deleted the outputs and kept the relevant showcases of how to actually run the examples. Since there is quite a lot of repetition/overlap in showing how to run examples, I only kept the showcases of one domain to decrease clutter as I didn't see anything specific that would be lost in this. I adjusted the preface text to match these changes. Let me know if something looks off / is missing

Rbiessy (approved):

examples/README.md
```diff
@@ -17,582 +17,25 @@ The example executable naming convention follows `example_<$domain>_<$routine>_<
   or `example_<$domain>_<$routine>` for run-time dispatching examples. 
   E.g. `example_blas_gemm_usm_mklcpu_cublas `  `example_blas_gemm_usm`
 
-## Example outputs (blas, rng, lapack, dft, sparse_blas)
+## Running examples (blas)
```
mkrainiuk: I don't think we need blas in the subtitle
```suggestion
## Running examples
```

mkrainiuk (approved):
Looks good to me with one minor comment

dnhsieh-intel:
I would suggest NOT changing the landing README file. The changes would make the table hard to interpret and inconsistent.

itopinsk:
@nmishra31 I'd like to suggest split this PR on two parts: (1) fix trade mark issues and (2) any further changes. So (1) can be merged and the (2) with the commit [Remove example output and make README more concise](https://github.com/uxlfoundation/oneMath/pull/515/commits/8bbb9d2279c27da891013d5b903572f08c8b8225) deserves a separate review.

dnhsieh-intel:
Closing since it has been resolved internally.

PR #523, opened 2024-06-28
[DFT][CMake] Fix inclusion of public include directories when linking against a DFT backend library

### LINKED ISSUE #442: [CMake][DFT] target_link_library(myTarget PRIVATE MKL::onemkl_dft_XXX) does not add includes
# Summary

The oneMKL interface library supports linking the the DFT backends (to allow for compile-time dispatch) instead of linking to the main library itself and using the dynamic dispatch mechanism. There are examples of using oneMKL in this way in the examples.

However, the CMake in the examples does is not representative of oneMKL usage after installation.

# Observed behavior

Using the example of the MKLGPU backend, I'd expect to the following:
```
        target_link_libraries(libgromacs PRIVATE MKL::onemkl_dft_mklgpu)
```
to also include the MKL header files. e.g. I'd expect to able to `#include <oneapi/mkl/dft.hpp>`. However, without additional cmake, this header is not available.

# Expected behavior

I think the DFT backends targets should include the relevent headers as part of their interface.

# Version
Report oneMKL version: develop tip: [f5dc527](https://github.com/oneapi-src/oneMKL/commit/f5dc52762a042fbfb5f48f2e01b81463ae9b05f8)

# Environment
DPC++ 2024.0
Ubuntu20

# Steps to reproduce

CMake:
```
cmake_minimum_required(VERSION 3.26)
project(exampleProject)
add_executable(myapp main.cpp)
find_package(oneMKL REQUIRED)
# target_link_libraries(myapp PRIVATE MKL::onemkl)
target_link_libraries(myapp PRIVATE MKL::onemkl_dft_mklgpu)
```

C++
```
#include <oneapi/mkl/dft.hpp>
int main(){ return 0; }
```
Gives:
```
FAILED: CMakeFiles/myapp.dir/main.cpp.o
/opt/intel/oneapi/compiler/2024.0/bin/icpx   -fsycl -MD -MT CMakeFiles/myapp.dir/main.cpp.o -MF CMakeFiles/myapp.dir/main.cpp.o.d -o CMakeFiles/myapp.dir/main.cpp.o -c /home/hugh/tmp/example/main.cpp
/home/hugh/tmp/example/main.cpp:1:10: fatal error: 'oneapi/mkl/dft.hpp' file not found
    1 | #include <oneapi/mkl/dft.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~
1 error generated.
```

hjabird:
# Description

This PR ensures that linking against backend such as `target_link_library(myApp PRIVATE onemkl_dft_cufft` correctly includes the oneMKL public include directories.

Fixes https://github.com/oneapi-src/oneMKL/issues/442

# Checklist

## All Submissions

  - [onemklinteldft.txt](https://github.com/user-attachments/files/16065255/onemklinteldft.txt)
  - [onemklrocmdft.txt](https://github.com/user-attachments/files/16065258/onemklrocmdft.txt)
  - [onemklcufftdft.txt](https://github.com/user-attachments/files/16065259/onemklcufftdft.txt)
  - [onemklcudadft_rebase.txt](https://github.com/user-attachments/files/16066516/onemklcudadft_rebase.txt) for after rebase, including CT example with MKLCPU DFT and CUFFT.
  - Tests build and run with *portFFT*, but some test cases failed. The changes in this PR should not be related to these problems.

## Bug fixes

- [ish] Have you added relevant regression tests?
  - The compile-time DFT example has been modified to depend on the described functionality.
      GitHub issue or in this PR)?

### DIFF
--- a/examples/dft/compile_time_dispatching/CMakeLists.txt
+++ b/examples/dft/compile_time_dispatching/CMakeLists.txt
@@ -30,19 +30,17 @@ foreach(dft_ct_source ${DFT_CT_SOURCES})
   add_executable(${EXAMPLE_NAME} ${dft_ct_source}.cpp)
   target_include_directories(${EXAMPLE_NAME}
       PUBLIC ${PROJECT_SOURCE_DIR}/examples/include
-      PUBLIC ${PROJECT_SOURCE_DIR}/include
       PUBLIC ${CMAKE_BINARY_DIR}/bin
   )
 
-if(domain STREQUAL "dft" AND ENABLE_MKLCPU_BACKEND AND ENABLE_CUFFT_BACKEND)
-  add_dependencies(${EXAMPLE_NAME} onemkl_${domain}_mklcpu onemkl_${domain}_cufft)
-  list(APPEND ONEMKL_LIBRARIES_${domain} onemkl_${domain}_mklcpu onemkl_${domain}_cufft)
-endif()
+  if(domain STREQUAL "dft" AND ENABLE_MKLCPU_BACKEND AND ENABLE_CUFFT_BACKEND)
+    add_dependencies(${EXAMPLE_NAME} onemkl_${domain}_mklcpu onemkl_${domain}_cufft)
+    list(APPEND ONEMKL_LIBRARIES_${domain} onemkl_${domain}_mklcpu onemkl_${domain}_cufft)
+  endif()
 
-target_link_libraries(${EXAMPLE_NAME} PUBLIC
-  ${ONEMKL_LIBRARIES_${domain}}
-  ONEMKL::SYCL::SYCL
-  onemkl_warnings
+  target_link_libraries(${EXAMPLE_NAME} PUBLIC
+    ${ONEMKL_LIBRARIES_${domain}}
+    onemkl_warnings
   )
 
   # Register example as ctest
--- a/src/dft/backends/cufft/CMakeLists.txt
+++ b/src/dft/backends/cufft/CMakeLists.txt
@@ -34,6 +34,9 @@ add_dependencies(onemkl_backend_libs_dft ${LIB_NAME})
 target_include_directories(${LIB_OBJ}
   PUBLIC ${ONEMKL_INTERFACE_INCLUDE_DIRS}
 )
+target_include_directories(${LIB_NAME}
+  PUBLIC ${ONEMKL_INTERFACE_INCLUDE_DIRS}
+)
 target_include_directories(${LIB_OBJ}
   PRIVATE ${PROJECT_SOURCE_DIR}/src
           ${CMAKE_BINARY_DIR}/bin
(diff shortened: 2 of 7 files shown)


src/dft/backends/cufft/CMakeLists.txt
```diff
@@ -32,8 +32,13 @@ add_library(${LIB_OBJ} OBJECT
 add_dependencies(onemkl_backend_libs_dft ${LIB_NAME})
 
 target_include_directories(${LIB_OBJ}
-  PRIVATE ${PROJECT_SOURCE_DIR}/include
-          ${PROJECT_SOURCE_DIR}/src
+  PUBLIC ${ONEMKL_INTERFACE_INCLUDE_DIRS}
+)
+target_include_directories(${LIB_NAME}
+  PUBLIC ${ONEMKL_INTERFACE_INCLUDE_DIRS}
+)
```
hjabird, line 39: Both of these are required, but I don't understand why. Later on there is `target_link_libraries(${LIB_NAME} PUBLIC ${LIB_OBJ})` which should make the second of these redundant.

If anyone could clear up my confusion, I'd be greatful!

Rbiessy: I don't get why this doesn't work either. I wonder if it can be related to the fact that `${LIB_OBJ}` is an `OBJECT` target which maybe are not be meant to have an include dir property?
Some of this is discussed there: https://stackoverflow.com/questions/38832528/transitive-target-include-directories-on-object-libraries.

hjabird: That seems like a reasonable explanation. If anyone thinks there is a better way of doing this, they can open a new comment.

Rbiessy (approved):

hjabird:
This affects only the build, so pinging @vmalia.

hjabird:
I'm told I pinged the wrong person - sorry vmalia! @mkrainiuk

examples/dft/compile_time_dispatching/CMakeLists.txt
```diff
@@ -30,19 +30,17 @@ foreach(dft_ct_source ${DFT_CT_SOURCES})
   add_executable(${EXAMPLE_NAME} ${dft_ct_source}.cpp)
   target_include_directories(${EXAMPLE_NAME}
       PUBLIC ${PROJECT_SOURCE_DIR}/examples/include
-      PUBLIC ${PROJECT_SOURCE_DIR}/include
```
mkrainiuk, line 33: Any specific reasons for removing `include`  from this target?

mkrainiuk (approved):
Looks good to me, thank you! I have one minor question.

examples/dft/compile_time_dispatching/CMakeLists.txt
hjabird: Yes.

If the backend target (eg. `onemkl_dft_cufft`) is configured correctly, then linking against the backend target using `target_link_libraries` should bring in this include directory anyway. The fact that it doesn't do this is the motivation for this PR - otherwise the gymnastics [like here](https://gitlab.com/gromacs/gromacs/-/blob/main/src/gromacs/CMakeLists.txt?ref_type=heads#L252) are required.

mkrainiuk: Thanks for details.

PR #525, opened 2024-07-04
[BLAS][portBLAS] Add try/catch for portblas runtime exception & minor fix

s-Nick:
# Description

This PR adds exception handling to portBLAS backend implementation. This would allow us to handle corner cases in portBLAS.

It also adds a minor fix to `omatcopy2` tests adding check if double precision is supported by the device before running tests.

# Checklist

## All Submissions

[intel_arc_log.txt](https://github.com/user-attachments/files/16096302/intel_arc_log.txt)

### DIFF
--- a/src/blas/backends/portblas/portblas_common.hpp
+++ b/src/blas/backends/portblas/portblas_common.hpp
@@ -202,7 +202,7 @@ struct throw_if_unsupported_by_device {
         try {                                                                                   \
             std::apply(fn, args);                                                               \
         }                                                                                       \
-        catch (const std::runtime_error& e) {                                                   \
+        catch (const ::blas::unsupported_exception& e) {                                        \
             throw unimplemented("blas", e.what());                                              \
         }                                                                                       \
     }                                                                                           \
@@ -223,7 +223,7 @@ struct throw_if_unsupported_by_device {
         try {                                                                     \
             return std::apply(fn, args);                                          \
         }                                                                         \
-        catch (const std::runtime_error& e) {                                     \
+        catch (const ::blas::unsupported_exception& e) {                          \
             throw unimplemented("blas", e.what());                                \
         }                                                                         \
     }                                                                             \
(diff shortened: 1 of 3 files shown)


src/blas/backends/portblas/portblas_common.hpp
```diff
@@ -199,7 +199,12 @@ struct throw_if_unsupported_by_device {
         auto fn = [](auto&&... targs) {                                                         \
             portBLASFunc(std::forward<decltype(targs)>(targs)...);                              \
         };                                                                                      \
-        std::apply(fn, args);                                                                   \
+        try {                                                                                   \
+            std::apply(fn, args);                                                               \
+        }                                                                                       \
+        catch (const std::runtime_error& e) {                                                   \
```
hjabird: I wonder if this is overly broad? I think any synchronous exceptions from the SYCL runtime are going to get rethrown as `unimplemented` which is misleading.

tests/unit_tests/blas/extensions/omatcopy2.cpp
```diff
@@ -185,6 +187,8 @@ TEST_P(Omatcopy2Tests, ComplexSinglePrecision) {
 }
 
 TEST_P(Omatcopy2Tests, ComplexDoublePrecision) {
+    CHECK_DOUBLE_ON_DEVICE(std::get<0>(GetParam()));
```
hjabird, line 190: Why are these needed only now?

src/blas/backends/portblas/portblas_common.hpp
Rbiessy: My thoughts too, I think we should only catch `blas::unsupported_exception` once https://github.com/codeplaysoftware/portBLAS/pull/524 is merged.

tests/unit_tests/blas/extensions/omatcopy2.cpp
s-Nick: This has always been necessary, but it was missing. It skips the test if the device doesn't support double precision otherwise the test fails. It wasn't catch before because tests always ran on devices that supports it.

src/blas/backends/portblas/portblas_common.hpp
s-Nick: Thank you both for the review. I agree and I fixed in [6b259b8](https://github.com/oneapi-src/oneMKL/pull/525/commits/6b259b84e3e7603a31f061ef28a26f2d289f7a46)

Rbiessy (approved):

hjabird (approved):

PR #528, opened 2024-07-05
[DFT] Add FWD/BWD_STRIDES to public API, deprecate INPUT/OUTPUT_STRIDES

### LINKED ISSUE #488: [DFT] Use FWD_ and BWD_STRIDE API
# Summary

oneMKL should support the FWD and BWD stride interface in the updated oneMKL specification

# Problem statement
The updated oneMKL specification (https://spec.oneapi.io/versions/latest/elements/oneMKL/source/domains/dft/enums_and_config_params.html) deprecates INPUT_STRIDES and OUTPUT_STRIDES and replaces them with FWD_STRIDES and BWD_STRIDES. oneMKL interace library should implment this.

# Preferred solution
Implement support for FWD and BWD strides, and add a deprecation warning to the old interface.

hjabird:
# Description

* Adds the FWD/BWD_STRIDES API from the interface spec
* Deprecates old INPUT/OUTPUT_STRIDES using the deprecated attribute.
* Modifies DFT tests to use FWD/BWD_STRIDES API
  * The old API is no longer thoroughly tested
* Adds additional descriptor tests

Fixes https://github.com/oneapi-src/oneMKL/issues/488

# Checklist

## All Submissions

  * [amdtest.txt](https://github.com/user-attachments/files/16113136/amdtest.txt)
  * [intel.txt](https://github.com/user-attachments/files/16113137/intel.txt)
  * [nvidiadft.txt](https://github.com/user-attachments/files/16113138/nvidiadft.txt)
  * [portfft2.txt](https://github.com/user-attachments/files/16113135/portfft2.txt)
    * PortFFT has failures that also exist in develop.

## New interfaces

- [x] What version of oneAPI spec the interface is targeted?
  - Feature from https://github.com/uxlfoundation/oneAPI-spec/commit/3717a9a2dfd162e91b5a5065df5e9411ee9b7de7

## New features

  * Tests updated to use new functionality

### DIFF
--- a/src/dft/backends/cufft/commit.cpp
+++ b/src/dft/backends/cufft/commit.cpp
@@ -34,6 +34,8 @@
 #include "oneapi/mkl/dft/detail/cufft/onemkl_dft_cufft.hpp"
 #include "oneapi/mkl/dft/types.hpp"
 
+#include "../stride_helper.hpp"
+
 #include <cufft.h>
 #include <cuda.h>
 
@@ -50,7 +52,7 @@ private:
     // We also need this because oneMKL uses a directionless "FWD_DISTANCE" and "BWD_DISTANCE" while cuFFT uses a directional "idist" and "odist".
     // plans[0] is forward, plans[1] is backward
     std::array<std::optional<cufftHandle>, 2> plans = { std::nullopt, std::nullopt };
-    std::array<std::int64_t, 2> offsets;
+    std::int64_t offset_fwd_in, offset_fwd_out, offset_bwd_in, offset_bwd_out;
 
 public:
     cufft_commit(sycl::queue& queue, const dft::detail::dft_values<prec, dom>& config_values)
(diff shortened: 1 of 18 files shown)


hjabird:
* Pinging @lhuot for DFT changes.
* I'll be addressing comments from the week of the 15th.

Rbiessy (approved):
Looks good to me! The strides in cuFFT are pretty difficult to get its head around but that was already the case before... We have pretty good tests so this should be fine!

anantsrivastava30 (commented):
Reviewed the MKLGPU/MKLCPU part of addition and deprication, looks good to me, only thing I was mulling over was the get_strides_api, if in/out_strides are set then fwd/bwd_strides (config_value) must be zero,  (which is a internal property of the DftiSetValue (in cpu codebase), set_value(gpu codebase)). But here you are checking config_value.fwd_strides is zero or not! Are we explicitly setting fwd_strides to zero in this PR when we set input_strides or vice-versa.

anantsrivastava30 (commented):
Approved, can have this https://github.com/oneapi-src/oneMKL/pull/528#pullrequestreview-2183590984 discussing without blocking.

anantsrivastava30 (approved):
Approved, can have this https://github.com/oneapi-src/oneMKL/pull/528#pullrequestreview-2183590984 discussing without blocking.

hjabird:
> Reviewed the MKLGPU/MKLCPU part of addition and deprication, looks good to me, only thing I was mulling over was the get_strides_api, if in/out_strides are set then fwd/bwd_strides (config_value) must be zero, (which is a internal property of the DftiSetValue (in cpu codebase), set_value(gpu codebase)). But here you are checking config_value.fwd_strides is zero or not! Are we explicitly setting fwd_strides to zero in this PR when we set input_strides or vice-versa.

* In this PR we [set fwd/bwd_strides](https://github.com/hjabird/oneMKL/blob/2bab1d529a58d3d75540e3cd56aa0a3727ecd7e9/src/dft/descriptor_config_helper.hpp#L183) to zero on setting input/output_strides, and [vice versa](https://github.com/hjabird/oneMKL/blob/2bab1d529a58d3d75540e3cd56aa0a3727ecd7e9/src/dft/descriptor_config_helper.hpp#L245).
  * [Tested here](https://github.com/hjabird/oneMKL/blob/2bab1d529a58d3d75540e3cd56aa0a3727ecd7e9/tests/unit_tests/dft/source/descriptor_tests.cpp#L212) and [here](https://github.com/hjabird/oneMKL/blob/2bab1d529a58d3d75540e3cd56aa0a3727ecd7e9/tests/unit_tests/dft/source/descriptor_tests.cpp#L150).
* We check if `config_value.fwd_strides` is zero [when trying to determine](https://github.com/hjabird/oneMKL/blob/2bab1d529a58d3d75540e3cd56aa0a3727ecd7e9/src/dft/backends/stride_helper.hpp#L121) which stride API was used. 
  * If no strides have been set, the input/output strides and fwd/bwd strides will contain identical values, so it doesn't matter which set of strides gets used.
  * If the input/output API was used, then these will have been zeroed. It will therefore test for the usage of the input/output stride API.

src/dft/descriptor.cxx
```diff
@@ -30,16 +30,19 @@ namespace detail {
...
         strides[i] = strides[i + 1] * dimensions[i];
     }
     strides[0] = 0;
-    output_strides = strides;
-    input_strides = std::move(strides);
+    // Fwd/Bwd strides and Input/Output strides being the same by default means
+    // that we don't have to specify if we default to using fwd/bwd strides or
+    // input/output strides.
+    bwd_strides = strides;
+    fwd_strides = std::move(strides);
```
raphael-egan, line 45: It looks like the default strides are identical for real and complex descriptors. These values are adequate default for complex descriptors only. Am I missing something?
If my suspicion is correct, this implementation is not aligned with [the specification](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/elements/onemkl/source/domains/dft/config_params/data_layouts#onemkl-dft-fwd-bwd-strides).
Note: having different default strides for real descriptors was one of the main motivations for deprecating i/o strides and substituting them with fwd/bwd strides, instead.

src/dft/descriptor.cxx
```diff
@@ -69,10 +72,15 @@ void descriptor<prec, dom>::set_value(config_param param, ...) {
...
+        case config_param::OUTPUT_STRIDES:
             detail::set_value<config_param::OUTPUT_STRIDES>(values_, va_arg(vl, std::int64_t*));
             break;
-        }
+        case config_param::FWD_STRIDES:
+            detail::set_value<config_param::FWD_STRIDES>(values_, va_arg(vl, std::int64_t*));
+            break;
+        case config_param::BWD_STRIDES:
+            detail::set_value<config_param::BWD_STRIDES>(values_, va_arg(vl, std::int64_t*));
+            break;
```
raphael-egan, line 83: Note: the `set_value` method is expecting an `std::vector<std::int64_t>` type for `(INPUT|OUTPUT|FWD|BWD)_STRIDES` according to the table in this part of the [specification](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/elements/onemkl/source/domains/dft/enums_and_config_params#onemkl-dft-enum-config-param).
That being said, non-trivial types can't be passed through variadic functions in fact, and addressing this discrepancy would require changing the `(s|g)et_value` parts of the specs. I will very likely bring up a proposal for that, in the near future but I don't know what's best to do about this, in the meantime...

src/dft/descriptor.cxx
```diff
@@ -142,8 +150,10 @@ descriptor<prec, dom>::descriptor(std::vector<std::int64_t> dimensions) {
                                         "Invalid dimension value (negative or 0).");
         }
     }
+    compute_default_strides(dimensions, values_.fwd_strides, values_.bwd_strides);
     // Assume forward transform.
-    compute_default_strides(dimensions, values_.input_strides, values_.output_strides);
+    values_.input_strides = values_.fwd_strides;
+    values_.output_strides = values_.bwd_strides;
```
raphael-egan, line 156: Since the default strides being set at construction are the fwd and bwd ones, wouldn't it make more sense to zero-initialize the input and output strides then? I believe a user invoking
`descriptor.get_value(config_param::INPUT_STRIDES, user_pointer_i);`
`descriptor.get_value(config_param::OUTPUT_STRIDES, user_pointer_o);`
`descriptor.get_value(config_param::FWD_STRIDES, user_pointer_fwd);`
`descriptor.get_value(config_param::BWD_STRIDES, user_pointer_bwd);`
immediately after construction would observe non-zero values in all cases which may seem slightly confusing (borderline contradictory) with the [second note](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/elements/onemkl/source/domains/dft/config_params/data_layouts#onemkl-dft-fwd-bwd-strides) in this section of the specification.

Also, is the `// Assume forward transform.` comment still relevant in the new paradigm?

src/dft/descriptor.cxx
```diff
@@ -218,6 +228,14 @@ void descriptor<prec, dom>::get_value(config_param param, ...) const {
...
                       va_arg(vl, std::int64_t*));
             break;
+        case config_param::FWD_STRIDES:
+            std::copy(values_.fwd_strides.begin(), values_.fwd_strides.end(),
+                      va_arg(vl, std::int64_t*));
+            break;
+        case config_param::BWD_STRIDES:
+            std::copy(values_.bwd_strides.begin(), values_.bwd_strides.end(),
+                      va_arg(vl, std::int64_t*));
+            break;
```
raphael-egan, line 238: Similar note here: the get_value method is expecting an `std::vector<std::int64_t>*` type for `(INPUT|OUTPUT|FWD|BWD)_STRIDES` according to the table in this part of the [specification](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/elements/onemkl/source/domains/dft/enums_and_config_params#onemkl-dft-enum-config-param).
In this case though, I believe such a pointer _can_ be passed through variadic functions so this difference with the specification might bring some serious confusion to users...

src/dft/backends/mklcpu/mklcpu_helpers.hpp
```diff
@@ -74,8 +74,6 @@ inline constexpr DFTI_CONFIG_PARAM to_mklcpu(dft::detail::config_param param) {
         case iparam::COMPLEX_STORAGE: return DFTI_COMPLEX_STORAGE;
         case iparam::REAL_STORAGE: return DFTI_REAL_STORAGE;
         case iparam::CONJUGATE_EVEN_STORAGE: return DFTI_CONJUGATE_EVEN_STORAGE;
-        case iparam::INPUT_STRIDES: return DFTI_INPUT_STRIDES;
-        case iparam::OUTPUT_STRIDES: return DFTI_OUTPUT_STRIDES;
```
raphael-egan, line 78: Was this used anywhere?

src/dft/backends/rocfft/commit.cpp
```diff
@@ -274,31 +275,40 @@ class rocfft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+        std::array<std::size_t, 3> stride_b_indices{ 0, 1, 2 };
+        std::sort(&stride_b_indices[0], &stride_b_indices[dimensions],
+                  [&](std::size_t a, std::size_t b) {
+                      return stride_vecs.vec_b[a] < stride_vecs.vec_b[b];
+                  });
         std::array<std::size_t, max_supported_dims> lengths_cplx = lengths;
         if (dom == dft::domain::REAL) {
             lengths_cplx[0] = lengths_cplx[0] / 2 + 1;
         }
         // When creating real-complex descriptions, the strides will always be wrong for one of the directions.
```
raphael-egan, line 292: The goal of the newly-introduced configuration parameters was to invalidate this kind of comment.

src/dft/backends/rocfft/commit.cpp
```diff
@@ -274,31 +275,40 @@ class rocfft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+                  stride_vecs.vec_a[stride_a_indices[2]]));
+        const bool vec_b_valid_as_reals =
             dimensions == 1 ||
-            (lengths_cplx[out_stride_indices[0]] <= out_strides[out_stride_indices[1]] &&
+            (lengths_cplx[stride_b_indices[0]] <= stride_vecs.vec_b[stride_b_indices[1]] &&
              (dimensions == 2 ||
-              lengths_cplx[out_stride_indices[0]] * lengths_cplx[out_stride_indices[1]] <=
-                  out_strides[out_stride_indices[2]]));
+              lengths_cplx[stride_b_indices[0]] * lengths_cplx[stride_b_indices[1]] <=
+                  stride_vecs.vec_b[stride_b_indices[2]]));
```
raphael-egan, line 306: Is there no similar requirement as the above for the complex domain? Say self-explanatory flags like `vec_(a|b)_valid_as_hermitians` (to use rocfft's terms)?

src/dft/backends/rocfft/commit.cpp
```diff
@@ -366,10 +376,10 @@ class rocfft_commit final : public dft::detail::commit_impl<prec, dom> {
                                                         nullptr, // in offsets
```
raphael-egan, line 376: Can `plan_desc` be safely re-used here if it was already used above in `valid_forward`? No memory leak to worry about for instance?

src/dft/backends/cufft/commit.cpp
```diff
@@ -147,126 +149,107 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
         else {
-            if (config_values.placement == config_value::INPLACE) {
-                onembed = inembed;
-            }
-            else if (o_min - onembed.begin() != i_min - inembed.begin()) {
+            if (a_min - stride_vecs.vec_a.begin() != b_min - stride_vecs.vec_b.begin()) {
                 throw mkl::unimplemented(
                     "dft/backends/cufft", __FUNCTION__,
                     "cufft requires that if ordered by stride length, the order of strides is the same for input and output strides!");
             }
```
raphael-egan, line 177: Maybe we should change the message to avoid referring to input/output strides explicitly? Just a suggestion...

Also, is the check implementation sufficient w.r.t. the detail captured in the exception message? For instance, wouldn't `{FWD|INPUT}_STRIDES <- {0, 1024, 32, 1}` and `{BWD|OUTPUT}_STRIDES <- {0, 32, 1024, 1}` miss the check?
Edit: it looks like other checks thereafter catch that kind of possible case, could you confirm?

src/dft/backends/cufft/commit.cpp
```diff
@@ -147,126 +149,107 @@ class cufft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+            int inner_sfwd = strides_fwd.back(); // inner strides of DFT
+            int inner_sbwd = strides_bwd.back();
+            bool valid = true;
+            for (int r = 1; r < rank; ++r) {
+                valid = valid && (inner_nfwd <= inner_sfwd) && (inner_nbwd <= inner_sbwd);
+                inner_nfwd *= n_copy[rank - r - 1];
+                inner_nbwd *= n_copy[rank - r - 1];
+                inner_sfwd *= strides_fwd[rank - r - 1];
+                inner_sbwd *= strides_bwd[rank - r - 1];
             }
```
raphael-egan, line 245: Does cuFFT support negative stride values? If yes, wouldn't these checks be invalid with negative strides?

raphael-egan (commented):
My review is coming a little late but I'd appreciate it if one could give it consideration.

Rbiessy:
I am so sorry @raphael-egan I forgot you were also planning to review it! I merged it as Hugh was going on holiday. I am sure he will be able to address your comments once he is back.

raphael-egan:
> I am so sorry @raphael-egan I forgot you were also planning to review it! I merged it as Hugh was going on holiday. I am sure he will be able to address your comments once he is back.

No problem, I'm sure he will.

src/dft/descriptor.cxx
hjabird: I missed that bit of the spec, so I will open a new PR to address this and other corrections / improvements.

src/dft/descriptor.cxx
hjabird: You're right here. The DFT domain has always used `*std::int64_t` everywhere as a result instead due to the constraints around variadic functions.

I think the variadic function API is error-prone at the best of times. If you're enthusiastic about improving the spec, can I suggest a template based API along the lines of:
```
template<enum ConfigParam>
void set_value(config_param_t<ConfigParam>& value);
```
such that argument types could be correctly enforced by the compiler. An easy mistake with the current setup is giving your strides as  *int, which doesn't produce an error I think. Or `LENGTHS` where the return type depends on the rank of the DFT.

src/dft/backends/mklcpu/mklcpu_helpers.hpp
hjabird: I don't think so, and removing this didn't break anything.

Some similar mappings are used, but several are unused. To me it made more sense to remove unneeded code than add more mappings that wouldn't be used.

src/dft/backends/rocfft/commit.cpp
hjabird: I find this code confusing, and it looks like I got it wrong the first time round with my understanding of it. I tried to modify the code such that it was the same as what already existed.

* `lengths_cplx` look like the storage required for the DFT in the complex domain.
* So we're checking if strides vs `lengths_complex`, we're checking that the strides are sufficient for writing the data in the complex domain.

To conclude, I've named this variable badly and misleadingly. It is valid for both real and complex domain DFTs. I will rename this in a new PR.

src/dft/backends/rocfft/commit.cpp
hjabird: I think so:
>The plan description object can be safely destroyed after it is given to the `rocfft_plan_create()`

https://rocm.docs.amd.com/projects/rocFFT/en/latest/reference/api.html#plan-description

src/dft/backends/cufft/commit.cpp
hjabird: I *think* this is correct.

For CUFFT, `strides_fwd` and `strides_bwd` correspond to `inembed` and `onembed`, [where](https://docs.nvidia.com/cuda/cufft/index.html#advanced-data-layout)
>`n[i] ≤ inembed[i]`, `n[i] ≤ onembed[i]`

The strides can be negative, if you set `a_stride` and `b_stride` negative.

raphael-egan: I see. So this particular logic is likely adequate w.r.t. the `(i|o)nembed` values but there are other bits (determining the `(i|o)nembed`) that might need revision before we reach this point then, in my opinion. For instance, `a_min` and `b_min` would be the "most negative" value as of now (not the minimal stride in _absolute_ value) which is not what we'd want in this case, unless I'm wrong. This is rather orthogonal to the task targeted herein though, I just wanted to point out a possible concern.

src/dft/backends/rocfft/commit.cpp
raphael-egan: Sorry, I didn't use the clearest language in my comment, I meant "complex domain" as in "backward domain of real transforms" in this case.

src/dft/descriptor.cxx
raphael-egan: Thanks!

hjabird: New PR at: https://github.com/oneapi-src/oneMKL/pull/549
This change is implemented at: https://github.com/hjabird/oneMKL/blob/128ef68cd2ed35c6543ca163a8d79bd69b678748/src/dft/descriptor.cxx#L33

src/dft/descriptor.cxx
hjabird: New PR at: https://github.com/oneapi-src/oneMKL/pull/549
This change is implemented at: https://github.com/hjabird/oneMKL/blob/128ef68cd2ed35c6543ca163a8d79bd69b678748/src/dft/descriptor.cxx#L164

src/dft/backends/rocfft/commit.cpp
hjabird: I've updated the code to check both fwd and bwd domains in the new PR: https://github.com/oneapi-src/oneMKL/pull/549
Change: https://github.com/hjabird/oneMKL/blob/128ef68cd2ed35c6543ca163a8d79bd69b678748/src/dft/backends/rocfft/commit.cpp#L304

src/dft/backends/rocfft/commit.cpp
hjabird: Clarified in new PR: https://github.com/oneapi-src/oneMKL/pull/549
Change: https://github.com/hjabird/oneMKL/blob/128ef68cd2ed35c6543ca163a8d79bd69b678748/src/dft/backends/rocfft/commit.cpp#L292

src/dft/descriptor.cxx
hjabird: I haven't updated this is the new PR because:
* We were non-spec-conformant in the same way with the I/O strides.
  * Users moving to FWD/BWD by just changing the config parameter name would have issues.
* For the `set_value` side, the spec is broken. I think its nicer if set/get value use the same parameter type.
* A much much better way of fixing this mess is to adopt your suggested updated type-safe spec for this.

raphael-egan: Agreed! That proposal will come in the hopefully near future.

PR #533, opened 2024-07-11
[RNG] Added -fsycl -fsycl-device-code-split=per_kernel at the linking stage to have device code

egrabovskaya:
# Description

We must pass `-fsycl` flag at the linking stage in addition to the compilation stage so that device code is added to executable file from object file on Windows.
Otherwise, there is the failure in Device API tests and example: **No kernel named _ZTSZZN12moments_testIN6oneapi3mkl3rng6device13philox4x32x10ILi1EEENS3_7uniformIfNS3_14uniform_method8standardEEEEclIN4sycl3_V15queueEEEvT_ENKUlRNSD_7handlerEE_clESH_EUlNSD_4itemILi1ELb1EEEE_ was found -46 (PI_ERROR_INVALID_KERNEL_NAME)**

Fixes # (GitHub issue)

# Checklist

## All Submissions

[logs_win_static.txt](https://github.com/user-attachments/files/16282892/logs_win_static.txt)

## New interfaces

it was accepted? # (RFC)

## New features

## Bug fixes

      GitHub issue or in this PR)?

### DIFF
--- a/examples/rng/device/CMakeLists.txt
+++ b/examples/rng/device/CMakeLists.txt
@@ -60,6 +60,7 @@ foreach(rng_device_source ${RNG_DEVICE_SOURCES})
   )
 
   if(is_dpcpp)
+    message("is_dpcpp in examples/rng/device")
     target_link_options(example_${domain}_${rng_device_source} PUBLIC -fsycl -fsycl-device-code-split=per_kernel)
   endif()
 
--- a/examples/rng/device/CMakeLists.txt
+++ b/examples/rng/device/CMakeLists.txt
@@ -59,7 +59,7 @@ foreach(rng_device_source ${RNG_DEVICE_SOURCES})
       ONEMKL::SYCL::SYCL
   )
 
-  if(NOT ${ONEMKL_SYCL_IMPLEMENTATION} STREQUAL "hipsycl")
+  if(is_dpcpp)
     target_link_options(example_${domain}_${rng_device_source} PUBLIC -fsycl -fsycl-device-code-split=per_kernel)
   endif()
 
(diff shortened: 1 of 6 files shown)


examples/rng/device/CMakeLists.txt
```diff
@@ -59,6 +59,10 @@ foreach(rng_device_source ${RNG_DEVICE_SOURCES})
       ONEMKL::SYCL::SYCL
   )
 
+  if(NOT ${ONEMKL_SYCL_IMPLEMENTATION} STREQUAL "hipsycl")
```
Rbiessy, line 62: I think it would be more robust to use `is_dpcpp` in these `if` statements. That variable is set if `-fsycl` is supported which seems to be what you want here.

Rbiessy (approved):

examples/rng/device/CMakeLists.txt
egrabovskaya: Hi Romain,
Unfortunately, `is_dpcpp` flag value is empty on windows. `check_cxx_compiler_flag("-fsycl" is_dpcpp)` fails and sets nothing. We're investigating why it's so. It may be a compiler bug. That's why I reverted these changes otherwise there will be failures on windows again

Rbiessy: I see, that's not a bug `find_package(Compiler)` is never called on Windows unfortunately: https://github.com/oneapi-src/oneMKL/blob/develop/CMakeLists.txt#L270

andreyfe1 (approved):

aelizaro (approved):

PR #532, opened 2024-07-10
[HIP] Fix libclang_rt.builtins.a not found

Rbiessy:
# Description

Using the compiler from oneAPI's Base Toolkit and the [Codeplay AMD plugin](https://developer.codeplay.com/products/oneapi/amd/home/) fails to compile by default with the error:
```
ninja: error: '/path/to/intel/oneapi/2024.2.0/compiler/2024.2/lib/clang/19/lib/x86_64-unknown-linux-gnu/libclang_rt.builtins.a', needed by 'lib/libonemkl_sparse_blas_rocsparse.so.0', missing and no known rule to make it
```
As far as I know this is an issue for all ROC backends. The issue is that the command `find_package(HIP)` invokes `${HIP_CXX_COMPILER} -print-libgcc-file-name --rtlib=compiler-rt` to find the path of a dependency where `HIP_CXX_COMPILER` is set as `CMAKE_CXX_COMPILER` by default. This command is meant to be run with HIP's `clang++` but when `icpx` is used the command is outputting a non-existant file.

As the documentation suggests this can be worked around by building intel/llvm which is not user friendly. The second workaround of setting `LIBRARY_PATH` did not work for me using ROCm 6.1.0 so I am suggesting to remove it.
The suggested solution should be more resilient and does not need any more action from the user as the environment variables `HIPROOT` or `ROCM_PATH` should already be set to use the ROC libraries.

Fixes https://github.com/oneapi-src/oneMKL/issues/336

# Checklist

## All Submissions

- [N/A] Do all unit tests pass locally? This does not affect the tests.
- [N/A] Have you formatted the code using clang-format?

## Bug fixes

- [N/A] Have you added relevant regression tests?
- [x] Have you included information on how to reproduce the issue (either in a
      GitHub issue or in this PR)?

PR #527, opened 2024-07-04
[SPARSE] Add support for cuSPARSE backend

Rbiessy:
# Description

Add support for the cuSPARSE backend.

Fixes #25

Depends on https://github.com/oneapi-src/oneMKL/pull/500. All of the changes specific to cuSPARSE are in the last [commit](https://github.com/oneapi-src/oneMKL/pull/527/commits/7ade4de7aa7baf879854d4a69a1611efa6f659b0).

Rendered documentation: [docs.zip](https://github.com/user-attachments/files/16100716/docs.zip)

# Checklist

## All Submissions

- [x] Do all unit tests pass locally?
  - [cusparse_log.txt](https://github.com/user-attachments/files/16100608/cusparse_log.txt)
  - Tested on A100 with oneAPI 2024.2 compiler + [Codeplay Nvidia plugin](https://developer.codeplay.com/products/oneapi/nvidia/home/) + CUDA 12.5.0

- [x] Have you formatted the code using clang-format?

PR #544, opened 2024-07-24
[SPARSE] Add support for rocSPARSE backend

Rbiessy:
# Description

Add support for the rocSPARSE backend.

Depends on https://github.com/oneapi-src/oneMKL/pull/527 and https://github.com/oneapi-src/oneMKL/pull/532.

Rendered docs: [docs.zip](https://github.com/user-attachments/files/16363101/docs.zip)

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? [amd_log.txt](https://github.com/user-attachments/files/16363103/amd_log.txt) [nvidia_log.txt](https://github.com/user-attachments/files/16363105/nvidia_log.txt) [intel_log.txt](https://github.com/user-attachments/files/16363106/intel_log.txt)
  - Tested the rocSPARSE backend on MI210 with oneAPI 2024.2 compiler + [Codeplay AMD plugin](https://developer.codeplay.com/products/oneapi/amd/home/) + ROCm 6.1.0


Rbiessy:
Tests log on W6800: [amd_w6800_log.txt](https://github.com/user-attachments/files/16430518/amd_w6800_log.txt)

gajanan-choudhary:
@Rbiessy, now that #527 is merged in, could you please rebase this branch against `main`, resolve conflicts, and force-push when you can so that I can start reviewing this PR?

Rbiessy:
Yes, this is in progress! I expect I will need a few days at least. I'm aiming to merge the PR by the end of the year.

Rbiessy:
@gajanan-choudhary I have updated the PR with recent changes from cuSPARSE. Note that it moved almost all the content from `cusparse_task.hpp` to `common_launch_task.hpp` with almost no changes.
New test logs: 
[log_mi210.txt](https://github.com/user-attachments/files/17601859/log_mi210.txt)
[log_a100.txt](https://github.com/user-attachments/files/17601862/log_a100.txt)
[log_pvc.txt](https://github.com/user-attachments/files/17601863/log_pvc.txt)

README.md
```diff
@@ -61,7 +61,11 @@ oneMKL is part of the [UXL Foundation](http://www.uxlfoundation.org).
             <td align="center">AMD GPU</td>
         </tr>
         <tr>
-            <td align="center"><a href="https://github.com/ROCmSoftwarePlatform/rocFFT">AMD rocFFT</a></td>
+            <td align="center"><a href="https://github.com/ROCmSoftwarePlatform/rocFFT"> AMD rocFFT</a></td>
+            <td align="center">AMD GPU</td>
+        </tr>
+        <tr>
+            <td align="center"><a href="https://github.com/ROCmSoftwarePlatform/rocSPARSE"> AMD rocSPARSE</a></td>
```
gajanan-choudhary, line 68: Please also add rocSPARSE version info in the table later in this README file in the table in the [Product and Version Information](https://github.com/oneapi-src/oneMKL?tab=readme-ov-file#product-and-version-information) section

docs/domains/sparse_linear_algebra.rst
```diff
@@ -68,6 +68,31 @@ Currently known limitations:
...
+- The CSR format requires the column indices to be sorted within each row. See
+  the `rocSPARSE CSR documentation
+  <https://rocm.docs.amd.com/projects/rocSPARSE/en/latest/how-to/basics.html#csr-storage-format>`_.
+  Sparse operations using matrices with the CSR format without the property
+  ``matrix_property::sorted`` will throw a ``oneapi::mkl::unimplemented``
+  exception.
+- The same sparse matrix handle cannot be reused for multiple operations
+  ``spmm``, ``spmv``, or ``spsv``. Doing so will throw a
+  ``oneapi::mkl::unimplemented`` exception. See `#332
+  <https://github.com/ROCm/rocSPARSE/issues/332>`_.
```
gajanan-choudhary: Wow, this is quite severe, but seems to be a legitimate issue on `rocSPARSE` side right now.

gajanan-choudhary: If and when they fix this issue, though, will it be easy for us to make changes (with a version check of course) that correctly performs the operations rather than throwing an `unimplemented` exception?

src/sparse_blas/backends/cusparse/cusparse_handles.cpp
```diff
@@ -22,6 +22,7 @@
 #include "cusparse_error.hpp"
 #include "cusparse_helper.hpp"
 #include "cusparse_handles.hpp"
+#include "cusparse_scope_handle.hpp"
```
gajanan-choudhary, line 25: Why is a cuSPARSE change coming in here?

gajanan-choudhary: Is it related to the `common_launch_task.hpp` to keep common code between cuSPARSE and rocSPARSE backends?

src/sparse_blas/backends/rocsparse/rocsparse_task.hpp
```diff
@@ -0,0 +1,44 @@
...
+*
+**************************************************************************/
+
+#ifndef _ONEMKL_SPARSE_BLAS_BACKENDS_ROCSPARSE_TASK_HPP_
+#define _ONEMKL_SPARSE_BLAS_BACKENDS_ROCSPARSE_TASK_HPP_
+
+#include "rocsparse_error.hpp"
+#include "sparse_blas/backends/common_launch_task.hpp"
+
+namespace oneapi::mkl::sparse::rocsparse::detail {
```
gajanan-choudhary: Other than cusparse -> rocsparse, this file, rocsparse/rocsparse_task.hpp appears to be nearly identical to cusparse/cusparse_task.hpp. Is there anything more than that in this file?

docs/domains/sparse_linear_algebra.rst
Rbiessy: Yes, it's easy to fix on oneMath side. The issue is also referenced in this comment: https://github.com/oneapi-src/oneMKL/pull/544/files#diff-3b8c1c2c71abd54f8f90f43415c2f17b2a7fdb81c2b882c210f3cba56b4679adR63
One would just need to remove the `used` member, its 2 usages below as well as the `mark_used` method.

src/sparse_blas/backends/cusparse/cusparse_handles.cpp
Rbiessy: It's just as you said, `cusparse_task.hpp` used to include `cusparse_scope_handle.hpp` but it's not needed anymore (with the already merged changes to use the ScopeHandle object less often) and not possible anymore with the introduction of the more generic `common_launch_task.hpp`.

src/sparse_blas/backends/rocsparse/rocsparse_task.hpp
Rbiessy: Yes, they are almost identical. I'm not sure I understand, why do you ask?

README.md
Rbiessy: Thanks for catching that, done in https://github.com/oneapi-src/oneMKL/pull/544/commits/dc22051919dff1cf439fa9a0962ce1ab17e8efab

src/sparse_blas/backends/rocsparse/rocsparse_handles.hpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    void check_valid_handle(const std::string& function_name) {
+        if (used) {
+            throw mkl::unimplemented(
+                "sparse_blas", function_name,
+                "The backend does not support re-using the same sparse matrix handle in multiple operations.");
+        }
+        if (this->format == detail::sparse_format::COO &&
+            !this->has_matrix_property(matrix_property::sorted)) {
```
gajanan-choudhary, line 91: In the documentation, `docs/domains/sparse_linear_algebra.rst`, you've written:
> - The CSR format requires the column indices to be sorted within each row.

So shouldn't we be handling both CSR and COO formats here instead of just COO?

src/sparse_blas/backends/rocsparse/rocsparse_task.hpp
gajanan-choudhary: I was just confirming for my understanding to make sure I wasn't missing anything.

src/sparse_blas/backends/rocsparse/rocsparse_scope_handle.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    rocsparse_handle handle;
+    ROCSPARSE_ERR_FUNC(rocsparse_create_handle, &handle);
+    ROCSPARSE_ERR_FUNC(rocsparse_set_stream, handle, streamId);
+
+    auto atomic_ptr = new std::atomic<rocsparse_handle>(handle);
+    handle_helper.rocsparse_handle_container_mapper_.insert(
+        std::make_pair(piPlacedContext_, atomic_ptr));
+
+    sycl::detail::pi::contextSetExtendedDeleter(*placedContext_, ContextCallback, atomic_ptr);
```
gajanan-choudhary, line 129: Is it unchanged between the removed PI versus new UR APIs? Isn't there a `sycl::detail::ur::contextSetExtendedDeleter` somewhere that you need to place under `#ifdef ONEAPI_ONEMKL_PI_INTERFACE_REMOVED #else #endif...` ?

src/sparse_blas/backends/rocsparse/rocsparse_handles.cpp
```diff
@@ -0,0 +1,491 @@
...
+}
+
+template <typename fpType>
+void set_dense_vector_data(sycl::queue& queue, dense_vector_handle_t dvhandle, std::int64_t size,
+                           sycl::buffer<fpType, 1> val) {
+    detail::check_can_reset_value_handle<fpType>(__func__, dvhandle, true);
+    auto event = queue.submit([&](sycl::handler& cgh) {
+        auto acc = val.template get_access<sycl::access::mode::read_write>(cgh);
+        detail::submit_host_task(cgh, queue, [=](sycl::interop_handle ih) {
+            if (dvhandle->size != size) {
```
gajanan-choudhary, line 78: I wonder if we should throw an exception if `dvhandle->size < size` instead of `!=`. E.g., users may want to pad the`sycl::buffer` for some reason with zeros in the end, which wouldn't work with the current `if` condition.

gajanan-choudhary: Nevermind, this is `set_dense_vector_data` used for replacing/switching out arrays in an existing handle. The condition is fine then.

src/sparse_blas/backends/rocsparse/rocsparse_handles.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+/**
+ * In this file RocsparseScopedContextHandler are used to ensure that a rocsparse_handle is created before any other rocSPARSE call, as required by the specification.
+*/
+
+// Dense vector
+template <typename fpType>
+void init_dense_vector(sycl::queue& queue, dense_vector_handle_t* p_dvhandle, std::int64_t size,
+                       sycl::buffer<fpType, 1> val) {
+    auto event = queue.submit([&](sycl::handler& cgh) {
```
gajanan-choudhary, line 39: Can we add a check here for `sycl::buffer` case that checks `buffer->size() >= size` and throws an exception otherwise?

src/sparse_blas/backends/rocsparse/rocsparse_handles.cpp
gajanan-choudhary: Although maybe it doesn't hurt to add the check both here and in `init_xxx_data`

src/sparse_blas/backends/rocsparse/operations/rocsparse_spmv.cpp
```diff
@@ -0,0 +1,350 @@
...
+    }
+    CHECK_DESCR_MATCH(spmv_descr, opA, "spmv_optimize");
+    CHECK_DESCR_MATCH(spmv_descr, A_view, "spmv_optimize");
+    CHECK_DESCR_MATCH(spmv_descr, A_handle, "spmv_optimize");
+    CHECK_DESCR_MATCH(spmv_descr, x_handle, "spmv_optimize");
+    CHECK_DESCR_MATCH(spmv_descr, y_handle, "spmv_optimize");
+    CHECK_DESCR_MATCH(spmv_descr, alg, "spmv_optimize");
+
+    A_handle->mark_used();
+    auto& buffer_size = spmv_descr->temp_buffer_size;
```
gajanan-choudhary: Is there a reason why this variable specifically is kept and captured by reference _outside_ `compute_functor` while other members of `spmv_descr` are created inside the functor? If it is because `buffer_size` is used later in an `if` condition, I'd prefer it if we change this approach and capture everything by `[=]` in `compute_functor`, even though it would mean replacing the last `buffer_size > 0` with `spmv_descr->temp_buffer_size > 0`. I know that what you have right now is expected to work in this particular case (because of the condition that `spmv_descr` must live as long as `spmv` is being called), it is normally a bad idea to capture variables by reference in SYCL functors that are going to be running asynchronously unless you have an immediate `event.wait()` (e.g., like what we are doing in `spmv_buffer_size()` function).

gajanan-choudhary (approved):
LGTM other than some minor change requests. This PR was a lot easier to review having reviewed #527. Thanks for the fantastic work!

src/sparse_blas/backends/rocsparse/rocsparse_scope_handle.cpp
Rbiessy: The `contextSetExtendedDeleter` is still in the `pi` namespace see https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/ur.hpp#L100. As the comment says there are plans to deprecate it. The oneAPI core team at Codeplay started to change the scopeContextHandle mechanism for BLAS in https://github.com/oneapi-src/oneMKL/pull/609. They will make sure that oneMKL Interface will keep working if/when `contextSetExtendedDeleter` is deprecated.
If we ever need to change the scopeContextHandle mechanism, I would rather that we do it for both cuSPARSE and rocSPARSE backends in a separate PR.

src/sparse_blas/backends/rocsparse/rocsparse_handles.cpp
Rbiessy: The check to verify that the buffer size is compatible with the handle size is done elsewhere, see https://github.com/oneapi-src/oneMKL/blob/develop/src/sparse_blas/generic_container.hpp#L164
This is a common place that the other backends also use. That check will run for `init_*_data` functions.

src/sparse_blas/backends/rocsparse/rocsparse_handles.cpp
Rbiessy: This is already done, see my comment above: https://github.com/oneapi-src/oneMKL/pull/544#discussion_r1863590450

src/sparse_blas/backends/rocsparse/rocsparse_handles.hpp
Rbiessy: Good point, I added this check in https://github.com/oneapi-src/oneMKL/pull/544/commits/b6518c90d986b7d65f37aed368fcaeac18adad79
I've also made sure we also have tests on AMD and Nvidia which are skipped when a property is not set. The runtime example also need to set this property now.

src/sparse_blas/backends/rocsparse/operations/rocsparse_spmv.cpp
Rbiessy: Sure that's fine, done in https://github.com/oneapi-src/oneMKL/pull/544/commits/c53ef8d2cc49f9ec695b1d01bd49d238513ac42d

Rbiessy:
Thanks for the review Gajanan!

For completeness, logs on MI210: [log_mi210.txt](https://github.com/user-attachments/files/17981584/log_mi210.txt)
and A100: [log_a100.txt](https://github.com/user-attachments/files/17981589/log_a100.txt)

FYI we will get a second approval and I aim to merge this after the oneMKL Interface renaming PR.

Rbiessy:
Our AMD HW is unavailable for a while unfortunately. I will work on fixing the conflicts as soon as I can test the changes.

Rbiessy:
I fixed conflicts with the oneMath renaming. New test log on MI210: [onemath_mi210.txt](https://github.com/user-attachments/files/18026827/onemath_mi210.txt)

s-Nick (approved):

PR #550, opened 2024-08-07
Set RPATH for the loader

al42and:
# Description

Having the user manage LD_LIBRARY_PATH is sub-optimal. It is mentioned three times in the docs, which suggest it is a common issue. By setting the `RUNPATH` on the loader library we can make `dlopen` automagically find the right libraries.

As a downside, we won't be able to point the loader to another oneMKL installation. I'd consider it more of an upside since it prevents accidentally mixing up two versions, but perhaps someone was relying on it?

Only tested on Linux, but official oneAPI plugins only support Linux, so :shrug:

# Checklist

## All Submissions

  - [ctest.log](https://github.com/user-attachments/files/16531860/ctest.log)
  - Not touching the code

## New interfaces

it was accepted? # (RFC)

## New features

## Bug fixes

      GitHub issue or in this PR)?


Rbiessy (approved):
Thanks for the PR, I think this is much better!

mkrainiuk (approved):
Thank you so much for fixing this problem!

src/CMakeLists.txt
```diff
@@ -81,6 +81,8 @@ if(BUILD_SHARED_LIBS)
   set (ONEMKL_LIBS ${TARGET_DOMAINS})
   list(TRANSFORM ONEMKL_LIBS PREPEND onemkl_)
   target_link_libraries(onemkl PUBLIC ${ONEMKL_LIBS})
+  set_target_properties(onemkl PROPERTIES
+      INSTALL_RPATH "\$ORIGIN")
```
Rbiessy:
```suggestion
  set_target_properties(onemkl PROPERTIES
      INSTALL_RPATH "\$ORIGIN"
      BUILD_RPATH "\$ORIGIN"
  )
```

I was just testing this. `BUILD_RPATH` is needed when running the examples without running the install step.

Rbiessy (changes requested):
See https://github.com/oneapi-src/oneMKL/pull/550#discussion_r1749977987

src/CMakeLists.txt
al42and: Nice catch! But I think `BUILD_WITH_INSTALL_RPATH TRUE` would be a bit better, since CMake (at least 3.30) is not smart enough to see that `INSTALL_RPATH` and `BUILD_RPATH` are the same, and still tries to modify the SO when installing. Quite fast and harmless, but still.

Rbiessy (approved):

Issue #542, opened 2024-07-24
When trying to use oneMKL with the portBLAS backend there is a check for Intel, AMD or Nvidia GPU.


al3x-jp:
# Summary
When trying to use oneMKL with the portBLAS backend, the current code structure checks for an Intel, AMD or NVidia GPU, which if not found causes an unsupported error. It is understood that portBLAS could work on any GPU that supports OpenCL.
 
# Version
HEAD of oneMKL Tree
 
# Environment
Ubuntu 22.04 with PowerVR GPU
 
# Steps to reproduce
Build oneMKL with "portBLAS" support specified and then execute the example GEMM test on a GPU that isn't Intel, AMD or Nvidia, but that supports SYCL OpenCL backend (e.g., a PowerVR GPU).
 
To build:
cmake .. -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DENABLE_MKLGPU_BACKEND=False -DENABLE_MKLGPU_BACKEND=False -DENABLE_PORTBLAS_BACKEND=True -DENABLE_CUBLAS_BACKEND=False -DENABLE_ROCBLAS_BACKEND=False -DENABLE_NETLIB_BACKEND=False -DBUILD_FUNCTIONAL_TESTS=False -DBUILD_EXAMPLES=True -DENABLE_MKLCPU_BACKEND=False
 
To run:
ONEAPI_DEVICE_SELECTOR=opencl:gpu SYCL_PI_TRACE=1 ./example_blas_gemm_usm
 
# Observed behaviour
The BLAS GEMM example prints the following message when run on a PowerVR OpenCL Device: 
"oneMKL: PowerVR B-Series BXT-32-1024 is not supported"
 
# Expected behaviour
It is expected that portBLAS could work on any GPU that supports the SYCL OpenCL backend.

Rbiessy:
Thank you for this issue @al3x-jp. You are right currently the oneMKL structure does not allow a backend to support devices that are not in a list of [supported device](https://github.com/oneapi-src/oneMKL/blob/develop/include/oneapi/mkl/detail/get_device_id.hpp#L43).
I think we should have some kind of generic device that would allow portBLAS and portFFT backends to run on any SYCL device. These devices won't necessarily be "supported" until we have a CI testing them but they could be documented as expected to work.
We'll let you know once we are able to work on this issue.

hjabird:
I've started looking into this issue.

hjabird:
[...]

PR #552, opened 2024-08-08
Add generic device; Initial support in portBLAS

hjabird:
# Description

* oneMKL Interfaces currently only supports known targets: Intel CPU/GPU, AMD GPU, Nvidia GPU
* This PR:
  * Enables a new generic target
  * Enables the generic target to use the portBLAS backend
  * Adds documentation
Relates to https://github.com/oneapi-src/oneMKL/issues/542

# Checklist

## All Submissions

- [NO] Do all unit tests pass locally? Attach a log.
  - I do not have a non Intel/Nvidia/AMD device to test with.
  - Tests appeared to run successfully when the code to recognise Intel devices was modified such that they fell through to the new "generic_device".

## New features

- [N/A] Have you added relevant tests?
  - The generic device is clearly documented as an unsupported configuration.


hjabird:
@al3x-jp I hope that this PR enables oneMKL interfaces to run on the PowerVR GPU with portBLAS as detailed in https://github.com/oneapi-src/oneMKL/issues/542. Could you try this out?

al3x-jp:
Hi, I have tried this (thank you) but it now fails with this error:

>> libze_intel_vpu.so.1: cannot open shared object file: No such file or directory
>> [New Thread 0x7ffff309d640 (LWP 5829)]
>> Caught std::exception during GEMM:
>>        oneMKL: Couldn't load selected backend

oneMKL (debug build) was built with the following command line: 

cmake .. -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DENABLE_MKLGPU_BACKEND=False -DENABLE_MKLGPU_BACKEND=False -DENABLE_PORTBLAS_BACKEND=True -DENABLE_CUBLAS_BACKEND=False -DENABLE_ROCBLAS_BACKEND=False -DENABLE_NETLIB_BACKEND=False -DBUILD_FUNCTIONAL_TESTS=False -DBUILD_EXAMPLES=True -DENABLE_MKLCPU_BACKEND=False -DENABLE_GENERIC_DEVICE=True

hjabird:
`libze_intel_vpu.so.1` is not part of oneMKL Interfaces. It sounds like the DPC++ runtime is failing to find/open an Intel level-zero VPU backend. I don't know how the DPC++ runtime normally works with PowerVR. Does it just work when you build portBLAS independently? Does `sycl-ls` work correctly?

hjabird:
Hi @al3x-jp . Did you manage to investigate your problem any further?

CMakeLists.txt
```diff
@@ -60,6 +60,9 @@ option(ENABLE_CUFFT_BACKEND "Enable the cuFFT backend for the DFT interface" OFF
 option(ENABLE_ROCFFT_BACKEND "Enable the rocFFT backend for the DFT interface" OFF)
 option(ENABLE_PORTFFT_BACKEND "Enable the portFFT DFT backend for the DFT interface. Cannot be used with other DFT backends." OFF)
 
+# Generic devices
+option(ENABLE_GENERIC_DEVICE "Enable generic devices. Requires the portBLAS backend." OFF)
```
Rbiessy, line 64: I'm not keen on adding yet another user option. Could this be automatically enabled whenever portBLAS is enabled and the default tuning target is used?

include/oneapi/mkl/detail/get_device_id.hpp
```diff
@@ -59,11 +59,19 @@ inline oneapi::mkl::device get_device_id(sycl::queue &queue) {
         else if (vendor_id == AMD_ID)
             device_id = device::amdgpu;
         else {
+#ifdef ENABLE_GENERIC_DEVICE
```
Rbiessy, line 62: It seems to me we could avoid these ifdef and always return a `generic_device` as a fallback. The `table_initializer` object should throw `backend_not_found` the `generic_device` is used with a backend that does not support it.

CMakeLists.txt
hjabird: We could.

Generic device really means an unsupported device. That means:
* It may not be tested, so there is a change of getting the wrong result.
* Errors are likely to come from somewhere other than oneMKL Interfaces.

Consequently, I think any of this functionality should explicitly be opt-in. Additionally, we currently attempt to auto-detect a tuning target if it isn't set. This behavour is probably useful, although admittedtly portBLAS is even more useful for targets that don't have a vendor library already part of oneMKL Interfaces.

Rbiessy: > admittedtly portBLAS is even more useful for targets that don't have a vendor library already part of oneMKL Interfaces.

Yes that was my main reason to avoid having this option.

Thinking some more about it, if portBLAS always uses the "generic device" even for supported devices then it wouldn't conflict with the other backends anymore. It would make it easier to have portBLAS used as a fallback. The fallback feature won't happen in this PR but I think we will likely want to remove this option in the future so might as well not add it. 
I think we could have portBLAS always use the generic device as part of this PR as well.

Rbiessy:
Closing in favor of https://github.com/oneapi-src/oneMKL/pull/566

PR #556, opened 2024-08-12
[Doc][Web] Fix build command with DPC++ in documentation webpage

gajanan-choudhary:
## Changes
* One-liner fix in the ["Building the project with DPC++" webpage](https://oneapi-src.github.io/oneMKL/building_the_project_with_dpcpp.html) to use `MKLCPU` backend instead of the incorrectly twice-used `MKLGPU` backend.


dnhsieh-intel (approved):
Thanks for the fix!

hjabird (approved):
Thanks for fixing this!

PR #557, opened 2024-08-16
[Build][Docs][CI] BUILD_FUNCTIONAL_TESTS=False by default

### LINKED ISSUE #443: [CMake] New users can struggle with the CBLAS testing dependency when building oneMKL - what to do?
**The problem**

Both in the issues ([here](https://github.com/oneapi-src/oneMKL/issues/436)), and privately, I've supported new users struggling to build the oneMKL library. Their aim was to build and install oneMKL. 

In both cases, the sticking point was the dependency of the BLAS unit tests on CBLAS. This dependency on CBLAS, combined with the fact that the unit tests are build by default, meant that they required support.

**Possible solutions**

1. Only build the oneMKL library by default: building anything else (examples, tests), should be opt-in. This would also reduce build/install times.
2. More helpful error messages for the CBLAS dependency. Error messages should make it clear how to avoid this dependency.

hjabird:
# Description

* Currently tests build by default
  * This adds additional dependencies, tripping users up.
  * This slows down builds where the tests are not used.
* This PR disabled tests by default and
  * Updates the documentation
  * Updates the Github CI

Fixes https://github.com/oneapi-src/oneMKL/issues/443

# Checklist

## All Submissions

- [N/A] Do all unit tests pass locally? Attach a log.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -75,7 +75,7 @@ jobs:
       if: steps.domain_check.outputs.result == 'true'
       run: |
         source /opt/intel/oneapi/setvars.sh
-        cmake -DTARGET_DOMAINS=${{ matrix.domain }} -DENABLE_MKLGPU_BACKEND=OFF -DBUILD_FUNCTIONAL_TESTS=ON -DCMAKE_VERBOSE_MAKEFILE=on ${{ matrix.build_options }} -B build
+        cmake -DTARGET_DOMAINS=${{ matrix.domain }} -DENABLE_MKLGPU_BACKEND=OFF -DBUILD_FUNCTIONAL_TESTS=ON -DCMAKE_VERBOSE_MAKEFILE=ON ${{ matrix.build_options }} -B build
         cmake --build build ${PARALLEL}
     - name: Run tests
       if: steps.domain_check.outputs.result == 'true'
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -75,7 +75,7 @@ jobs:
       if: steps.domain_check.outputs.result == 'true'
       run: |
         source /opt/intel/oneapi/setvars.sh
-        cmake -DTARGET_DOMAINS=${{ matrix.domain }} -DENABLE_MKLGPU_BACKEND=off -DCMAKE_VERBOSE_MAKEFILE=on ${{ matrix.build_options }} -B build
+        cmake -DTARGET_DOMAINS=${{ matrix.domain }} -DENABLE_MKLGPU_BACKEND=OFF -DBUILD_FUNCTIONAL_TESTS=ON -DCMAKE_VERBOSE_MAKEFILE=on ${{ matrix.build_options }} -B build
         cmake --build build ${PARALLEL}
     - name: Run tests
       if: steps.domain_check.outputs.result == 'true'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,7 @@ set(ONEMKL_SYCL_IMPLEMENTATION "dpc++" CACHE STRING "Name of the SYCL compiler")
 set(HIP_TARGETS "" CACHE STRING "Target HIP architectures")
 
 ## Testing
-option(BUILD_FUNCTIONAL_TESTS "" ON)
+option(BUILD_FUNCTIONAL_TESTS "" OFF)
 
 ## Examples
 option(BUILD_EXAMPLES "" ON)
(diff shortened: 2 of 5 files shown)


.github/workflows/pr.yml
```diff
@@ -75,7 +75,7 @@ jobs:
       if: steps.domain_check.outputs.result == 'true'
       run: |
         source /opt/intel/oneapi/setvars.sh
-        cmake -DTARGET_DOMAINS=${{ matrix.domain }} -DENABLE_MKLGPU_BACKEND=off -DCMAKE_VERBOSE_MAKEFILE=on ${{ matrix.build_options }} -B build
+        cmake -DTARGET_DOMAINS=${{ matrix.domain }} -DENABLE_MKLGPU_BACKEND=OFF -DBUILD_FUNCTIONAL_TESTS=ON -DCMAKE_VERBOSE_MAKEFILE=on ${{ matrix.build_options }} -B build
```
Rbiessy:
```suggestion
        cmake -DTARGET_DOMAINS=${{ matrix.domain }} -DENABLE_MKLGPU_BACKEND=OFF -DBUILD_FUNCTIONAL_TESTS=ON -DCMAKE_VERBOSE_MAKEFILE=ON ${{ matrix.build_options }} -B build
```
You probably want to also change this for consistency?

Rbiessy (approved):

CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ set(ONEMKL_SYCL_IMPLEMENTATION "dpc++" CACHE STRING "Name of the SYCL compiler")
 set(HIP_TARGETS "" CACHE STRING "Target HIP architectures")
 
 ## Testing
-option(BUILD_FUNCTIONAL_TESTS "" ON)
+option(BUILD_FUNCTIONAL_TESTS "" OFF)
```
mkrainiuk, line 67: I agree that this extra Netlib dependency causes problems for unexperienced users, so it makes sense to disable it, but I think we still need some default validation to make sure the build is fine. As far as I know extra dependency is a problem only for BLAS and LAPACK, does it make sense to keep testing for other domains since they don't have extra dependencies? And for BLAS/LAPACK we can come up with more light-weight testing in the future.

Rbiessy: Are you talking about a default validation for unexperienced users? The examples are still built by default and don't require any external dependencies. This should be enough to make sure everything built fine.

mkrainiuk: I'm not sure I can agree that passed examples are enough to confirm that built oneMKL Interfaces functionality is fine, we cover just 1-2 functions per domain, only USM API and mostly Intel oneMKL CPU backend for compiler time dispatcher from the build and we do not check correctness which could be a tricky problem to catch in the application where this oneMKL Interfaces build will be used.

Rbiessy: I'm not saying the examples are enough to test all the functionalities, just that they are enough to show oneMKL built. The point of this PR is that we have seen users getting confused why they couldn't build oneMKL and the issue was due to the additional dependencies needed by the tests. This PR aims to clarify this and force the users to manually enable the tests if they want them.
I think this makes it easier for first time users.

Rbiessy: Gentle ping @mkrainiuk. What are your thoughts on this?

mkrainiuk: I support the idea of disabling testing for BLAS and LAPACK for the first time users and come up with some light testing that won't require Netlib package in the future, but I don't understand why we need to disable testing for DFT, RNG, and spBLAS that have proper testing coverage by default and don't have any extra dependencies that might confuse users. Could you please clarify why it's required to change testing mode for these domains as part of Netlib dependency problem?

Rbiessy: It is much simpler for users and developers if a CMake option always has the same default value. Some people use CMake GUI tools which show all the options with the default value but this causes confusions when the value change due to another option. We would need to add some documentation to explain when the tests are enabled or disabled and hope users find it.

mkrainiuk: That's true, in this case I don't think modifying common option will be the right solution, instead it will be better to change requirements for Netlib dependency in two specific domains, where it's used, we can make it optional and based on the Netlib availability enable/disable unit tests where Netlib is used (later on adding some tests where it's not used so they can be built without Netlib).

mkrainiuk: To be more specific here is a draft with some quick changes for what I've proposed before as potential solution for Netlib dependency problem https://github.com/oneapi-src/oneMKL/pull/577 
Please let me know what do you think about this approach.

Rbiessy: That's a good idea. Your solution looks fine to me!

Rbiessy:
Closing in favor of https://github.com/oneapi-src/oneMKL/pull/577

Issue #559, opened 2024-08-19
[ROCFFT] RocFFT fails tests when using ROCm 6.0 or later


hjabird:
# Summary

The current tip of oneMKL interfaces fails unit tests when using ROCm 6.0 or later. Using ROCm 5.4.3 does not fail tests.

The failed tests are real-to-complex multi-dimensional tests.

# Version

The issue was introduced with https://github.com/oneapi-src/oneMKL/pull/528 and can be reproduced with oneMKL interfaces developr.

# Environment

* AMD MI210
* Reproduceable RocFFT from ROCm 6.0 or 6.1
* Ubuntu 22.04
* ICPX 2024.2 with Codeplay's plugin for AMD (which supports ROCm 6.0.2 for MI210).

# Steps to reproduce

Build with RocFFT enabled. Test with

```
./bin/test_main_dft_ct --gtest_filter=*REAL_SINGLE_in_place_USM*batches_1*
```
to observe failures. 

# Observed behavior
Tests give wrong results or memory faults.

# Expected behavior
Tests should pass.

hjabird:
* It seems to be the backward DFT that fails. 
* If the tests prior to regression-causing PR are used, the tests still pass. (ie. tests using the INPUT/OUTPUT API with recommit, instead of using FWD/BWD API).
* This problem can also be reproduced on an AMD W6800.

hjabird:
This turned out to be a rocFFT bug. RocFFT issue https://github.com/ROCm/rocFFT/issues/504

hjabird:
There is a work-around for the above at https://github.com/hjabird/oneMKL/tree/hjab/fix_rocfft6_issue

Unfortunately there is still are still failing tests - out-of-place complex 4x4x4_fwd_strides_2_4_1_16_bwd_strides_1_4_16_1_batches_2. These tests pass with ROCm 5.4.3, but fail with ROCm 5.7.1 and later.

hjabird:
This looks like a bug in rocFFT to me. I've described the issue at https://github.com/ROCm/rocFFT/issues/507. I think we oneMKL Interfaces will have to throw unsupported on these tranposing DFTs.

Rbiessy:
The 2 workarounds have been merged and the issues fixed in more recent versions of rocFFT. Closing.

PR #563, opened 2024-08-30
[DFT][rocFFt] Address rocFFT failing tests

s-Nick:
# Description

This PR addresses two issues arose recently in rocFFT backend. Both issues are related to rocFFT internal changes.
One issue is fixed by small modification to oneMKL backend. The other issue cannot be fixed internally, but it was fixed by AMD and it will available in the future versions of rocFFT. 
To avoid that users incur with the second issues a warning is emitted during the configuration and eventually an exception is thrown. 

first issue -> https://github.com/ROCm/rocFFT/issues/504 
second issue -> https://github.com/ROCm/rocFFT/issues/507

Fixes # (GitHub issue)

This PR address oneMKL issue #559 

# Checklist

## All Submissions

older rocm version(5.4.3) -> [rocm5_4_3_full_log.txt](https://github.com/user-attachments/files/16817642/rocm5_4_3_full_log.txt)
current release -> [rocm6_1_fix_log.txt](https://github.com/user-attachments/files/16817651/rocm6_1_fix_log.txt)
rocFFT develop branch -> [develop_fix_log.txt](https://github.com/user-attachments/files/16817657/develop_fix_log.txt)

## Bug fixes

      GitHub issue or in this PR)?
 issue #559

### DIFF
--- a/src/dft/backends/rocfft/CMakeLists.txt
+++ b/src/dft/backends/rocfft/CMakeLists.txt
@@ -49,6 +49,14 @@ find_package(HIP REQUIRED)
 # Require the minimum rocFFT version matching with ROCm 5.4.3.
 find_package(rocfft 1.0.21 REQUIRED)
 
+if (${rocfft_VERSION_MAJOR} EQUAL "1" AND ${rocfft_VERSION_MINOR} EQUAL "0"
+    AND ((${rocfft_VERSION_PATCH} GREATER "22")
+    AND (${rocfft_VERSION_PATCH} LESS "31")  ))
+    message(WARNING "Due to a bug in rocFFT some tests fail with the version in\
+      use. If possible use a version greater of 1.0.30 or less of 1.0.23.
+      Current rocFFT version ${rocfft_VERSION}")
+endif()
+
 target_link_libraries(${LIB_OBJ} PRIVATE hip::host roc::rocfft)
 
 # Allow to compile for different ROCm versions. See the README for the supported
@@ -62,6 +70,7 @@ find_path(
   NO_DEFAULT_PATH
   REQUIRED
 )
+
 target_include_directories(${LIB_OBJ} PRIVATE ${rocfft_EXTRA_INCLUDE_DIR})
 
 target_link_libraries(${LIB_OBJ} PUBLIC ONEMKL::SYCL::SYCL)
--- a/src/dft/backends/rocfft/CMakeLists.txt
+++ b/src/dft/backends/rocfft/CMakeLists.txt
@@ -62,7 +62,6 @@ find_path(
   NO_DEFAULT_PATH
   REQUIRED
 )
-
 target_include_directories(${LIB_OBJ} PRIVATE ${rocfft_EXTRA_INCLUDE_DIR})
 
 target_link_libraries(${LIB_OBJ} PUBLIC ONEMKL::SYCL::SYCL)
--- a/src/dft/backends/rocfft/CMakeLists.txt
+++ b/src/dft/backends/rocfft/CMakeLists.txt
@@ -49,14 +49,6 @@ find_package(HIP REQUIRED)
 # Require the minimum rocFFT version matching with ROCm 5.4.3.
 find_package(rocfft 1.0.21 REQUIRED)
 
-if (${rocfft_VERSION_MAJOR} EQUAL "1" AND ${rocfft_VERSION_MINOR} EQUAL "0"
-    AND ((${rocfft_VERSION_PATCH} GREATER "22")
-    AND (${rocfft_VERSION_PATCH} LESS "31")  ))
-    message(WARNING "Due to a bug in rocFFT some tests fail with the version in\
-      use. If possible use a version greater of 1.0.30 or less of 1.0.23.
-      Current rocFFT version ${rocfft_VERSION}")
-endif()
-
 target_link_libraries(${LIB_OBJ} PRIVATE hip::host roc::rocfft)
 
 # Allow to compile for different ROCm versions. See the README for the supported
--- a/src/dft/backends/rocfft/commit.cpp
+++ b/src/dft/backends/rocfft/commit.cpp
@@ -266,6 +266,8 @@ public:
         // Link to rocFFT issue: https://github.com/ROCm/rocFFT/issues/507
         if constexpr (rocfft_version_major == 1 && rocfft_version_minor == 0 &&
                       (rocfft_version_patch > 22 && rocfft_version_patch < 31)) {
+            // rocFFT's functional status for problems like cfoA:B:1xB:1:A is unknown as
+            // of 4ed3e97bb7c11531684168665d5a980fde0284c9 (due to project's implementation preventing testing thereof)
             if (dom == dft::domain::COMPLEX &&
                 config_values.placement == dft::config_value::NOT_INPLACE && dimensions > 2) {
                 if (stride_vecs.vec_a != stride_vecs.vec_b)
--- a/src/dft/backends/rocfft/commit.cpp
+++ b/src/dft/backends/rocfft/commit.cpp
@@ -39,6 +39,7 @@
 #include "rocfft_handle.hpp"
 
 #include <rocfft.h>
+#include <rocfft-version.h>
 #include <hip/hip_runtime_api.h>
 
 namespace oneapi::mkl::dft::rocfft {
@@ -258,6 +259,28 @@ public:
         stride_vecs.vec_a.pop_back(); // Offset is not included.
         std::reverse(stride_vecs.vec_b.begin(), stride_vecs.vec_b.end());
         stride_vecs.vec_b.pop_back(); // Offset is not included.
+
+        // This workaround is needed due to a confirmed issue in rocFFT from version
+        // 1.0.23 to 1.0.30. Those rocFFT version correspond to rocm version from
+        // 5.6.0 to 6.3.0.
+        // Link to rocFFT issue: https://github.com/ROCm/rocFFT/issues/507
+        if constexpr (rocfft_version_major == 1 && rocfft_version_minor == 0 &&
+                      (rocfft_version_patch > 22 && rocfft_version_patch < 31)) {
+            if (dom == dft::domain::COMPLEX && dimensions > 2) {
+                auto stride_checker = [&](const auto& a, const auto& b) {
+                    for (ulong i = 0; i < dimensions; ++i) {
+                        if (a[i] != b[i])
+                            return false;
+                    }
+                    return true;
+                };
+                std::printf("hello\n");
+                if (!stride_checker(stride_vecs.vec_a, stride_vecs.vec_b))
+                    throw oneapi::mkl::unimplemented(
+                        "DFT", func,
+                        "due to a bug in rocfft version in use, it requires fwd and bwd stride to be the same for COMPLEX out_of_place computations");
+            }
+        }
 
         rocfft_plan_description plan_desc_fwd, plan_desc_bwd; // Can't reuse with ROCm 6 due to bug.
         if (rocfft_plan_description_create(&plan_desc_fwd) != rocfft_status_success) {
--- a/src/dft/backends/rocfft/commit.cpp
+++ b/src/dft/backends/rocfft/commit.cpp
@@ -274,7 +274,6 @@ public:
                     }
                     return true;
                 };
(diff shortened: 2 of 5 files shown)


src/dft/backends/rocfft/CMakeLists.txt
```diff
@@ -49,6 +49,14 @@ find_package(HIP REQUIRED)
 # Require the minimum rocFFT version matching with ROCm 5.4.3.
 find_package(rocfft 1.0.21 REQUIRED)
 
+if (${rocfft_VERSION_MAJOR} EQUAL "1" AND ${rocfft_VERSION_MINOR} EQUAL "0"
+    AND ((${rocfft_VERSION_PATCH} GREATER "22")
+    AND (${rocfft_VERSION_PATCH} LESS "31")  ))
+    message(WARNING "Due to a bug in rocFFT some tests fail with the version in\
+      use. If possible use a version greater of 1.0.30 or less of 1.0.23.
+      Current rocFFT version ${rocfft_VERSION}")
+endif()
```
Rbiessy: I don't think this is needed since users should see the unimplemented exception when using this rocFFT version, not the tests failing.

src/dft/backends/rocfft/commit.cpp
```diff
@@ -259,12 +260,37 @@ class rocfft_commit final : public dft::detail::commit_impl<prec, dom> {
         std::reverse(stride_vecs.vec_b.begin(), stride_vecs.vec_b.end());
         stride_vecs.vec_b.pop_back(); // Offset is not included.
 
-        rocfft_plan_description plan_desc;
-        if (rocfft_plan_description_create(&plan_desc) != rocfft_status_success) {
+        // This workaround is needed due to a confirmed issue in rocFFT from version
+        // 1.0.23 to 1.0.30. Those rocFFT version correspond to rocm version from
+        // 5.6.0 to 6.3.0.
```
Rbiessy, line 265: Where did you get the information about ROCm 6.3.0 and rocFFT 1.0.30? The latest versions I can see are [ROCm 6.2.0](https://github.com/ROCm/ROCm/releases) which uses [rocFFT 1.0.28](https://rocm.docs.amd.com/en/docs-6.2.0/compatibility/compatibility-matrix.html). I think we can assume that the fix will be part of rocFFT 1.0.29 and ROCm 6.3.0.

src/dft/backends/rocfft/commit.cpp
```diff
@@ -259,12 +260,37 @@ class rocfft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+            if (dom == dft::domain::COMPLEX && dimensions > 2) {
+                auto stride_checker = [&](const auto& a, const auto& b) {
+                    for (ulong i = 0; i < dimensions; ++i) {
+                        if (a[i] != b[i])
+                            return false;
+                    }
+                    return true;
+                };
+                std::printf("hello\n");
+                if (!stride_checker(stride_vecs.vec_a, stride_vecs.vec_b))
```
Rbiessy: Can't you use `stride_vecs.vec_a != stride_vecs.vec_b`? Or is `dimensions` smaller than the length of these vectors?

src/dft/backends/rocfft/commit.cpp
```diff
@@ -39,6 +39,7 @@
 #include "rocfft_handle.hpp"
 
 #include <rocfft.h>
+#include <rocfft-version.h>
```
s-Nick, line 42: This header is now needed to check rocFFT version if necessary

src/dft/backends/rocfft/commit.cpp
```diff
@@ -259,12 +260,37 @@ class rocfft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+                        if (a[i] != b[i])
+                            return false;
+                    }
+                    return true;
+                };
+                std::printf("hello\n");
+                if (!stride_checker(stride_vecs.vec_a, stride_vecs.vec_b))
+                    throw oneapi::mkl::unimplemented(
+                        "DFT", func,
+                        "due to a bug in rocfft version in use, it requires fwd and bwd stride to be the same for COMPLEX out_of_place computations");
```
Rbiessy, line 276: I think this whole check should be run in the out-of-place compute functions otherwise it looks like this will prevent users from running complex in-place FFTs for no reasons.

src/dft/backends/rocfft/commit.cpp
```diff
@@ -259,12 +260,36 @@ class rocfft_commit final : public dft::detail::commit_impl<prec, dom> {
         std::reverse(stride_vecs.vec_b.begin(), stride_vecs.vec_b.end());
         stride_vecs.vec_b.pop_back(); // Offset is not included.
 
-        rocfft_plan_description plan_desc;
-        if (rocfft_plan_description_create(&plan_desc) != rocfft_status_success) {
+        // This workaround is needed due to a confirmed issue in rocFFT from version
+        // 1.0.23 to 1.0.30. Those rocFFT version correspond to rocm version from
+        // 5.6.0 to 6.3.0.
+        // Link to rocFFT issue: https://github.com/ROCm/rocFFT/issues/507
+        if constexpr (rocfft_version_major == 1 && rocfft_version_minor == 0 &&
```
s-Nick, line 267: This if-statement could be converted to `#if` directive. I preferred to use this style instead, but I am open to change it

src/dft/backends/rocfft/CMakeLists.txt
s-Nick: Thank you for your review @Rbiessy . I added it to warn the user before compiling the library even without the test, but if you think it is redundant I can remove it.

src/dft/backends/rocfft/commit.cpp
s-Nick: I took from their change log in the rocFFT repo https://github.com/ROCm/rocFFT/blob/develop/CHANGELOG.md
According to it, there should be a ROCM 6.2.1 that uses rocFFT 1.0.29 then ROCm 6.3.0 with rocFFT 1.0.30 and then the fix for one of the issues

src/dft/backends/rocfft/commit.cpp
s-Nick: Honestly, I did not check. Dimensions should be the same. I'll try and update it.

src/dft/backends/rocfft/commit.cpp
s-Nick: yes, that would make more sense, but we can let that function clean and I can add another check in the `if` at [line 269](https://github.com/oneapi-src/oneMKL/blob/54e0cf91910a6584e8c81c522f0a2417cb7a6e10/src/dft/backends/rocfft/commit.cpp#L269). I added a comment directly to that line.

If you think change the compute function it is a better choice I am ok with it too.

src/dft/backends/rocfft/commit.cpp
```diff
@@ -259,12 +260,37 @@ class rocfft_commit final : public dft::detail::commit_impl<prec, dom> {
...
+                      (rocfft_version_patch > 22 && rocfft_version_patch < 31)) {
+            if (dom == dft::domain::COMPLEX && dimensions > 2) {
```
s-Nick: This change should allow `in_place` routine to run
```suggestion
            if (dom == dft::domain::COMPLEX &&  config_values.placement == dft::config_value::NOT_INPLACE && dimensions > 2) {
```

src/dft/backends/rocfft/CMakeLists.txt
Rbiessy: Yeah it can be removed.

src/dft/backends/rocfft/commit.cpp
Rbiessy: Thanks that looks good then.

src/dft/backends/rocfft/commit.cpp
Rbiessy: You're right, I forgot that the placement was already known at commit time so that makes. Btw you should make sure the changes are formatted with clang-format-9 to avoid issues in the CI!

src/dft/backends/rocfft/CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ find_path(
   NO_DEFAULT_PATH
   REQUIRED
 )
+
```
Rbiessy: Might as well remove this new line to reduce noise

src/dft/backends/rocfft/descriptor.cpp
```diff
@@ -22,9 +22,7 @@
 
 #include "oneapi/mkl/dft/detail/rocfft/onemkl_dft_rocfft.hpp"
 
-namespace oneapi {
-namespace mkl {
-namespace dft {
+namespace oneapi::mkl::dft::detail {
```
lhuot, line 25: It seems the namespace in this file is not consistent in all backends.
With this change rocFFT, oneMKL CPU, oneMKL GPU have the oneapi::mkl::dft::detail namespace but cuFFT and portFFT have oneapi::mkl::dft namespace.
I know this is unrelated to the original issue, but since you changed one of them, could you please make them consistent across all backends?

s-Nick: Thank you for your review @lhuot 
No problem, I updated them in [aed1f63](https://github.com/oneapi-src/oneMKL/pull/563/commits/aed1f63fea977b65a5c2da785b5405e4d1ae8850)

Rbiessy (approved):

src/dft/backends/rocfft/commit.cpp
```diff
@@ -259,12 +260,30 @@ class rocfft_commit final : public dft::detail::commit_impl<prec, dom> {
...
-        rocfft_plan_description plan_desc;
-        if (rocfft_plan_description_create(&plan_desc) != rocfft_status_success) {
+        // This workaround is needed due to a confirmed issue in rocFFT from version
+        // 1.0.23 to 1.0.30. Those rocFFT version correspond to rocm version from
+        // 5.6.0 to 6.3.0.
+        // Link to rocFFT issue: https://github.com/ROCm/rocFFT/issues/507
+        if constexpr (rocfft_version_major == 1 && rocfft_version_minor == 0 &&
+                      (rocfft_version_patch > 22 && rocfft_version_patch < 31)) {
+            if (dom == dft::domain::COMPLEX &&
+                config_values.placement == dft::config_value::NOT_INPLACE && dimensions > 2) {
```
raphael-egan, line 272: Are we certain this is limited to `dimensions > 2`? Is a 2D case like
`lengths = {A, B}`
`inStrides = {B 1}`
`outStrides = {1, A}`
present in the test base?
Possibly relevant as well are 1D batch cases like
`length = N;`
`batch = M;`
`inStrides = 1; inDistance = N;`
`outStrides = M; outDistance = 1;`

s-Nick: Thank you for your review @raphael-egan 
The 2D configuration above is not directly tested. If added, rocfft backend skips the tests with the following output:
```bash
Skipping test because: "oneMKL: DFT/commit: function is not implemented rocfft requires a stride to be divisible by all dimensions associated with smaller strides!"
```
If a configuration that respect inputs requirement is provided, tests are passed. Similar configurations are already in the test suite, so I don't think adding this test is a priority.

For 1D there isn't a test that reproduces what you suggested. I added it to check. It is computed correctly with the input above, although the `outDistance` is computed internally to perform the operation correctly. In any case, both configuration do not cause any issue with rocFFT.

Therefore, I wouldn't change the if-statement condition

raphael-egan: Given that the exception is thrown by this project and not by rocFFT itself, I would recommend to add a comment like
`// rocFFT's functional status for problems like cfoA:B:1xB:1:A is unknown as of [commit hash] (due to project's implementation preventing testing thereof)`

Unless an explicit mention of that as a known limitation is found in rocFFT's documentation/release notes (I could not find any), it looks to me that the check triggering this exception to be thrown is present here to ensure consistency of subsequent stride-related implementation steps (sorting operations, in particular).

s-Nick: I agree, added comment in [e84a5b2](https://github.com/oneapi-src/oneMKL/pull/563/commits/e84a5b28c44e6823e5a08d1a06e40b3171858666)

raphael-egan (approved):

Issue #501, opened 2024-05-27
How to setup CMakeLists to use OneMKL on a GPU (Intel Arc/Nvidia sm_89)


salehjg:
Hi there,

I am having trouble using oneMKL with sycl on GPUs.
So far, I have tried to compile and [run this repository](https://github.com/salehjg/my-oneapi-workspace/tree/master/e00-onemkl-gemm) on Intel Arc A770, Nvidia RTX A2000 (sm_89) and Nvidia gtx1650 with `icpx` from HPC toolkit 2024.1 with the codeplay plugin for nvidia gpus.
For Intel Arc, I get an exception from oneMKL stating that the given device is not supported.
For Nvidia GPUs (A2000 and gtx1650) I get a SIGSEGV crash on:
```
sycl::_V1::detail::is_kernel_info_desc<sycl::_V1::info::kernel::num_args>::return_type sycl::_V1::kernel::get_info<sycl::_V1::info::kernel::num_args>() const:
```
It works on CPUs (on intel 13950HX, intel 6700,  and i7 4770K to be exact).

Could anyone please advise me on how to approach this problem? Should I build oneMKL from the source to enable GPU support?

Thank you,
Saleh

hjabird:
Hi Saleh,
Does 'sycl-ls' run successfully? And what version of CUDA are you using?

salehjg:
Thank you for the quick reply @hjabird 
```
nvidia-smi
Mon May 27 19:21:23 2024       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.171.04             Driver Version: 535.171.04   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA RTX 2000 Ada Gene...    Off | 00000000:01:00.0 Off |                  N/A |
| N/A   44C    P3              N/A /  60W |      8MiB /  8188MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
                                                                                         
+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
[...]

Issue #438, opened 2024-01-19
#include <oneapi/mkl.hpp>


tb5874:
hello, i'm windows10, oneapi user.

when i write oneapi code with onemkl, i can't use oneapi::mkl::rng::device namespace.
cmakelists.txt include compile options
/Qmkl:sequential
/Qmkl-ilp64:sequential

https://github.com/oneapi-src/oneMKL/blob/develop/include/oneapi/mkl.hpp
https://github.com/oneapi-src/oneMKL/blob/develop/include/oneapi/mkl/rng.hpp
in there, header did not include 'rng/device.hpp'.

i just want to know reason.

Rbiessy:
I'm not familiar with the rng domain but I can see that the [example](https://github.com/oneapi-src/oneMKL/blob/develop/examples/rng/device/uniform.cpp#L40) and [tests](https://github.com/oneapi-src/oneMKL/blob/develop/tests/unit_tests/rng/device/include/moments.hpp#L38) directly include this file as well. Maybe someone from @oneapi-src/onemkl-rng-write can confirm whether this is intentional?

aelizaro:
@Rbiessy, yes, having a standalone rng/device.hpp is intentional, as we wanted to stress out what device API are a completely different entity and has header-based implementation. So the expectation is that users may independently include rng/device.hpp functionality within sycl kernels and don't need to link to any binaries.

Rbiessy:
Closing as there are no issues, this is done by design.

PR #564, opened 2024-09-04
[RFC] oneMKL Interface renaming

Rbiessy:
Link to rendered README: https://github.com/Rbiessy/oneMKL/tree/romain/rfcs/onemkl_rename/rfcs/20240903-onemkl-interface-rename

The RFC proposes a solution to rename oneMKL Interface while moving it to the UXL foundation organization.

### DIFF
--- a/rfcs/20240903-onemkl-interface-rename/README.md
+++ b/rfcs/20240903-onemkl-interface-rename/README.md
@@ -1,0 +1,164 @@
+# OneMKL Interface renaming
+
+### Revision
+
+
+|Date       |Revision | Comments                                                                 |
+|-----------|---------|--------------------------------------------------------------------------|
+|  20240903 |  1.0    | Initial version                                                          |
+
+## Motivation
+
+As oneMKL interface is moving to the UXL foundation we should make sure that it
+does not reuse the same name as the existing [Intel oneMKL
+product](https://www.intel.com/content/www/us/en/developer/tools/oneapi/onemkl.html).
+
+We have been discussing 2 solutions to avoid this issue:
+1. Renaming the [oneMKL](https://github.com/oneapi-src/oneMKL) GitHub project to
+   a new name to be discussed and replacing or clarifying the occurrences of
+   "oneMKL" in that repository.
+2. Using this opportunity to also split this GitHub project per domain. In this
+   case the names that are advertised would be based on each domain of oneMKL
+   like `one<Domain>`.
+
+We think that the first solution should be enough to answer the main concern. It
+is not clear whether the second solution could better answer the need of users
+so we are not planning to split oneMKL per domain unless there is clear and
+strong feedback from users.
+
+**We aim to agree on a solution by October 4, 2024 and have the proposal
+implemented by the end of November 2024.**
+
+There has been a number issues of where we have had to clarify the differences
+between oneMKL Interface and the Intel oneMKL product
+[#501](https://github.com/oneapi-src/oneMKL/issues/501#issuecomment-2134681621);
+[#377](https://github.com/oneapi-src/oneMKL/issues/377);
+[#253](https://github.com/oneapi-src/oneMKL/issues/253);
+[#222](https://github.com/oneapi-src/oneMKL/issues/222);
+[#211](https://github.com/oneapi-src/oneMKL/issues/211);
+[#207](https://github.com/oneapi-src/oneMKL/issues/207);
+[#206](https://github.com/oneapi-src/oneMKL/issues/206). Many of these issues
+are from 2022. I believe this is still an issue today since this has been an
+issue on multiple occasion internally when Codeplay started to contribute to
+oneMKL Interface. The oneMKL Interface
+[README](https://github.com/oneapi-src/oneMKL?tab=readme-ov-file#onemkl) already
+explains the differences between oneMKL Interface and Intel oneMKL product but
+this is not enough for people who are not working directly on the projects.
+
+## Outline
+
+1. [Introduction](#introduction)
+2. [Proposal](#proposal)
+3. [User impact](#user-impact)
+4. [Open questions](#open-questions)
+
+## Introduction
+
+oneMKL Interface has historically been implemented and supported by Intel. As
+the project is moving to the UXL foundation it we want to avoid using the name
+of an Intel product. This RFC describes the solution of renaming oneMKL
+Interface to a new name.
+
+## Proposal
+
+The main purpose of this RFC is to agree on a new name for oneMKL Interface.
+Some of the suggested names are:
+* **oneMath**
+* **oneSLA** (SYCL Linear Algebra)
+
+Other suggestions are welcomed.
+
+The suggested solution is to proceed in the following steps:
+1. The UXL foundation agrees on the new name.
+2. Codeplay submits a oneMKL Interface PR to:
+   * Update the root README to use the new name, with a mention that the project
+     was formerly called oneMKL Interface.
+   * Update the references to "oneMKL" and `onemkl_` in the documentation as
+     seen in the first few lines of
+     [docs/onemkl-datatypes.rst](https://github.com/oneapi-src/oneMKL/blob/develop/docs/onemkl-datatypes.rst?plain=1#L1)
+     for instance.
+   * Update occurrences of "onemkl" in internal functions such as
+     [onemkl_cublas_host_task](https://github.com/oneapi-src/oneMKL/blob/1ce98a699f93bd3a78350269b2e34d822fe43b91/src/blas/backends/cublas/cublas_task.hpp#L77).
+   * Update macros such as include guards and other internal macros like
+     `ONEMKL_EXPORT` to use the new name.
+   * Rename CMake targets `onemkl` and `onemkl_<domain>_<backend>` to use the
+     new name. The existing targets name can be added with a deprecation
+     messages for anyone using them. See the section on [CMake target
+     deprecation](#cmake-deprecated-target) for more details.
+3. Codeplay submits a oneAPI-spec PR to rename the occurrences of "oneMKL" to
+   the new name.
+4. Once the PRs are approved, Codeplay transfers the
+   [oneMKL](https://github.com/oneapi-src/oneMKL) GitHub project to the
+   [uxlfoundation](https://github.com/uxlfoundation) organization under the new
+   name. We use the
+   [transferring](https://docs.github.com/en/repositories/creating-and-managing-repositories/transferring-a-repository)
+   feature from GitHub so the links from the previous repository are redirected
+   to the new one.
+5. The PRs from the step 2 and 3 are merged.
+
+We are not planning to rename the occurrences of "mkl" such as the `oneapi::mkl`
+namespace, the `include/oneapi/mkl` folder or the `include/oneapi/mkl.hpp` file.
+Whether this is needed is an open question.
+
+### CMake target deprecation
+
+CMake allows to set a
+[`DEPRECATION`](https://cmake.org/cmake/help/latest/prop_tgt/DEPRECATION.html)
+property on a target which will print a custom message whenever the target is
+used. The property cannot be set on an [alias
+target](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#alias-targets)
+as they are read-only. The property can be set on an imported target instead
+like the example below:
+
+```cmake
+add_library(onemath lib.cpp) # New main target, for the example
+
+add_library(onemkl INTERFACE IMPORTED) # onemkl works like an alias of onemath which can have different properties
+target_link_libraries(onemkl INTERFACE onemath)
+set_target_properties(onemkl PROPERTIES DEPRECATION "onemkl target is deprecated, please use onemath instead")
+
+add_executable(main main.cpp)
+target_link_libraries(main PUBLIC onemkl) # Prints a warning at CMake configuration time
+```
+
+The same solution can be used for the `onemkl_<domain>_<backend>` targets. This does
+not add any extra targets to the generated `Makefile` or `build.ninja` files so
+the library will not be built twice.
+
+### Other Considered Approaches
+
+Another considered approach is to split the existing oneMKL Interface per domain
+like so: oneBLAS, oneLAPACK, oneDFT, oneRNG, oneSPARSE. This shifts the need of
+renaming "oneMKL Interface" as the main visible names will be based on the
+domain. It is not clear whether this better answers the users needs.
+
+With this approach the suggested solution is to have a common repository for
+common types (`transpose`, `uplo`, `diag`, `side`, `offset`, `index_base`,
+`layout`), exceptions and some CMake logic. Each domain would have its own
+repository automatically pulling the common headers. Another repository could be
+created to automatically pull multiple domains which would mimic the behavior of
+the current oneMKL Interface.
+
+We are not planning to proceed with this approach unless users express a strong
+preference.
+
+## User impact
+
+The suggested solution does not break any existing code.
+* The repository is transferred using the [GitHub
+  transfer](https://docs.github.com/en/repositories/creating-and-managing-repositories/transferring-a-repository)
+  feature so users accessing or pulling from
+  https://github.com/oneapi-src/oneMKL will be redirected to the new link.
+* The changes in oneMKL do not affect the public API. The macros renamed or the
+  header files in the `detail` folders renamed should not be used outside of the
+  oneMKL Interface project.
+* The CMake changes will still provide the same targets but will print a warning
+  message if users use targets with the `onemkl` name.
+
+## Open questions
+
+* Other suggestions for new names are welcomed.
+* Is it needed to rename the occurrences of "mkl"?
+   * This will have a bigger impact and require more time to complete.
+   * It should be possible to rename these occurrences without any breaking
+     change. This would need to be further investigated.
--- a/rfcs/20240903-onemkl-interface-rename/README.md
+++ b/rfcs/20240903-onemkl-interface-rename/README.md
@@ -70,7 +70,9 @@ Other suggestions are welcomed.
 
 The suggested solution is to proceed in the following steps:
 1. The UXL foundation agrees on the new name.
-2. Codeplay submits a oneMKL Interface PR to:
+2. Codeplay submits a oneAPI-spec PR to rename the occurrences of "oneMKL" to
+   the new name.
+3. Codeplay submits a oneMKL Interface PR to:
    * Update the root README to use the new name, with a mention that the project
      was formerly called oneMKL Interface.
    * Update the references to "oneMKL" and `onemkl_` in the documentation as
@@ -85,8 +87,6 @@ The suggested solution is to proceed in the following steps:
      new name. The existing targets name can be added with a deprecation
      messages for anyone using them. See the section on [CMake target
      deprecation](#cmake-deprecated-target) for more details.
-3. Codeplay submits a oneAPI-spec PR to rename the occurrences of "oneMKL" to
-   the new name.
 4. Once the PRs are approved, Codeplay transfers the
    [oneMKL](https://github.com/oneapi-src/oneMKL) GitHub project to the
    [uxlfoundation](https://github.com/uxlfoundation) organization under the new
--- a/rfcs/20240903-onemkl-interface-rename/README.md
+++ b/rfcs/20240903-onemkl-interface-rename/README.md
@@ -61,10 +61,12 @@ to a new name.
 
 ## Proposal
 
-The main purpose of this RFC is to agree on a new name for oneMKL Interface.
-Some of the suggested names are:
+The main purpose of this RFC is to agree on a new name for the oneMKL specification and oneMKL Interface.
+Some of the suggested names for the implementations are:
 * **oneMath**
+  * The specification would be renamed oneMath Specification
 * **oneSLA** (SYCL Linear Algebra)
+  * The specification would be renamed oneSLA Specification
 
 Other suggestions are welcomed. The name **oneMath** will be chosen if there are
 no objections by October 4, 2024.
@@ -163,3 +165,12 @@ The suggested solution does not break any existing code.
    * This will have a bigger impact and require more time to complete.
    * It should be possible to rename these occurrences without any breaking
      change. This would need to be further investigated.
+* Should the specification and the existing implementation have different names?
+  * Currently both the specification and implementation are named based on
+    "oneMKL" which suggests that the oneMKL Interface is the main or only
+    implementation of the specification. If the goal of the specification is to
+    encourage for multiple implementations to co-exist then it should be named
+    differently than the implementation.
+  * Given the nature of the project that allows for multiple backends I don't
+    see any value in encouraging multiple implementations as of today.
+  * Using multiple names may create more confusion.
--- a/rfcs/20240903-onemkl-interface-rename/README.md
+++ b/rfcs/20240903-onemkl-interface-rename/README.md
@@ -6,6 +6,7 @@
 |Date       |Revision | Comments                                                                 |
 |-----------|---------|--------------------------------------------------------------------------|
 |  20240903 |  1.0    | Initial version                                                          |
+|  20240916 |  1.1    | Rename occurrences of "mkl" and add an open question                     |
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


al42and:
My 2 cents as a user:

- Re: Is it needed to rename the occurrences of "mkl"? 
  - I don't see how keeping `mkl` around reduces the confusion. "To use oneXYZ library (not to be confused with Intel oneMKL) library, we include `oneapi/mkl.hpp` file, use functions from `oneapi::mkl` namespace, and link to `oneXYZ` CMake target".
  - If the Intel and UXL APIs are in sync, then keeping the old name in the code could help users code duplication for handling different namespaces / macros prefixes / include paths. But if the APIs diverge, having them use the same namespace / macros would become error-prone.
  -  If `mkl` is fully expunged from the library code, it's better to do it in one swoop (keeping the old symbols deprecated but available for a while). It would be unpleasant to deal with a mix of old and new naming that's changing from version to version.
- As long as installing a single domain is supported, I see no benefits for the user from splitting the library. No downsides either.
  -  Where would `mkl::stats` and `mkl::vm` go if the library is split by domain?
- Deprecation / renaming are ok for us in terms of updating our code to use the new name. But that's mainly because we don't currently have open-source oneMKL in the typical builds for end-users.
- We had very little actual issues with the old name, although I personally agree that the current situation is confusing; and ["There are three oneMKLs"](https://github.com/oneapi-src/oneMKL/issues/501#issuecomment-2134681621) strongly indicates the need for better naming too.
- Naming:
  - oneSLA: my only association with SLA is "service-level agreement"; two (FFT and RNG) out of five supported domains are not linear algebra; "SYCL" in the name is redundant since most oneAPI projects are SYCL-related.
  - I like oneMath is better, even though it departs from the `one[A-Z]{3}` scheme used by other projects.
  - Any `one[A-Z][A-Za-z]+` name could be confused with Intel product, IMO.
- Would it make sense to also introduce a clear distinction between the library (code) and the interface (spec)? If the spec is supposed to be open, it's not good/fair towards (hypothetical) other implementations to so strongly mark them as second-class.

Rbiessy:
Thank you for your valuable feedback @al42and!

Regarding the occurrences of "mkl", as far as I know the plan is to keep the API of the Intel oneMKL product and the open-source interface as close as possible. Introducing a new namespace and header files for the interface would break the ability to transition between the product and the interface seamlessly (without a change in the product) so I have left it as an open question.

> Where would mkl::stats and mkl::vm go if the library is split by domain?

They would also have their own repository like oneStats and oneVM. I didn't mention them since they have not been implemented yet.

> Would it make sense to also introduce a clear distinction between the library (code) and the interface (spec)?

If I understand you correctly, you are suggesting the specification should not mention the name of the implementation. So we would provide the specification for "XYZ" and the implementation would be an implementation of "XYZ" and have a different name. It's a fair point. I am concerned this will only add more confusion. Given the nature of the project with the different backends I'm not sure another implementation would be that valuable to be honest.

al42and:
> Given the nature of the project with the different backends I'm not sure another implementation would be that valuable to be honest.

Agree with you on the practical aspect, but I think the messaging is important too. If oneXYZ is a _library_, then it is a specific piece of code with a documented interface (which someone might decide to copy). If oneXYZ is a _standard_, it strongly implies that there can be several implementations of it (and they all are equal). E.g., Python vs. C++.

[oneAPI spec page](https://www.oneapi.io/spec/) implies there can be multiple implementations of the spec: "The oneAPI initiative encourages collaboration on the **specification** and compatible oneAPI **implementations** across the ecosystem." (emphasis mine). But that may be me reading too much into it.

> as far as I know the plan is to keep the API of the Intel oneMKL product and the open-source interface as close as possible

Then, is Intel oneMKL an implementation of the oneXYZ API and also a backend of oneXYZ library? Since oneMKL implements oneXYZ, why is oneXYZ using `mkl` namespace?

Or, if it is oneXYZ following Intel oneMKL, then what's the point of the working groups, discussions and all that if it just copies whatever Intel oneMKL does?

*Disclaimer*: The flow of ideas is never as strictly uni-directional as I'm portraying above, especially early on in the project's life. But it's good to have a vision of how things are supposed/intended to work.

> Given the nature of the project with the different backends I'm not sure another implementation would be that valuable to be honest.

If Intel oneMKL follows the oneXYZ spec, then there are two implementations already: Intel oneMKL and the open-source oneXYZ we have here :)

> Introducing a new namespace and header files for the interface would break the ability to transition between the product and the interface seamlessly (without a change in the product) so I have left it as an open question.

Can't say it's totally seamless now. Pretty close, but not drop-in (not to mention the build system).

If it's just a name of the namespace, then `using` would solve this problem with about the same effort as [the macro/enum problem is solved in GROMACS now](https://gitlab.com/gromacs/gromacs/-/blob/d3c4fb99663d9adfe199b552e784ad39a40375b4/src/gromacs/fft/gpu_3dfft_sycl_mkl.cpp#L73-85).

**UPDATE**: I don't want to derail this PR into a discussion about project governance, but since the goal of the proposal is to disambiguate three entities, I feel like an informed decision requires a clear picture of how these entities are related.

rfcs/20240903-onemkl-interface-rename/README.md
```diff
@@ -0,0 +1,164 @@
...
+the project is moving to the UXL foundation we want to avoid using the name of
+an Intel product. This RFC describes the solution of renaming oneMKL Interface
+to a new name.
+
+## Proposal
+
+The main purpose of this RFC is to agree on a new name for oneMKL Interface.
+Some of the suggested names are:
+* **oneMath**
+* **oneSLA** (SYCL Linear Algebra)
```
mkrainiuk, line 72: 'oneMath' looks good to me. I'm not sure 'oneSLA' will be accurate in this case, since FFT, RNG, and Vector math are not technically part of Linear Algebra

rfcs/20240903-onemkl-interface-rename/README.md
```diff
@@ -0,0 +1,164 @@
...
+     for instance.
+   * Update occurrences of "onemkl" in internal functions such as
+     [onemkl_cublas_host_task](https://github.com/oneapi-src/oneMKL/blob/1ce98a699f93bd3a78350269b2e34d822fe43b91/src/blas/backends/cublas/cublas_task.hpp#L77).
+   * Update macros such as include guards and other internal macros like
+     `ONEMKL_EXPORT` to use the new name.
+   * Rename CMake targets `onemkl` and `onemkl_<domain>_<backend>` to use the
+     new name. The existing targets name can be added with a deprecation
+     messages for anyone using them. See the section on [CMake target
+     deprecation](#cmake-deprecated-target) for more details.
+3. Codeplay submits a oneAPI-spec PR to rename the occurrences of "oneMKL" to
```
mkrainiuk: Does it make sense to update spec first (somewhere in the upcoming next spec revision), and after it change the opensource project?

rfcs/20240903-onemkl-interface-rename/README.md
```diff
@@ -0,0 +1,164 @@
...
+  header files in the `detail` folders renamed should not be used outside of the
+  oneMKL Interface project.
+* The CMake changes will still provide the same targets but will print a warning
+  message if users use targets with the `onemkl` name.
+
+## Open questions
+
+* Other suggestions for new names are welcomed.
+* Is it needed to rename the occurrences of "mkl"?
+   * This will have a bigger impact and require more time to complete.
```
mkrainiuk: I believe it will also affect specification

rfcs/20240903-onemkl-interface-rename/README.md
```diff
@@ -0,0 +1,164 @@
...
+The same solution can be used for the `onemkl_<domain>_<backend>` targets. This does
+not add any extra targets to the generated `Makefile` or `build.ninja` files so
+the library will not be built twice.
+
+### Other Considered Approaches
+
+Another considered approach is to split the existing oneMKL Interface per domain
+like so: oneBLAS, oneLAPACK, oneDFT, oneRNG, oneSPARSE. This shifts the need of
+renaming "oneMKL Interface" as the main visible names will be based on the
+domain. It is not clear whether this better answers the users needs.
```
mkrainiuk, line 155: I have one idea to share about how to improve rebase simplicity when multiple people work on different domains in one repo at the same time, as an option we might consider switching merge policy from 'rebase and merge' to 'merge commits'.

mkrainiuk (commented):
The RFC looks good to me, thank you for summarizing all this work. I have several comments/questions.

rfcs/20240903-onemkl-interface-rename/README.md
Rbiessy: Some developers, myself included, are not fond of using merge commits in the main branch. I find it makes the history harder to read and I don't see what difference it would make for developers working on a single domain.

rfcs/20240903-onemkl-interface-rename/README.md
Rbiessy: The intention is to review the 2 PRs in parallel and merge them roughly at the same time. I've swapped the 2 in https://github.com/oneapi-src/oneMKL/pull/564/commits/9dc802e460837abce61e1c78513dd4239f22a020

Rbiessy:
@al42and these are great questions. They  are about the relationship between the oneMKL Interface and the Intel oneMKL product. I believe others can provide a better answer, maybe @mkrainiuk, @spencerpatty or @rscohn2?

etienne-renault:
This RFC LGTM. I do have few questions / comments : 

- Regarding the "opportunity to split this GitHub project per domain". I'm also not sure about this, and I would say that it will add more confusion.
- I'm puzzled with the usage of the prefix "one..." which is very Intel related. I would suggest to move to something drastically different to remove confusion. Otherwise, something `<prefix>Math` looks good to me (and I agree about the inadequation of the <prefix>SLA name)
- Regarding the renaming of  the occurrences of  `oneapi::mkl` namespace, the `include/oneapi/mkl` folder or the `include/oneapi/mkl.hpp` : i would recomand to perform this renaming to avoid any confusion / future troubles.

colleeneb:
I agree with comments above about oneMath is preferred to oneSLA since not everything is linear algebra.

Just to check my understanding is correct: There's the "oneMKL specification" which is the [specification](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/elements/onemkl/source/), "oneMath" (calling it by the new name for now) which is an implementation of the oneMKL specification with many backends. Then there is the Intel product oneMKL, which is an implementation of the oneMKL specification by Intel.

Since this is an implementation of the oneMKL specification, I think it makes sense to continue to use the `oneapi::mkl::` namespaces since that's in the specification. However, the term "MKL" is so connected in people's minds with Intel that I think it will possibly always confuse people unless the namespaces are changed to some other name. But changing to another name would break a lot of people's code so I don't know if that's practical. 

Overall I think the RFC looks good (good to try to prevent confusion and distinguish from Intel's implementation!), just that I think there will still be confusion since the term "MKL" is still in the specification and it's so connected with Intel. 

(If the oneapi::mkl namespaces change in the future, one option is to match the C++26 APIs, at least for linear algebra: https://en.cppreference.com/w/cpp/numeric/linalg . Just as an option, but not something I'd necessarily suggest.)

al42and:
> They are about the relationship between the oneMKL Interface and the Intel oneMKL product. 

To clarify, I don't want to derail this PR into a discussion about project governance, but since the goal of the proposal is to disambiguate three entities, I feel like an informed decision requires a clear picture of how all these entities are related.

> Given the nature of the project with the different backends I'm not sure another implementation would be that valuable to be honest.

Another use case to consider here could be 3rd party domain-specific libraries implementing the oneAPI interface on their own. E.g., what if https://github.com/intel/double-batched-fft-library decides to adopt oneAPI interface for DFT? If oneAPI gains traction, that would not be unreasonable (just like Intel MKL has FFTW-like interface).

Rbiessy:
Thank you for the good feedback so far. I note the preference for `oneMath` for now although the prefix `one` is debatable. The `oneAPI` specification has been moved as in to the UXL foundation repository and I am not aware of a plan to rename this so I kept this prefix for now.

I agree that not every domain can be classified as "linear algebra". This is one of the reason for the initial proposition to split per domain, there isn't much in common between all of the domains so it is hard to come up with a new meaningful name.

Regarding the Intel oneMKL product, I try to see it as 2 backends (for MKLCPU and MKLGPU) like any others. Technically it has its own [C specification](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2024-2/overview.html) and [C++ specification](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2024-2/overview.html). I also think it would be best to remove the occurrences of "mkl". I could write more details about what this would look like. It would be useful to get some feedback from the Intel oneMKL product owners on this and the suggestion of using different names for the specification and implementation.

rfcs/20240903-onemkl-interface-rename/README.md
gajanan-choudhary: I second that, I personally tend to avoid merge commits at all costs exactly because the history is harder to figure out. Having a clean history also means the pretty output of `git log --all --decorate --oneline --graph` is easier to decipher.

gajanan-choudhary:
> Regarding the Intel oneMKL product, I try to see it as 2 backends (for MKLCPU and MKLGPU) like any others. Technically it has its own [C specification](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2024-2/overview.html) and [C++ specification](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2024-2/overview.html).

Calling them _specifications_ (or in [Wikipedia terms](https://en.wikipedia.org/wiki/Programming_language_specification), _standards_) is technically incorrect. Those are just documentation pages of the C and SYCL APIs in the Intel oneMKL library.

spencerpatty:
So I am part of Intel oneMKL team and I have discussed this concern about the name oneMKL being in 3 places and the potential for separating the Specification and the Interfaces project in name from the Intel oneMKL Product with several people from oneMKL team and I think we all see the concerns and feel similarly to those asking about if names could change.
 
To be clear, I am talking about the following potential changes:
 
- **oneMKL Specification**  -> **oneMath Specification**  
    - (and using `oneapi::math::`  namespace instead of `oneapi::mkl::` namespace)
- **oneMKL Interfaces** Project -> **oneMath Interfaces** Project
    - (again using `oneapi::math::` namespace instead of `oneapi::mkl::` namespace)
- **Intel oneMKL Library** (or Product) -> stays **Intel oneMKL Library** (or Product)
    - (using `oneapi::mkl::` namespaces)

 
It makes a lot of sense to me (and at least to those I have talked with) to make this kind of change.  We at Intel introduced **(now using new proposed names)** the **oneMath Specification** to try to encode a common set of math related APIs that could be supported by few or many different library implementations that allow applications to move beyond a single software/hardware configuration and be more productive on multiple hardware/software systems.  **Intel oneMKL Library** already had a wide range of supported math domains and so we modelled much of the original specification around existing and future planned solutions, and now that it is part of the UXL foundation, we hope it will continue to evolve according to many different users needs into a common C++ language solution.
 
The **oneMath Interfaces** project was always meant to be a direct implementation of the **oneMath specification** and move with it. It was also designed to have many potential backends which would facilitate the end goal of making the hardware/software switch simple.  **Intel oneMKL Library** provides two such backend implementations for the **oneMath Interfaces**: namely  `MKLCPU` and `MKLGPU`.  There were many tricky things that came from having the **Intel oneMKL library** as backends even though both use the same namespaces, but as I understand it, those have been worked through already, so it is not as tricky anymore ...
 
On the other hand, the **Intel oneMKL Library** relationship with the **oneMath Specification** has always been a little bit more complicated.  Our goal has been to be compliant with the specification, but we also often have many more things available than are in the specification, and in fact often do some of the design work that leads to a new function being introduced to the specification through actually implementing things and getting customer feedback on our implementations before adding it.  So from our perspective, we are both behind and in front of the specification. Additionally, there are many things which we offer using particular hardware features which may not belong in the specification as they are not general enough for common implementations. The question for us is this: "Can we continue considering ourselves as an implementation of the **oneMath specification** if we do not have exactly the same interfaces (say that differs in a namespace) ?  The answer is a resounding: _**maybe**_ :) - haha!    We definitely want to try to keep our interfaces consistent with the specification as much as we overlap, and we might consider to add additional `oneapi::math::` based interfaces to our product to be exactly compliant, or we might be able to come up with some sort of aliasing or namespace manipulation to convert what is available in the library into what is defined in the spec and written in the application…  On the other hand, there are several existing applications which are written with `oneapi::mkl::` and using the **oneMath Interfaces**, so either they would have to change their code or a similar conversion might need to be constructed for them in their applications.
 
All that said I think it makes sense (if that is the direction this UXL group discussion is heading) to move away from the mkl namespace in the **oneMath Specification** and in the **oneMath Interfaces** project.

Rbiessy:
Thank you all for the feedback so far. I have added some clarifications on the impact of renaming the occurrences of "mkl". I have bumped the estimation to implement this change by January 2025 as a consequence.
I have also added an open question on whether to use a different name for the specification.

rfcs/20240903-onemkl-interface-rename/README.md
```diff
@@ -0,0 +1,205 @@
...
+    new name. The existing targets name can be added with a deprecation
+    messages for anyone using them. See the section on [CMake target
+    deprecation](#cmake-deprecated-target) for more details. The namespace of
+    the exported and installed targets are changed to use the new name.
+  * The namespace `oneapi::mkl` is deprecated and aliased to the new namespace
+     name.
+  * The main header and domain headers are deprecated and include the new headers as shown below:
+    * `include/oneapi/`
+      * `mkl.hpp` -> `onemath.hpp`
+      * `mkl/` -> `onemath/`
```
spencerpatty: why onemath.hpp and not math.hpp ? and math/ instead of onemath/ ?  since it is in the include/oneapi/ folder where the "one" presumably comes from ?

rfcs/20240903-onemkl-interface-rename/README.md
```diff
@@ -0,0 +1,205 @@
...
+   case the names that are advertised would be based on each domain of oneMKL
+   like `one<Domain>`.
+
+We think that the first solution should be enough to answer the main concern. It
+is not clear whether the second solution could better answer the need of users
+so we are not planning to split oneMKL per domain unless there is clear and
+strong feedback from users.
+
+**We aim to agree on a solution by October 4, 2024 and have the proposal
+implemented by the end of January 2025.**
```
spencerpatty, line 32: seems to me that a decision about changing namespaces and not just the repo name should be considered for a longer period of time and advertised quite widely to get feedback. I was thinking several months for such a change to be discussed and decided on, so decision would be made more around Jan 2025 and then implemented.  Or do we have a rush on this ?  honestly it is a far reaching choice, so might suggest to give it time for advetisement ... :)

Rbiessy: I feel that this topic has been discussed for some time already. One of the goal is to move the oneMKL Interface repository to the UXL foundation organization by the end of the year and we thought it would be a good timing to also rename it.
I don't see why we would several months to agree on the renaming. Do you have a specific concern? As far as I can tell the main contributors agree on the general idea already. The impact on user is manageable, they should only need to rename a CMake namespace to get it to compile, see the section [user impact](https://github.com/Rbiessy/oneMKL/tree/romain/rfcs/onemkl_rename/rfcs/20240903-onemkl-interface-rename#user-impact).
We're planning to present this topic in the next UXL open-source WG on September 24. I think that will help us determine if people have concerns and need more time.

rfcs/20240903-onemkl-interface-rename/README.md
Rbiessy: I have a small preference for `onemath` as I find `math` is just too generic. I don't think it is an issue to repeat the `one` prefix. I have renamed it to https://github.com/oneapi-src/oneMKL/pull/564/commits/1182a951fb51a347a2a0a40437d430a759bdf55e since I seem outvoted for now.

al42and: Other oneAPI parts don't repeat the prefix: [`<oneapi/dpl/execution>`](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.4-provisional-rev-1/elements/onedpl/source/parallel_api/execution_policies), [`"oneapi/dal.hpp"`](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.4-provisional-rev-1/elements/onedal/source/programming_model/end_to_end_example), [`<oneapi/tbb/collaborative_call_once.h>`](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.4-provisional-rev-1/elements/onetbb/source/algorithms/functions/collaborative_call_once_func).

rfcs/20240903-onemkl-interface-rename/README.md
```diff
@@ -0,0 +1,205 @@
...
+  * Update macros such as include guards and other internal macros like
+    `ONEMKL_EXPORT` to use the new name.
+  * Rename CMake targets `onemkl` and `onemkl_<domain>_<backend>` to use the
+    new name. The existing targets name can be added with a deprecation
+    messages for anyone using them. See the section on [CMake target
+    deprecation](#cmake-deprecated-target) for more details. The namespace of
+    the exported and installed targets are changed to use the new name.
+  * The namespace `oneapi::mkl` is deprecated and aliased to the new namespace
+     name.
+  * The main header and domain headers are deprecated and include the new headers as shown below:
```
Rbiessy, line 103: I think this will run into this issue with the rng domain: https://github.com/oneapi-src/oneMKL/issues/438
I assumed `rng/device.hpp` should be included by `rng.hpp` but I can see it is explicitly mentioned in the [README](https://github.com/oneapi-src/oneMKL?tab=readme-ov-file#device-api) that the file should be included separately. Can someone from @oneapi-src/onemkl-rng-write confirm that this is intentional? If so we should make sure that `include/oneapi/mkl/rng/device.hpp` is deprecated and includes the new equivalent header.

aelizaro: @Rbiessy, yes, having a standalone rng/device.hpp is intentional, as we wanted to stress out what device API are a completely different entity and has header-based implementation. So the expectation is that users may independently include rng/device.hpp functionality within sycl kernels and don't need to link to any binaries.

Rbiessy: Thank you for the quick reply, I have added it to the list of headers that will be deprecated so users can still include it: https://github.com/oneapi-src/oneMKL/pull/564/commits/637610d8cabd400c20f9f32d90704528473e8211

rfcs/20240903-onemkl-interface-rename/README.md
aelizaro: and SLA first association is service Level Agreement, not Linear Algebra

Rbiessy:
FYI we discussed this RFC in the UXL open-source WG on September 24. There were no major concern on the RFC, I have applied some feedback in https://github.com/oneapi-src/oneMKL/pull/564/commits/3847831a75b1c15e0aa09190e89458f4bacee153. The outcomes are:

- Update the RFC to mention that the `onemkl_<domain>` CMake targets will also be renamed, like the other CMake targets using the onemkl name.
- @al42and we clarified that the goal of the specification is not to encourage for more implementations to exist and we plan to keep using the same name in the specification and implementation. The preference is for users to contribute to the existing implementation.
- @spencerpatty we discussed how long should an RFC be up for review before being accepted. The members present suggested that October 4 should be enough time which coincide with the deadline we already planned. I'm not seeing any concerns that haven't been answered here. Unless there are more discussions here I think this deadline is reasonable.

rfcs/20240903-onemkl-interface-rename/README.md
spencerpatty: Reading the description below of discussion in Sept 24 UXL WG meeting, I accept the suggestion and  Oct 4 is fine with me then.  :)

Rbiessy:
Today is the deadline for this RFC. All the questions and comments have been addressed as far as I can tell. @mkrainiuk is the next step to merge this PR?
We are planning to start working on the suggested changes for the specification and the implementation ASAP.

mkrainiuk (approved):

mkrainiuk:
> Today is the deadline for this RFC. All the questions and comments have been addressed as far as I can tell. @mkrainiuk is the next step to merge this PR? We are planning to start working on the suggested changes for the specification and the implementation ASAP.

Hi @Rbiessy, yes, before merging RFC could you please highlight what is the name we selected as the final one? RFC contains information about at least two options, but not what was selected at the end.

Rbiessy:
Do you mean it should be clarified in the RFC itself? Currently the document says:
```quote
The name **oneMath** will be chosen if there are no objections by October 4, 2024.
```

Rbiessy:
Who can give a second approval on this RFC so that we merge it @oneapi-src/onemkl-admin, @spencerpatty, @rscohn2?

rscohn2 (approved):

spencerpatty (commented):
I also approve of this

PR #566, opened 2024-09-05
Add generic device and initial support in portBLAS

s-Nick:
# Description

 - oneMKL Interfaces currently only supports known targets: Intel CPU/GPU, AMD GPU, Nvidia GPU
- This PR:
    - Enables a new generic target
    - Enables the generic target to use the portBLAS backend
    - Adds documentation
    - Relates to https://github.com/oneapi-src/oneMKL/issues/542

# Checklist

## All Submissions

  - I don't have a non Intel/NVIDIA/AMD device to test with. Tests are successful when forcing Intel devices to go through `generic_device` path.

Following logs show that all domains still work properly:
[arc_mkl_BLAS_log.txt](https://github.com/user-attachments/files/16886641/arc_mkl_BLAS_log.txt)
[arc_mkl_DFT_log.txt](https://github.com/user-attachments/files/16886644/arc_mkl_DFT_log.txt)
[arc_mkl_lapack_log.txt](https://github.com/user-attachments/files/16886646/arc_mkl_lapack_log.txt)
[arc_mkl_rng_log.txt](https://github.com/user-attachments/files/16886648/arc_mkl_rng_log.txt)

## New features

The generic device doesn't requires new tests.


src/include/function_table_initializer.hpp
```diff
@@ -99,8 +99,17 @@ class table_initializer {
                 break;
         }
         if (!handle) {
-            std::cerr << ERROR_MSG << '\n';
-            throw mkl::backend_not_found();
+#ifndef ENABLE_PORTBLAS_BACKEND
```
Rbiessy: I think it would look cleaner if this `ifndef` is separated to create a variable like
```cpp
#if defined(ENABLE_PORTBLAS_BACKEND)
static constexpr bool is_generic_device_supported = true;
#else
static constexpr bool is_generic_device_supported = false;
#endif
```
That will keep the runtime `if` cleaner.

s-Nick: Thank you for the comment @Rbiessy I addressed it in [3466d6f](https://github.com/oneapi-src/oneMKL/pull/566/commits/3466d6f87bc5e0f720dd9a538cd491c5439fe893)

Rbiessy (approved):
That looks good to me, thanks!

s-Nick:
@oneapi-src/onemkl-maintain Hi all, in this PR I am adding a new generic device to allow portable libraries such as portBLAS and portFFT to be compiled and run on devices that are not known to us right now, but that have an OpenCL backend available. 
 
Changes are few and simple, but since I am making an important one to `table_initializer` and it touches all domains, I would like to have your approval on it. These changes don't cause any issue to current backends as you can check from logs attached.

tests/unit_tests/main_test.cpp
```diff
@@ -101,15 +101,17 @@ int main(int argc, char** argv) {
             auto plat_devs = plat.get_devices();
             for (auto dev : plat_devs) {
                 try {
-                    /* Do not test for OpenCL backend on GPU */
-                    if (dev.is_gpu() && plat.get_info<sycl::info::platform::name>().find(
-                                            "OpenCL") != std::string::npos)
+                    unsigned int vendor_id =
+                        static_cast<unsigned int>(dev.get_info<sycl::info::device::vendor_id>());
+                    /* Do not test for OpenCL backend on Intel GPU */
```
andrewtbarker, line 106: Does this logic continue to do what we want on AMD and nVidia GPUs?

src/include/function_table_initializer.hpp
```diff
@@ -59,14 +59,20 @@ class table_initializer {
     using dlhandle = std::unique_ptr<LIB_TYPE, handle_deleter>;
 
 public:
-    function_table_t &operator[](oneapi::mkl::device key) {
-        auto lib = tables.find(key);
+    function_table_t &operator[](std::pair<oneapi::mkl::device, sycl::queue> device_queue_pair) {
```
andrewtbarker, line 62: It's not clear to me why this change is necessary. Where do we need the `sycl::queue` in the lookup tables?

andrewtbarker (commented):
This is a useful change, I have just a couple questions.

tests/unit_tests/main_test.cpp
s-Nick: Thank you for your review @andrewtbarker 
As far as I know NVIDIA and AMD GPUs don't use OpenCL backends, so I expect they continue to work as intended. In any case, I am happy to add a condition to check if the vendor is one of the currently known one and skip the OpenCL backend.

src/include/function_table_initializer.hpp
s-Nick: Since I moved the `unsupported_device` exception in the table initializer, I need the queue to query the device name. This is the only reason why I need the `sycl::queue` here.

tests/unit_tests/main_test.cpp
Rbiessy: > Does this logic continue to do what we want on AMD and nVidia GPUs?

What is the behavior that we want on AMD and Nvidia exactly? We've been wondering whether we should remove this check entirely since it's not clear why would we want to skip OpenCL backends. To me users should use `ONEAPI_DEVICE_SELECTOR` instead.

src/include/function_table_initializer.hpp
andrewtbarker: That makes sense. It seems like a large change for just the exception - I would support reverting the function table and throwing a more generic exception (or maybe implement `mkl::unsupported_device` with no device argument?). But it is also okay as is.

andrewtbarker (approved):

al3x-jp:
Hi, I got the following error when trying to build this checkout: 

[ 51%] Linking CXX executable ../../bin/test_main_blas_ct
terminate called after throwing an instance of 'sycl::_V1::runtime_error'
  what():  No device of requested type 'info::device_type::cpu' available. -1 (PI_ERROR_DEVICE_NOT_FOUND)
CMake Error at /usr/share/cmake-3.22/Modules/GoogleTestAddTests.cmake:83 (message):
  Error running test executable.
 
    Path: '/<path>/oneMKL/build/bin/test_main_blas_ct'
    Result: Subprocess aborted
    Output:
 
Call Stack (most recent call first):
  /usr/share/cmake-3.22/Modules/GoogleTestAddTests.cmake:179 (gtest_discover_tests_impl)
 
gmake[2]: *** [tests/unit_tests/CMakeFiles/test_main_blas_ct.dir/build.make:388: bin/test_main_blas_ct] Error 1

Does the build require a SYCL CPU target being available?
Thanks

s-Nick:
Hi @al3x-jp, thank you for highlighting this issue here. 
We were able to reproduce the same linking error using the `develop` branch and an open source DPC++ build.
In any case, commit [29e94fc](https://github.com/oneapi-src/oneMKL/pull/566/commits/29e94fcb56ce84e7f732bd0f8a63a3762a307611) addresses what we think caused it and on our machine and everything works fine. Could you please test again this PR including this last commit? Your feedback would be very helpful for us.

docs/building_the_project_with_dpcpp.rst
```diff
@@ -408,6 +423,21 @@ set, the backend libraries to enable the use of BLAS, LAPACK and RNG with MKLGPU
...
+.. code-block:: bash
+
+  cmake $ONEMKL_DIR \ 
+      -DCMAKE_CXX_COMPILER=clang++ \
+      -DCMAKE_C_COMPILER=clang \ 
+      -DENABLE_MKLCPU_BACKEND=False \ 
+      -DENABLE_MKLGPU_BACKEND=False \
+      -DENABLE_PORTBLAS_BACKEND=True
+
+Note that this is not a supported configuration. This builds oneMKL Interfaces
```
mkrainiuk: Does it make sense to mention that we do not test it, but it's supported by portBLAS backend? To me "not supported" means we don't have backend for that, but we have, right?
```suggestion
Note that this is not a tested configuration. This builds oneMKL Interfaces
```

tests/unit_tests/main_test.cpp
```diff
@@ -151,14 +153,6 @@ int main(int argc, char** argv) {
 #endif
     }
 
-#if defined(ENABLE_MKLCPU_BACKEND) || defined(ENABLE_NETLIB_BACKEND) || \
```
mkrainiuk, line 154: Could you please comment why we don't need this block anymore?

docs/building_the_project_with_dpcpp.rst
s-Nick: Thank you for your review @mkrainiuk,
I agree with you, we have a backend for it (portBLAS) so tested it's better.

tests/unit_tests/main_test.cpp
s-Nick: This block always adds a CPU device for testing, but CPUs devices are added also in this [block of code](https://github.com/oneapi-src/oneMKL/blob/develop/tests/unit_tests/main_test.cpp#L113-L143). I was going to fix it only for portBLAS but it seems to me that it repeats the logic above so we don't need it.

mkrainiuk: Got it, thank you.

mkrainiuk (approved):

lhuot (approved):
LGTM, thanks!

PR #572, opened 2024-09-19
[lapack][blas][cuda] Update host task impl to use enqueue_native_command

### LINKED ISSUE #216: [CUSOLVER] cuSOLVER handler does not support multiple streams

JackAKirk:
# Description

Update host task impl to use enqueue_native_command for blas/lapack using the cuda backend (cublas/cusolver). I did both backends in a single PR because the cusolver backend uses the cublas backend of oneMKL.

The sycl_ext_codeplay_enqueue_native_command extension reduces latency wrt the host_task for native library submissions, and allows integration with sycl task_graph / events. See https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_codeplay_enqueue_native_command.asciidoc
for details.

This extension has already been shown to lead to considerable performance improvements for applications that call oneMKL, such as Gromacs for the oneMKL fft backend. We expect similar improvements for the lapack and blas backends implemented here.

I had to update the lapack tests because they previously relied on the synchronous behaviour of the native calls due to the fact we had to sync the native streams, since previously with standard host_task we are not able to integrate the native event into the sycl task_graph/ sycl::event.
I did not need to update the blas tests since they already take into account asynchronous behaviour.

# Checklist

https://github.com/oneapi-src/oneMKL/issues/216 is for the most part fixed, but technically this PR maximally enables ooo queue interoperability so we can say that this 
fixes https://github.com/oneapi-src/oneMKL/issues/216

## All Submissions

I've added a test for each backend for each of the possible codepaths:

- SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND is defined: "..native_command"
- SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND is not defined so use previous code path with standard host_task

[test_main_blas_ct_host_task.txt](https://github.com/user-attachments/files/17063141/test_main_blas_ct_host_task.txt)
[test_main_blas_rt_host_task.txt](https://github.com/user-attachments/files/17063142/test_main_blas_rt_host_task.txt)
[test_main_lapack_rt_native_command.txt](https://github.com/user-attachments/files/17063143/test_main_lapack_rt_native_command.txt)
[test_main_lapack_ct_native_command.txt](https://github.com/user-attachments/files/17063144/test_main_lapack_ct_native_command.txt)
[test_main_lapack_ct_host_task.txt](https://github.com/user-attachments/files/17063145/test_main_lapack_ct_host_task.txt)
[test_main_lapack_rt_host_task.txt](https://github.com/user-attachments/files/17063146/test_main_lapack_rt_host_task.txt)
[test_main_blas_ct_res_native_command.txt](https://github.com/user-attachments/files/17063147/test_main_blas_ct_res_native_command.txt)
[test_main_blas_rt_res_native_command.txt](https://github.com/user-attachments/files/17063149/test_main_blas_rt_res_native_command.txt)

### DIFF
--- a/src/blas/backends/cublas/cublas_batch.cpp
+++ b/src/blas/backends/cublas/cublas_batch.cpp
@@ -29,170 +29,152 @@ namespace column_major {
 
 // Buffer APIs
 
-void copy_batch(sycl::queue &queue, int64_t n, sycl::buffer<float, 1> &x,
-                int64_t incx, int64_t stridex, sycl::buffer<float, 1> &y,
-                int64_t incy, int64_t stridey, int64_t batch_size) {
-  throw unimplemented("blas", "copy_batch", "for column_major layout");
-}
-
-void copy_batch(sycl::queue &queue, int64_t n, sycl::buffer<double, 1> &x,
-                int64_t incx, int64_t stridex, sycl::buffer<double, 1> &y,
-                int64_t incy, int64_t stridey, int64_t batch_size) {
-  throw unimplemented("blas", "copy_batch", "for column_major layout");
-}
-
-void copy_batch(sycl::queue &queue, int64_t n,
-                sycl::buffer<std::complex<float>, 1> &x, int64_t incx,
-                int64_t stridex, sycl::buffer<std::complex<float>, 1> &y,
-                int64_t incy, int64_t stridey, int64_t batch_size) {
-  throw unimplemented("blas", "copy_batch", "for column_major layout");
-}
-
-void copy_batch(sycl::queue &queue, int64_t n,
-                sycl::buffer<std::complex<double>, 1> &x, int64_t incx,
-                int64_t stridex, sycl::buffer<std::complex<double>, 1> &y,
-                int64_t incy, int64_t stridey, int64_t batch_size) {
-  throw unimplemented("blas", "copy_batch", "for column_major layout");
-}
-
-void axpy_batch(sycl::queue &queue, int64_t n, float alpha,
-                sycl::buffer<float, 1> &x, int64_t incx, int64_t stridex,
-                sycl::buffer<float, 1> &y, int64_t incy, int64_t stridey,
+void copy_batch(sycl::queue &queue, int64_t n, sycl::buffer<float, 1> &x, int64_t incx,
+                int64_t stridex, sycl::buffer<float, 1> &y, int64_t incy, int64_t stridey,
                 int64_t batch_size) {
-  throw unimplemented("blas", "axpy_batch", "for column_major layout");
+    throw unimplemented("blas", "copy_batch", "for column_major layout");
 }
 
-void axpy_batch(sycl::queue &queue, int64_t n, double alpha,
-                sycl::buffer<double, 1> &x, int64_t incx, int64_t stridex,
-                sycl::buffer<double, 1> &y, int64_t incy, int64_t stridey,
+void copy_batch(sycl::queue &queue, int64_t n, sycl::buffer<double, 1> &x, int64_t incx,
+                int64_t stridex, sycl::buffer<double, 1> &y, int64_t incy, int64_t stridey,
                 int64_t batch_size) {
-  throw unimplemented("blas", "axpy_batch", "for column_major layout");
+    throw unimplemented("blas", "copy_batch", "for column_major layout");
+}
+
+void copy_batch(sycl::queue &queue, int64_t n, sycl::buffer<std::complex<float>, 1> &x,
+                int64_t incx, int64_t stridex, sycl::buffer<std::complex<float>, 1> &y,
+                int64_t incy, int64_t stridey, int64_t batch_size) {
+    throw unimplemented("blas", "copy_batch", "for column_major layout");
+}
+
+void copy_batch(sycl::queue &queue, int64_t n, sycl::buffer<std::complex<double>, 1> &x,
+                int64_t incx, int64_t stridex, sycl::buffer<std::complex<double>, 1> &y,
+                int64_t incy, int64_t stridey, int64_t batch_size) {
+    throw unimplemented("blas", "copy_batch", "for column_major layout");
+}
+
+void axpy_batch(sycl::queue &queue, int64_t n, float alpha, sycl::buffer<float, 1> &x, int64_t incx,
+                int64_t stridex, sycl::buffer<float, 1> &y, int64_t incy, int64_t stridey,
+                int64_t batch_size) {
+    throw unimplemented("blas", "axpy_batch", "for column_major layout");
+}
+
+void axpy_batch(sycl::queue &queue, int64_t n, double alpha, sycl::buffer<double, 1> &x,
+                int64_t incx, int64_t stridex, sycl::buffer<double, 1> &y, int64_t incy,
+                int64_t stridey, int64_t batch_size) {
+    throw unimplemented("blas", "axpy_batch", "for column_major layout");
 }
 
 void axpy_batch(sycl::queue &queue, int64_t n, std::complex<float> alpha,
-                sycl::buffer<std::complex<float>, 1> &x, int64_t incx,
-                int64_t stridex, sycl::buffer<std::complex<float>, 1> &y,
-                int64_t incy, int64_t stridey, int64_t batch_size) {
-  throw unimplemented("blas", "axpy_batch", "for column_major layout");
+                sycl::buffer<std::complex<float>, 1> &x, int64_t incx, int64_t stridex,
+                sycl::buffer<std::complex<float>, 1> &y, int64_t incy, int64_t stridey,
+                int64_t batch_size) {
+    throw unimplemented("blas", "axpy_batch", "for column_major layout");
 }
 
 void axpy_batch(sycl::queue &queue, int64_t n, std::complex<double> alpha,
-                sycl::buffer<std::complex<double>, 1> &x, int64_t incx,
-                int64_t stridex, sycl::buffer<std::complex<double>, 1> &y,
-                int64_t incy, int64_t stridey, int64_t batch_size) {
-  throw unimplemented("blas", "axpy_batch", "for column_major layout");
+                sycl::buffer<std::complex<double>, 1> &x, int64_t incx, int64_t stridex,
+                sycl::buffer<std::complex<double>, 1> &y, int64_t incy, int64_t stridey,
+                int64_t batch_size) {
+    throw unimplemented("blas", "axpy_batch", "for column_major layout");
 }
 
-void gemv_batch(sycl::queue &queue, transpose transa, int64_t m, int64_t n,
-                float alpha, sycl::buffer<float, 1> &a, int64_t lda,
-                int64_t stride_a, sycl::buffer<float, 1> &x, int64_t incx,
-                int64_t stride_x, float beta, sycl::buffer<float, 1> &y,
-                int64_t incy, int64_t stride_y, int64_t batch_size) {
-  throw unimplemented("blas", "gemv_batch", "for column_major layout");
-}
-
-void gemv_batch(sycl::queue &queue, transpose transa, int64_t m, int64_t n,
-                double alpha, sycl::buffer<double, 1> &a, int64_t lda,
-                int64_t stride_a, sycl::buffer<double, 1> &x, int64_t incx,
-                int64_t stride_x, double beta, sycl::buffer<double, 1> &y,
-                int64_t incy, int64_t stride_y, int64_t batch_size) {
-  throw unimplemented("blas", "gemv_batch", "for column_major layout");
-}
-
-void gemv_batch(sycl::queue &queue, transpose transa, int64_t m, int64_t n,
-                std::complex<float> alpha,
-                sycl::buffer<std::complex<float>, 1> &a, int64_t lda,
-                int64_t stride_a, sycl::buffer<std::complex<float>, 1> &x,
-                int64_t incx, int64_t stride_x, std::complex<float> beta,
-                sycl::buffer<std::complex<float>, 1> &y, int64_t incy,
+void gemv_batch(sycl::queue &queue, transpose transa, int64_t m, int64_t n, float alpha,
+                sycl::buffer<float, 1> &a, int64_t lda, int64_t stride_a, sycl::buffer<float, 1> &x,
+                int64_t incx, int64_t stride_x, float beta, sycl::buffer<float, 1> &y, int64_t incy,
                 int64_t stride_y, int64_t batch_size) {
-  throw unimplemented("blas", "gemv_batch", "for column_major layout");
+    throw unimplemented("blas", "gemv_batch", "for column_major layout");
+}
+
+void gemv_batch(sycl::queue &queue, transpose transa, int64_t m, int64_t n, double alpha,
+                sycl::buffer<double, 1> &a, int64_t lda, int64_t stride_a,
+                sycl::buffer<double, 1> &x, int64_t incx, int64_t stride_x, double beta,
+                sycl::buffer<double, 1> &y, int64_t incy, int64_t stride_y, int64_t batch_size) {
+    throw unimplemented("blas", "gemv_batch", "for column_major layout");
 }
 
 void gemv_batch(sycl::queue &queue, transpose transa, int64_t m, int64_t n,
-                std::complex<double> alpha,
-                sycl::buffer<std::complex<double>, 1> &a, int64_t lda,
-                int64_t stride_a, sycl::buffer<std::complex<double>, 1> &x,
-                int64_t incx, int64_t stride_x, std::complex<double> beta,
-                sycl::buffer<std::complex<double>, 1> &y, int64_t incy,
-                int64_t stride_y, int64_t batch_size) {
-  throw unimplemented("blas", "gemv_batch", "for column_major layout");
+                std::complex<float> alpha, sycl::buffer<std::complex<float>, 1> &a, int64_t lda,
+                int64_t stride_a, sycl::buffer<std::complex<float>, 1> &x, int64_t incx,
+                int64_t stride_x, std::complex<float> beta, sycl::buffer<std::complex<float>, 1> &y,
+                int64_t incy, int64_t stride_y, int64_t batch_size) {
+    throw unimplemented("blas", "gemv_batch", "for column_major layout");
+}
+
+void gemv_batch(sycl::queue &queue, transpose transa, int64_t m, int64_t n,
+                std::complex<double> alpha, sycl::buffer<std::complex<double>, 1> &a, int64_t lda,
+                int64_t stride_a, sycl::buffer<std::complex<double>, 1> &x, int64_t incx,
+                int64_t stride_x, std::complex<double> beta,
+                sycl::buffer<std::complex<double>, 1> &y, int64_t incy, int64_t stride_y,
+                int64_t batch_size) {
+    throw unimplemented("blas", "gemv_batch", "for column_major layout");
 }
 
 void dgmm_batch(sycl::queue &queue, side left_right, int64_t m, int64_t n,
-                sycl::buffer<float, 1> &a, int64_t lda, int64_t stride_a,
-                sycl::buffer<float, 1> &x, int64_t incx, int64_t stride_x,
-                sycl::buffer<float, 1> &c, int64_t ldc, int64_t stride_c,
-                int64_t batch_size) {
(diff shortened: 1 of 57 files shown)


src/lapack/backends/cusolver/cusolver_task.hpp
```diff
@@ -50,10 +50,13 @@ namespace cusolver {
...
 template <typename H, typename F>
 static inline void host_task_internal(H &cgh, sycl::queue queue, F f) {
+#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
+    cgh.ext_codeplay_enqueue_native_command([f, queue](sycl::interop_handle ih){
+#else
     cgh.host_task([f, queue](sycl::interop_handle ih) {
+#endif
         auto sc = CusolverScopedContextHandler(queue, ih);
         f(sc);
-        sc.wait_stream(queue);
```
JackAKirk, line 56: This is the only change that also affects the code logic when "SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND" is not defined. I removed it because it is a duplication of the calls to the SYNC macro that happens within lapack function implementations.

src/blas/backends/cublas/cublas_batch.cpp
```diff
@@ -167,12 +167,25 @@ inline void gemm_batch_impl(sycl::queue &queue, transpose transa, transpose tran
...
+            CUBLAS_ERROR_FUNC_T_SYNC("cublasGemmStridedBatchedEx",
+                                     cublasGemmStridedBatchedEx, err, handle,
+                                     get_cublas_operation(transa),
+                                     get_cublas_operation(transb), m, n, k, &alpha, a_,
+                                     get_cublas_datatype<cuTypeA>(), lda, stride_a, b_,
+                                     get_cublas_datatype<cuTypeB>(), ldb, stride_b, &beta,
+                                     c_, get_cublas_datatype<cuTypeC>(), ldc, stride_c,
+                                     batch_size, get_cublas_datatype<cuTypeS>(),
+                                     cublas_gemm_algo);
+#endif
```
Rbiessy: Can't we use `cublas_native_named_func` in this file?

Rbiessy (commented):
LGTM, thanks for the PR

src/blas/backends/cublas/cublas_batch.cpp
JackAKirk: Yeah so for the few cases that I didn't use `cublas_native_named_func` this is because the functions don't pass the function name as a template parameter, so that I cannot use the generic function `cublas_native_named_func`.
I thought about having a neater solution for those cases also, but in the end I decided to just do it like this because there aren't many cases like this (that don't pass the name as a templated function), and also I'm not sure if there is a clean way to do it without making quite a big change to the implementation. 
Although I didn't think too hard about this, there might be a better way.

JackAKirk: Ah maybe I can.. Now that I think about it maybe it will work if I just explicitly pass the template. I'll try.

JackAKirk: It seems that for some reason the compiler cannot infer the type of the cublas functions ending in `*Ex`. I don't know why that is, but seems that I cannot use the `cublas_native_named_func` in those cases. However I spotted that there was one case in this file where I should have used it and didn't, so I've just updated that.

JackAKirk:
@hdelan please review this when you are back.

src/blas/backends/cublas/cublas_batch.cpp
Rbiessy: That's unfortunate, thanks for checking.

Rbiessy (approved):

JackAKirk:
I have a small patch ready to update cublas backend just a little to implement missing `GEMV_BATCH`.
This I think puts cublas backend to a status where everything that maps directly between oneMKL and cublas APIs is supported to some degree (some types remain unimplemented, such as bfloat16/some mixed precisions already identified in the issues board etc).
Is it OK for me to add it here, to save the PR review overhead?
@Rbiessy  what do you think?

Rbiessy:
I would prefer to have a separate PR to make it clearer which commit implements what.

JackAKirk:
> I would prefer to have a separate PR to make it clearer which commit implements what.

Yeah OK, I'll be patient, thanks.

src/blas/backends/cublas/cublas_batch.cpp
```diff
@@ -167,12 +167,25 @@ inline void gemm_batch_impl(sycl::queue &queue, transpose transa, transpose tran
...
+#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
+            CUBLAS_ERROR_FUNC_T("cublasGemmStridedBatchedEx", cublasGemmStridedBatchedEx,
+                                err, handle, get_cublas_operation(transa),
+                                get_cublas_operation(transb), m, n, k, &alpha, a_,
+                                get_cublas_datatype<cuTypeA>(), lda, stride_a, b_,
+                                get_cublas_datatype<cuTypeB>(), ldb, stride_b, &beta, c_,
+                                get_cublas_datatype<cuTypeC>(), ldc, stride_c, batch_size,
+                                get_cublas_datatype<cuTypeS>(), cublas_gemm_algo);
+#else
+            CUBLAS_ERROR_FUNC_T_SYNC("cublasGemmStridedBatchedEx",
```
hdelan: Why is this formatted differently from the thing it's replacing?

JackAKirk: I used clang-format on it and previously it either didn't use clang-format or had a different setting etc. I can change it back.

hdelan: There is a local clang format config [here](https://github.com/oneapi-src/oneMKL/blob/develop/_clang-format). You should make sure you are using this config.

hdelan: I am delighted to see that we now have CI on oneMKL tests. Would another clang format job in CI be a good idea @Rbiessy ?

src/lapack/backends/cusolver/cusolver_task.hpp
hdelan: Where does this `SYNC` macro get called? Does it know not to do so when using `enqueue_native_command`?

JackAKirk:
Hi @oneapi-src/onemkl-blas-write @oneapi-src/onemkl-lapack-write
would it be possible for you to review this? 

Thanks

hdelan (approved):
LGTM

src/lapack/backends/cusolver/cusolver_task.hpp
JackAKirk: mainly here: https://github.com/oneapi-src/oneMKL/pull/572/files#diff-b374cd674b06d7901c4ad5607b38b48de30a5f9e7215d58130ff80c7db51e680

and here: https://github.com/oneapi-src/oneMKL/pull/572/files#diff-bca3dc194e431af2adf1dcb0fdcd5bf8ba243163fdcac5bc304bf340e933d9d8

src/blas/backends/cublas/cublas_batch.cpp
Rbiessy: We've not done anything on the GitHub CI but it is nice indeed. A clang format job is a good idea. I think we would need to discuss which clang-format version to use. We have seen differences between different versions, the internal CI uses clang-format-9 AFAIK. I think it would be easier to use the one shipped with DPC++ instead. I'll make a note to create an issue about that.

src/blas/backends/cublas/cublas_batch.cpp
```diff
@@ -167,12 +167,25 @@ inline void gemm_batch_impl(sycl::queue &queue, transpose transa, transpose tran
             auto b_ = sc.get_mem<cuTypeB *>(b_acc);
             auto c_ = sc.get_mem<cuTypeC *>(c_acc);
             cublasStatus_t err;
-            CUBLAS_ERROR_FUNC_T_SYNC(
-                "cublasGemmStridedBatchedEx", cublasGemmStridedBatchedEx, err, handle,
-                get_cublas_operation(transa), get_cublas_operation(transb), m, n, k, &alpha, a_,
-                get_cublas_datatype<cuTypeA>(), lda, stride_a, b_, get_cublas_datatype<cuTypeB>(),
-                ldb, stride_b, &beta, c_, get_cublas_datatype<cuTypeC>(), ldc, stride_c, batch_size,
-                get_cublas_datatype<cuTypeS>(), cublas_gemm_algo);
+#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
```
andrewtbarker, line 170: Can you use `cublas_native_named_func("cublasGemmStridedBatchedEx", ...)` to avoid the `#ifdef`? Here and in a few similar places below.

src/blas/backends/cublas/cublas_level3.cpp
```diff
@@ -94,10 +94,17 @@ inline void gemm_ex(DATATYPE_A DT_A, DATATYPE_B DT_B, DATATYPE_C DT_C, sycl::que
             auto b_ = sc.get_mem<cuDataType_B *>(b_acc);
             auto c_ = sc.get_mem<cuDataType_C *>(c_acc);
             cublasStatus_t err;
+#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
             CUBLAS_ERROR_FUNC_SYNC(cublasGemmEx, err, handle, get_cublas_operation(transa),
```
andrewtbarker, line 98: Why not `cublas_native_func` here?

src/blas/backends/cublas/cublas_batch.cpp
JackAKirk: Hi Andrew, thanks for the review, please see https://github.com/oneapi-src/oneMKL/pull/572/files#r1768732674 for the reasoning here.

andrewtbarker: That link doesn't show me anything relevant, but if you have a reason I trust you, I just thought it might be an oversight.

JackAKirk: This is what I meant mainly 

"Yeah so for the few cases that I didn't use cublas_native_named_func this is because the functions don't pass the function name as a template parameter, so that I cannot use the generic function cublas_native_named_func."

"It seems that for some reason the compiler cannot infer the type of the cublas functions ending in *Ex. I don't know why that is, but seems that I cannot use the cublas_native_named_func in those cases."

andrewtbarker (approved):
The BLAS part looks good, thank you. I'm less familiar with the LAPACK part.

src/blas/backends/cublas/cublas_batch.cpp
andrewtbarker: Got it, thanks! Sorry I missed that comment.

tests/unit_tests/lapack/source/gebrd.cpp
```diff
@@ -76,6 +76,7 @@ bool accuracy(const sycl::device& dev, int64_t m, int64_t n, int64_t lda, uint64
         TEST_RUN_LAPACK_CT_SELECT(
             queue, scratchpad_size = oneapi::mkl::lapack::gebrd_scratchpad_size<fp>, m, n, lda);
 #endif
+        queue.wait_and_throw();
```
sknepper: Since the scratchpad_size functions return the scratchpad_size value (rather than an event), that seems to make them synchronous, and so is it required to wait on the queue?

tests/unit_tests/lapack/source/gebrd.cpp
```diff
@@ -149,6 +150,7 @@ bool usm_dependency(const sycl::device& dev, int64_t m, int64_t n, int64_t lda,
...
```
sknepper: Same question about scratchpad_size

tests/unit_tests/lapack/source/gebrd.cpp
```diff
@@ -166,6 +168,7 @@ bool usm_dependency(const sycl::device& dev, int64_t m, int64_t n, int64_t lda,
                                   d_dev, e_dev, tauq_dev, taup_dev, scratchpad_dev, scratchpad_size,
                                   std::vector<sycl::event>{ in_event });
 #endif
+        queue.wait_and_throw();
```
sknepper: And this part of the test is checking dependency handling. If we wait on the queue before we call check_dependency, does this test even make sense?
(In the accuracy testing part of the test, we do explicitly wait on the queue.)

sknepper (commented):
Could you please provide more details about why the LAPACK tests required updating? I see your comment in the description of the PR, but I still have a few questions. Thanks!
Also, if the LAPACK tests do need updating, how will that impact testing the other (non-cuSOLVER) backends?

tests/unit_tests/lapack/source/gebrd.cpp
JackAKirk: Thanks, you are right that I should (and I will) remove these waits. However be aware I also have had to remove a part of the dependency check here https://github.com/oneapi-src/oneMKL/pull/572/commits/af4d1fd3335447c2d486d31d1da0ab67aa616260
so that the test will pass for the cuda backend once these waits are removed.

It is a slightly subtle point:

The enqueue native command is free (and encouraged) to submit a command to the same native queue (cuda stream) that was used for a dependent event on the same device. According to the sycl specification once commands are submitted to a native device they should be marked `event_command_status::running`.

Then (from https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_event_information_and_profiling_descriptors) the only relevant part of the spec wrt whether dep events can be "running" at the same time as events that depend on them is:

"Implementations are encouraged to return a timestamp that is as close as possible to the point when the action starts running on the device, but there is no specific accuracy that is guaranteed."

enqueue native event is such a case where the timestamp returned indicating that the event action has started running on a device might not be accurate, in the case that it has a dependent event that has also already been submitted on the native queue for execution and also not yet begun. There is a priori no way of knowing exactly when the device begins work on the natively submitted operation, even in the case that there are no dependent events.

I don't see any other relevant mention of requirements for the check I removed https://github.com/oneapi-src/oneMKL/pull/572/commits/af4d1fd3335447c2d486d31d1da0ab67aa616260 in the sycl spec: In short I don't think it is a requirement of the sycl spec, and it is not desirable for it to be a requirement since it would lead to a performance drop if we were to enforce such a constraint in the runtime, at no cost of correctness.

Therefore I have gone ahead and removed this check from the `check_dependency` function. However I keep in the check that a dependent event cannot `complete` after a event that is dependent upon it has been marked `complete`.

I understand that this may catch less truly problematic behaviour in other backends that the stronger check. However since having the two events simultaneously marked "running" is technically allowed (and desired in some cases that I mention above), I don't see any choice but to remove the "stronger" but technically wrong check.
Is that OK?

tests/unit_tests/lapack/source/gebrd.cpp
JackAKirk: Yeah these queue.wait_and_throw() should be unnecessary, thanks for pointing it out. This made me realize that actually I can still improve the implementation for scratchpad functions, since in the cuda backend they are asynchronous. I will update them soon.

tests/unit_tests/lapack/source/gebrd.cpp
sknepper: Thanks for the explanation and changes, @JackAKirk ! I think this is a good change. Perhaps in the future, if we start doing accuracy testing on the device (without copying back to host with the subsequent waits and verifying results there), then the dependency checking could be removed altogether as likely the accuracy tests would then also pick up any dependency issues.

hdelan: Just to clarify

> There is a priori no way of knowing exactly when the device begins work on the natively submitted operation, even in the case that there are no dependent events.

Enqueue native command will record a native event before the beginning of the interop work (on the chosen ‘native command queue’) if profiling is enabled on the SYCL queue. In this way we have the same event recording patterns as normal SYCL commands.

JackAKirk: > Just to clarify
> 
> > There is a priori no way of knowing exactly when the device begins work on the natively submitted operation, even in the case that there are no dependent events.
> 
> Enqueue native command will record a native event before the beginning of the interop work (on the chosen ‘native command queue’) if profiling is enabled on the SYCL queue.

Yeah I should have clarified that this is really putting interop work via enqueue_native_command into the situation where event recording patterns match more closely standard SYCL commands. It is rather a characteristic of the CUDA backend that you can have subsequent events from a native stream return "running" simultaneously, since "running" actually means they have been submitted to the native cuda scheduler, rather than that the native cuda scheduler has actually started work from a command before a command that it depends on has finished.

This may actually be the situation in L0 also, but I'm not sure.

hdelan: Ah yes OK this is through the `running` mechanism that you mention. Gotcha

tests/unit_tests/lapack/source/gebrd.cpp
JackAKirk: I've now reverted all changes to these tests. I've also improved the synchronous implementation of scratchpad_size functions in this commit: https://github.com/oneapi-src/oneMKL/pull/572/commits/c082d4a0a90bfa3ecd4a144bd9cb4b0f5adb54cc

I swapped the queue.wait() with a cuStreamSync call that waits only on the stream that is used, in addition to waiting on the event returned from the host_task (in the case that it returns before cuStreamSync has been returned). This means that we don't unnecessarily sync all streams in the queue anymore.

src/blas/backends/cublas/cublas_batch.cpp
JackAKirk: Should I just leave this as it is then or should I clang-format all files I touched with one or other of the format configs?

I didn't worry about it before because it seemed to me to have a mix of clangformat configs currently.

tests/unit_tests/lapack/source/gebrd.cpp
JackAKirk: I've now removed the changes I made to the tests and this PR should be ready for merge subject to whether any further clang-format changes are requested.

sknepper (approved):
Thanks for all your work here to improve performance, @JackAKirk !

src/blas/backends/cublas/cublas_batch.cpp
Rbiessy: For reference, created this issue https://github.com/oneapi-src/oneMKL/issues/580

Rbiessy: I don't know if there is an issue with clang-format configs.
I would suggest to download clang-format-9 and run `git-clang-format --binary /path/to/clang-format-9_linux-amd64 HEAD~8` to format all your changes.
The clang-format config should be automatically used if you are running this from the root of the repository.

src/blas/backends/cublas/cublas_batch.cpp
JackAKirk: Fixed the format now. Thanks

src/blas/backends/cublas/cublas_batch.cpp
JackAKirk: Fixed the format now, thanks.

tests/unit_tests/lapack/source/gebrd.cpp
JackAKirk: For future reference: 
Answered here https://github.com/oneapi-src/oneMKL/pull/572#discussion_r1784569376

tests/unit_tests/lapack/source/gebrd.cpp
JackAKirk: clang-format fixed.

src/blas/backends/cublas/cublas_level3.cpp
JackAKirk: For future reference, this was answered here: https://github.com/oneapi-src/oneMKL/pull/572#discussion_r1778762606

Issue #554, opened 2024-08-09
Add prefixes to macros in in config.hpp.in


al42and:
# Summary

To prevent potential macro naming conflicts between oneMKL and client applications, this RFC proposes adding a unique prefix (such as `ONEMKL_` or `ONEAPI_ONEMKL_`) to all macros defined in oneMKL's public API.

# Problem statement

oneMKL defines a bunch of configuration macros in `config.hpp`, which is transitively included in the `<oneapi/mkl.hpp>` file. 

Those are `ENABLE_<backend>_BACKEND` for each enabled backend, `BUILD_SHARED_LIBS` and `REF_(BLAS|CBLAS)_LIBNAME`.

As a consequence, the client application that tries to use oneMKL will also have them all defined. This could be useful, but, since macros don't obey any scope rules, they can also easily cause confusion if the client application itself uses, e.g., `ENABLE_CUFFT_BACKEND` or `BUILD_SHARED_LIBS`.

[ES.33 in C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-MACROS) recommends adding _supposedly unique prefixes (e.g., your organization’s name)_ to the macro names.

As such, I suggest adding `ONEMKL_` or `ONEAPI_ONEMKL_` prefix to all the macros defined in the public API.

# Details

* Those macros are not part of the specification as far as I can tell, but they can (intentionally or not) affect user applications.

Rbiessy:
Thanks for the RFC, I agree this would be nicer. I would suggest using `ONEAPI_ONEMKL_` for now. Is this blocking you? We may have the bandwidth to work on this but contributions are welcome if anyone needs this soon.

al42and:
> Is this blocking you?

Not blocking, just wanted to raise a concern since it's part of public API (I don't have a link at hand, but it was suggested in one of the examples to check these variables).

> We may have the bandwidth to work on this but contributions are welcome if anyone needs this soon.

Ok! :wink:

s-Nick:
#571 addressed this issue

PR #576, opened 2024-09-30
[BLAS][portBLAS] Add bindings for half and some gemm_batch group APIs

Rbiessy:
# Description

Add missing features needed to run llama.cpp with oneMKL+portBLAS:

- Add an option to enable `sycl::half` bindings
- Add some missing bindings for `batch_gemm`
- Avoid some warnings in the tests

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Log using the CT tests and examples: [test_log_portblas_example.txt](https://github.com/user-attachments/files/17192918/test_log_portblas_example.txt) [test_log_portblas_ct.txt](https://github.com/user-attachments/files/17192919/test_log_portblas_ct.txt). The tests are currently very slow so not all the RT tests were run.


Rbiessy:
Hello @al3x-jp, I have been told you are keen to test oneMKL+portBLAS with llama.cpp. This branch should enable everything that is needed. Would you be able to try it and let us know if you run into any issue?
You should just need to add `-DENABLE_PORTBLAS_BACKEND=ON -DPORTBLAS_ENABLE_HALF=ON` when compiling oneMKL Interface.

src/CMakeLists.txt
```diff
@@ -45,6 +45,11 @@ foreach(domain ${TARGET_DOMAINS})
   add_subdirectory(${domain})
 endforeach()
 
+if (PORTBLAS_ENABLE_HALF)
+  # Set the variable used for C++ macro
+  set(ENABLE_PORTBLAS_HALF ON)
```
s-Nick, line 50: According to #554 and PR #571 this var should be named with the prefix `ONEAPI_ONEMKL_`. Could you update it?
I know it would be the only one named correctly now, but I hope that PR will be merged soon.

src/blas/backends/portblas/portblas_batch.cxx
```diff
@@ -695,7 +710,7 @@ sycl::event gemm_batch(sycl::queue &queue, oneapi::mkl::transpose *transa,
                        const double **b, std::int64_t *ldb, double *beta, double **c,
                        std::int64_t *ldc, std::int64_t group_count, std::int64_t *group_size,
                        const std::vector<sycl::event> &dependencies) {
-    throw unimplemented("blas", "gemm_batch", " for USM");
+    throw unimplemented("blas", "gemm_batch", " for USM using double");
```
s-Nick: Could you explain me why it doesn't work with `double`? I thought that in portBLAS if it works with `float` it works with `double`. Did you test it?

src/blas/backends/portblas/portblas_level3_half.cpp
```diff
@@ -53,31 +68,56 @@ sycl::event gemm(sycl::queue &queue, oneapi::mkl::transpose transa, oneapi::mkl:
...
+}
+
 // BUFFER
 void gemm(sycl::queue &queue, oneapi::mkl::transpose transa, oneapi::mkl::transpose transb,
           std::int64_t m, std::int64_t n, std::int64_t k, sycl::half alpha,
           sycl::buffer<sycl::half, 1> &a, std::int64_t lda, sycl::buffer<sycl::half, 1> &b,
           std::int64_t ldb, sycl::half beta, sycl::buffer<sycl::half, 1> &c, std::int64_t ldc) {
+#ifdef ENABLE_PORTBLAS_HALF
+    CALL_PORTBLAS_FN(::blas::_gemm, queue, transa, transb, m, n, k, alpha, a, lda, b, ldb, beta, c,
+                     ldc);
```
s-Nick, line 105: Currently, we don't support any row major operator in portBLAS, so they are generally not enabled. 
Even if it works now, unless it is necessary, I would not enable it to avoid the confusion of having only one row major operator supported.

src/CMakeLists.txt
Rbiessy: Yes I was planning to update this PR once https://github.com/oneapi-src/oneMKL/pull/571 is merged. It will be easier to do once I can add `ONEAPI_ONEMKL_ENABLE_PORTBLAS_HALF` to the list you introduce in https://github.com/oneapi-src/oneMKL/pull/571/files#diff-148715d6ea0c0ea0a346af3f6bd610d010d490eca35ac6a9b408748f7ca9e3f4R54

src/blas/backends/portblas/portblas_level3_half.cpp
Rbiessy: There are already checks that throw unimplemented when `row_major` is used here: https://github.com/oneapi-src/oneMKL/blob/develop/src/blas/backends/portblas/portblas_common.hpp#L193
All the other functions have already their bindings ready for `row_major` because they use this pattern: https://github.com/oneapi-src/oneMKL/blob/develop/src/blas/backends/portblas/portblas_level3_float.cpp#L54
Just `portblas_level3_half.cpp` looks a bit different

src/blas/backends/portblas/portblas_batch.cxx
Rbiessy: It can probably be added, I'm looking this. I have found that the gemm_batch tests only use `group_count=5`. I need to test this with `group_count=1`.

Rbiessy:
Closing as I don't think there is enough interest right now.

PR #577, opened 2024-10-01
[Unit tests] skip BLAS and LAPACK unit tests if Netlib is not found

mkrainiuk:
# Description

This PR implements one more possible solution for the problem described in https://github.com/oneapi-src/oneMKL/pull/557 
In case Netlib libraries are not available it shouldn't block completely oneMKL Interfaces building, instead we can generate warning that BLAS and LAPACK unit tests will be skipped because of Netlib absence. 

This PR also includes some pre-work for adding BLAS and LAPACK tests that won't require Netlib.

* Example of the output users will get if Netlib is found
```
-- Found LAPACKE: /usr/lib64/liblapacke64.so
-- Found LAPACKE: /usr/lib64/liblapack64.so
-- Found LAPACKE: /usr/lib64/libcblas64.so
-- Found LAPACKE: /usr/lib64/libblas64.so
-- Found LAPACKE: /usr/include
...
-- Configuring done
-- Generating done
```

* Example of the output users will get if Netlib is NOT found
```
-- Could NOT find LAPACKE (missing: LAPACKE64_file)
-- Could NOT find LAPACKE (missing: LAPACK64_file)
-- Could NOT find LAPACKE (missing: CBLAS64_file)
-- Could NOT find LAPACKE (missing: BLAS64_file)
-- Could NOT find LAPACKE (missing: LAPACKE_INCLUDE LAPACKE_LINK)
CMake Warning at tests/unit_tests/CMakeLists.txt:33 (message):
  Netlib LAPACKE headers or libraries are not found, LAPACK unit tests will
  be skipped
...
-- Configuring done
-- Generating done
```

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Attach a log. Both public and internal CIs were passed
- [x] Have you formatted the code using clang-format? N/A, all changes are CMake related only

### DIFF
--- a/tests/unit_tests/CMakeLists.txt
+++ b/tests/unit_tests/CMakeLists.txt
@@ -17,15 +17,27 @@
 # SPDX-License-Identifier: Apache-2.0
 #===============================================================================
 
-if("blas" IN_LIST TARGET_DOMAINS)
-  find_package(CBLAS REQUIRED)
+set(TEST_TARGET_DOMAINS ${TARGET_DOMAINS})
+
+if("blas" IN_LIST TEST_TARGET_DOMAINS)
+  find_package(CBLAS)
+  if(NOT CBLAS_FOUND)
+    # TODO: add list of tests without Netlib dependency
+    message(WARNING "Netlib CBLAS headers or libraries are not found, BLAS unit tests will be skipped")
+    list(REMOVE_ITEM TEST_TARGET_DOMAINS "blas")
+  endif()
 endif()
 
-if("lapack" IN_LIST TARGET_DOMAINS)
-    find_package(LAPACKE REQUIRED)
+if("lapack" IN_LIST TEST_TARGET_DOMAINS)
+  find_package(LAPACKE)
+  if(NOT LAPACKE_FOUND)
+    # TODO: add list of tests without Netlib dependency
+    message(WARNING "Netlib LAPACKE headers or libraries are not found, LAPACK unit tests will be skipped")
+    list(REMOVE_ITEM TEST_TARGET_DOMAINS "lapack")
+  endif()
 endif()
 
-foreach(domain ${TARGET_DOMAINS})
+foreach(domain ${TEST_TARGET_DOMAINS})
   # Build tests first
   add_subdirectory(${domain})
 endforeach()
@@ -50,7 +62,9 @@ set(blas_TEST_LINK "")
 set(lapack_TEST_LIST
         lapack_source)
 
-set(lapack_TEST_LINK ${LAPACKE_LINK})
+if(LAPACKE_FOUND)
+  set(lapack_TEST_LINK ${LAPACKE_LINK})
+endif()
 
 # RNG config
 set(rng_TEST_LIST
@@ -75,7 +89,7 @@ set(sparse_blas_TEST_LIST
 
 set(sparse_blas_TEST_LINK "")
 
-foreach(domain ${TARGET_DOMAINS})
+foreach(domain ${TEST_TARGET_DOMAINS})
   # Generate RT and CT test lists
   set(${domain}_TEST_LIST_RT ${${domain}_TEST_LIST})
   set(${domain}_TEST_LIST_CT ${${domain}_TEST_LIST})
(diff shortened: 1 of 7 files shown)


tests/unit_tests/CMakeLists.txt
```diff
@@ -18,11 +18,21 @@
 #===============================================================================
 
 if("blas" IN_LIST TARGET_DOMAINS)
-  find_package(CBLAS REQUIRED)
+  find_package(CBLAS)
+  if(NOT CBLAS_FOUND)
+    # TODO: add list of tests without Netlib dependency
+    message(WARNING "Netlib CBLAS headers or libraries are not found, BLAS unit tests will be skipped")
+    list(REMOVE_ITEM TARGET_DOMAINS "blas")
```
Rbiessy: I would suggest to introduce a `TESTED_TARGET_DOMAINS` variable to avoid modifying `TARGET_DOMAINS` which is a global variable used elsewhere. This works fine as long as the tests are included last in the top-level CMakeLists.txt but I think it is best to avoid.

tests/unit_tests/blas/batch/CMakeLists.txt
```diff
@@ -18,38 +18,40 @@
...
+set(BATCH_SOURCES_W_CBLAS "copy_batch_stride.cpp" "axpy_batch_stride.cpp" "dgmm_batch_stride.cpp" "gemm_batch_stride.cpp" "gemv_batch_stride.cpp" "trsm_batch_stride.cpp" "syrk_batch_stride.cpp" "copy_batch_usm.cpp" "copy_batch_stride_usm.cpp" "axpy_batch_usm.cpp" "axpy_batch_stride_usm.cpp" "dgmm_batch_usm.cpp" "dgmm_batch_stride_usm.cpp" "gemm_batch_usm.cpp" "gemm_batch_stride_usm.cpp" "gemv_batch_usm.cpp" "gemv_batch_stride_usm.cpp" "trsm_batch_usm.cpp" "trsm_batch_stride_usm.cpp" "syrk_batch_usm.cpp" "syrk_batch_stride_usm.cpp" "omatcopy_batch_stride.cpp" "omatcopy_batch_stride_usm.cpp" "imatcopy_batch_stride.cpp" "imatcopy_batch_stride_usm.cpp" "omatadd_batch_stride.cpp" "omatadd_batch_stride_usm.cpp" "omatcopy_batch_usm.cpp" "imatcopy_batch_usm.cpp")
+# TODO: add list of tests without Netlib dependency
+set(BATCH_SOURCES)
 
 if(BUILD_SHARED_LIBS)
-  add_library(blas_batch_rt OBJECT ${BATCH_SOURCES})
+  add_library(blas_batch_rt OBJECT $<$<BOOL:${CBLAS_FOUND}>:${BATCH_SOURCES_W_CBLAS}> ${BATCH_SOURCES})
```
Rbiessy: You could directly set the variable `BATCH_SOURCES` to `$<$<BOOL:${CBLAS_FOUND}>:${BATCH_SOURCES_W_CBLAS}>`  or create another variable that list all the sources.
The pattern `$<$<BOOL:${CBLAS_FOUND}>:${BATCH_SOURCES_W_CBLAS}> ${BATCH_SOURCES}` is used 4 times so it is worth simplifying.

Rbiessy (approved):

Rbiessy:
LGTM!

tests/unit_tests/blas/batch/CMakeLists.txt
```diff
@@ -29,7 +36,7 @@ if(BUILD_SHARED_LIBS)
       PUBLIC ${PROJECT_SOURCE_DIR}/include
       PUBLIC ${PROJECT_SOURCE_DIR}/deps/googletest/include
       PUBLIC ${CMAKE_BINARY_DIR}/bin
-      PUBLIC ${CBLAS_INCLUDE}
+      $<$<BOOL:${CBLAS_FOUND}>:${CBLAS_INCLUDE}>
```
andrewtbarker, line 39: I don't really follow this syntax, but are we changing this argument from `PUBLIC` to non-public?

andrewtbarker (approved):
Look good, thanks for the changes!

tests/unit_tests/blas/batch/CMakeLists.txt
mkrainiuk: New logic helps us to drop `CBLAS_INCLUDE` from the compilation line when CBLAS header file is not found. Otherwise it might appear in the compilation line as `-ICBLAS_INCLUDE_not_found` and cause test compilation issue. 

As for `PUBLIC` keyword since each BLAS test is a final binary, not library, we don't need to mark this dependency as needed to be propagated to other targets that depend on this target, because we don't have such targets.

PR #578, opened 2024-10-02
[rocfft][cufft] DFT update host task to use native command

JackAKirk:
# Description

Similar to https://github.com/oneapi-src/oneMKL/pull/572 (see the discussions in that PR for technical details) except this covers fft backends for both amd and nvidia cases.

Update host task impl to use enqueue_native_command for DFT using the cuda/hip backends.

tests:

[test_main_dft_ct_amd.txt](https://github.com/user-attachments/files/17233320/test_main_dft_ct_amd.txt)
[test_main_dft_rt_amd.txt](https://github.com/user-attachments/files/17233321/test_main_dft_rt_amd.txt)
[test_main_dft_rt_nvidia.txt](https://github.com/user-attachments/files/17233322/test_main_dft_rt_nvidia.txt)
[test_main_dft_ct_nvidia.txt](https://github.com/user-attachments/files/17233323/test_main_dft_ct_nvidia.txt)

author: @hjabird


JackAKirk:
@oneapi-src/onemkl-dft-write could you please review this?

Thanks

src/dft/backends/rocfft/execute_helper.hpp
```diff
@@ -76,20 +76,46 @@ inline hipStream_t setup_stream(const std::string &func, sycl::interop_handle &i
...
+       throw oneapi::mkl::exception("dft/backends/rocfft", func,
+                                    "hipStreamSynchronize returned " + std::to_string(result));
+   }
 }
 
-inline void execute_checked(const std::string &func, const rocfft_plan plan, void *in_buffer[],
+inline void execute_checked(const std::string &func, hipStream_t stream, const rocfft_plan plan, void *in_buffer[],
                             void *out_buffer[], rocfft_execution_info info) {
-    auto result = rocfft_execute(plan, in_buffer, out_buffer, info);
+    auto result = rocfft_execute(plan, in_buffer, out_buffer, info);
```
Rbiessy:
```suggestion
    auto result = rocfft_execute(plan, in_buffer, out_buffer, info);
```

Rbiessy (approved):
Thanks for the PR! Note we have used this in 2 GROMACS benchmarks to get between 2% to 7% improvements on MI210.

JackAKirk:
@anantsrivastava30 Is this OK to merge?

Thanks

src/dft/backends/cufft/execute_helper.hpp
```diff
@@ -143,6 +147,27 @@ inline CUstream setup_stream(const std::string &func, sycl::interop_handle ih, c
...
+ * 
+ * @tparam HandlerT The command group handler type
+ * @tparam FnT The body of the enqueued task
+ *
+ * Either uses host task interop API, or enqueue native command extension.
+ * This extension avoids host synchronization after 
+ * the CUDA call is complete.
+ */
+template <typename HandlerT, typename FnT>
+static inline void cufft_enqueue_task(HandlerT&& cgh, FnT&& f) {
```
lhuot, line 161: Why is it needed to duplicate this in both cuFFT and rocFFT backends and in various domains (BLAS, LAPACK, FFT)? Can't we have one wrapper used across all domains and backends?

src/dft/backends/cufft/execute_helper.hpp
```diff
@@ -143,6 +147,27 @@ inline CUstream setup_stream(const std::string &func, sycl::interop_handle ih, c
...
+#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
+    cgh.ext_codeplay_enqueue_native_command([=](sycl::interop_handle ih){
+#else
+    cgh.host_task([=](sycl::interop_handle ih){
+#endif
+        f(std::move(ih));
```
lhuot: Why is it needed to duplicate this in both cuFFT and rocFFT backends and in various domains (BLAS, LAPACK, FFT)? Can't we have one wrapper used across all domains and backends?

JackAKirk:
Hi @lhuot 

I don't know, I am following the existing design. I think that it is a good question and it would be good to open this as an issue. However I don't think it is a good idea to block this PR for this, since this is a critical patch for Gromacs performance, which does not introduce this duplication that already exists in all other backends.

src/dft/backends/cufft/execute_helper.hpp
Rbiessy: My understanding is that the domains have always been separated on purpose. This makes review much easier as any change affecting common code would technically require an approval from every domain owners.
I agree this could be discussed in an issue. To my knowledge there is very little code that could be common across domains, other than the types and exceptions which are already common.

lhuot:
> Hi @lhuot
> 
> I don't know, I am following the existing design. I think that it is a good question and it would be good to open this as an issue. However I don't think it is a good idea to block this PR for this, since this is a critical patch for Gromacs performance, which does not introduce this duplication that already exists in all other backends.

Then, let's fix the code duplication in the DFT domain please.

JackAKirk:
> > Hi @lhuot
> > I don't know, I am following the existing design. I think that it is a good question and it would be good to open this as an issue. However I don't think it is a good idea to block this PR for this, since this is a critical patch for Gromacs performance, which does not introduce this duplication that already exists in all other backends.
> 
> Then, let's fix the code duplication in the DFT domain please.

OK sure, do you want me to add a new header e.g. "execute_helper_generic.hpp" in oneMKL/src/dft/ with the fft_host_task implementation that is portable across hip and cuda backends, and then only include this header in the rocfft and cufft backends?
I can't include such a generic header in include/oneapi/mkl/detail/ because the `enqueue_native_command` extension is not available in intel backends and trying to link this header will break intel builds.

lhuot:
> > > Hi @lhuot
> > > I don't know, I am following the existing design. I think that it is a good question and it would be good to open this as an issue. However I don't think it is a good idea to block this PR for this, since this is a critical patch for Gromacs performance, which does not introduce this duplication that already exists in all other backends.
> > 
> > 
> > Then, let's fix the code duplication in the DFT domain please.
> 
> OK sure, do you want me to add a new header e.g. "execute_helper_generic.hpp" in oneMKL/src/dft/ with the fft_host_task implementation that is portable across hip and cuda backends, and then only include this header in the rocfft and cufft backends? I can't include such a generic header in include/oneapi/mkl/detail/ because the `enqueue_native_command` extension is not available in intel backends and trying to link this header will break intel builds.

Sounds like a reasonable approach to me. Thanks!

src/dft/execute_helper_generic.hpp
```diff
@@ -0,0 +1,53 @@
...
+* software distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions
+* and limitations under the License.
+*
+*
+* SPDX-License-Identifier: Apache-2.0
+*******************************************************************************/
+
+#ifndef _ONEMKL_DFT_SRC_CUFFT_EXECUTE_GENERIC_HPP_
```
Rbiessy: `_ONEMKL_DFT_SRC_CUFFT_EXECUTE_GENERIC_HPP_` -> `_ONEMKL_DFT_SRC_EXECUTE_HELPER_GENERIC_HPP_`

The `endif` should also have a comment with the same name (looks like it was missed in a few headers).

Rbiessy: Sorry for the nitpicking, it should really end with `EXECUTE_HELPER_GENERIC_HPP_` not just `EXECUTE_GENERIC_HPP_`!

Rbiessy (approved):

JackAKirk:
> > > > Hi @lhuot
> > > > I don't know, I am following the existing design. I think that it is a good question and it would be good to open this as an issue. However I don't think it is a good idea to block this PR for this, since this is a critical patch for Gromacs performance, which does not introduce this duplication that already exists in all other backends.
> > > 
> > > 
> > > Then, let's fix the code duplication in the DFT domain please.
> > 
> > 
> > OK sure, do you want me to add a new header e.g. "execute_helper_generic.hpp" in oneMKL/src/dft/ with the fft_host_task implementation that is portable across hip and cuda backends, and then only include this header in the rocfft and cufft backends? I can't include such a generic header in include/oneapi/mkl/detail/ because the `enqueue_native_command` extension is not available in intel backends and trying to link this header will break intel builds.
> 
> Sounds like a reasonable approach to me. Thanks!

I've updated this now as requested. Tests still pass as posted in PR summary on both hip and cuda backends.

lhuot (approved):
LGTM, thanks!

PR #584, opened 2024-10-07
Remove reference libraries define from config header

npmiller:
# Description

The `config.hpp` oneMKL header contains both the macros for which domains are enabled, but it also contains the absolute path to the reference libraries when the tests are built.

This is a problem because then when using the `install` target, this header gets put into the install directory and still contains absolute path to testing infrastructure.

This patch splits out the reference libraries path into a separate config header only used by the tests and that won't get installed, that way a oneMKL installation will have a clean config header.

# Checklist

## All Submissions

  - I don't have easy access to all the hardware to check every single configuration but the change is the same across the board so it should be fine, here are the logs on CUDA:
  - [blas.log](https://github.com/user-attachments/files/17276389/blas.log)
  - [lapack.log](https://github.com/user-attachments/files/17276397/lapack.log)


tests/unit_tests/CMakeLists.txt
```diff
@@ -25,6 +25,11 @@ if("lapack" IN_LIST TARGET_DOMAINS)
     find_package(LAPACKE REQUIRED)
 endif()
 
+# Setup config header with path to reference libraries
+set(ONEMKL_TESTS_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/include/config.hpp.in
+               ${ONEMKL_TESTS_BINARY_DIR}/include/config.hpp)
```
Rbiessy: I'm not keen to have the same name `config.hpp`. I would suggest to either rename it to `test_config.hpp` for instance or even better IMO: do not use a config file and instead directly define `REF_BLAS_LIBNAME` and `REF_CBLAS_LIBNAME` as macros when compiling the tests. This could be done in https://github.com/oneapi-src/oneMKL/blob/develop/tests/unit_tests/CMakeLists.txt when the domain is "blas" or "lapack" using `target_compile_definitions`.
I've just never been a fan of generated headers.

npmiller: That's a good point, I just pushed a much simpler solution using `add_definitions`, also it turns out only the blas domain needs this so it's also much cleaner that way.

Rbiessy (approved):
Thanks, LGTM.
@s-Nick FYI this will have an impact on https://github.com/oneapi-src/oneMKL/pull/571 once this is merged.

mkrainiuk (approved):
Looks good to me, thanks.

Issue #580, opened 2024-10-03
[CI] Add GitHub CI job running clang-format


Rbiessy:
# Summary
It would be useful if the GitHub CI would check that a PR is formatted before merging it.

# Problem statement
Automatically checking the code is formatted is safer than having a check box on a PR.
I have also seen some differences in the formatted code which may be due to different `clang-format` versions. Integrating this check in GitHub would clarify what the source of truth is.

# Preferred solution
The preferred solution is for GitHub to have a job that runs `clang-format`. It should be possible to set it up so that the action automatically pushes the commit fixing the formatting. If this is not possible showing the diff would be useful.
I would suggest to use whichever clang-format version is shipped with the oneAPI base toolkit in `oneapi/compiler/latest/bin/compiler/clang-format`, assuming there is no plan to remove that.

Rbiessy:
Pinging @oneapi-src/onemkl-arch-write and @rscohn2 who may have opinions on that.

dnhsieh-intel:
I plan to work on this. If we are OK with the old `clang-format-9`, which is used in the internal CI, we can continue using that version. Or we can switch to a newer version together in our CI. This direction needs more work. For `clang-format` in the base toolkit, my concern is that it might be extra work for contributors to download the package if they want to run the format check and it could be annoying if it gives different results from release to release.

Rbiessy:
That's good to know, thanks!
On the contrary, I think my suggestion to use the `clang-format` binary from the oneAPI package should avoid developers to download more packages. As of today DPC++ is by far the most supported SYCL implementation in oneMKL Interface so most people will have it available already.
Besides `clang-format-9` is very old, the latest version available is 19. This means that not only very few developers have this package lying around already but also we can run into weirdly formatted code.
I don't expect there will be much changes from release to release if any to be honest.

rscohn2:
I use pre-commit to manage the code checks. It will automatically download clang-format and you have full control over versions. It will optionally manage a git commit hook to run the checkers, or you can run it manually. It supports excluding directories, etc. It integrates 100's of checkers.

pre-commit is a python package so developers would create a python virtualenv and install pre-commit into it. You do not typically have to install anything else.

Here is a sample: https://github.com/oneapi-src/distributed-ranges/blob/main/.pre-commit-config.yaml
The local id at the bottom are project-specific checks that are integrated into the framework.

The minimal bootstrap is:
```
python3 -m venv .venv
source .venv/bin/activate
[...]

PR #594, opened 2024-10-15
[clang-format] Reformat due to transition from v9.0.0 to v19.1.0 and update of style configuration

dnhsieh-intel:
# Description

This PR is a preparation step for issue #580. Since we will move from `clang-format` v9.0.0 to v19.1.0, this PR reformats the code according to v19.1.0. This PR is separated from the CI enabling ([PR #595](https://github.com/oneapi-src/oneMKL/pull/595)) in order to have a more informative git blame.

There are some changes I want to discuss.
- `src/lapack/backends/cusolver/cusolver_lapack.cpp`
  The following changes are suggested by `clang-format`:
  ```diff
  -    queue.submit([&](sycl::handler &cgh) {
  -        onemkl_cusolver_host_task(cgh, queue, [=](CusolverScopedContextHandler &sc) {
  -            auto handle = sc.get_handle(queue);
  -            cusolverStatus_t err;
  -            CUSOLVER_ERROR_FUNC_T_SYNC(func_name, func, err, handle, m, n, scratch_size);
  -        });
  -    }).wait();
  +    queue
  +        .submit([&](sycl::handler &cgh) {
  +            onemkl_cusolver_host_task(cgh, queue, [=](CusolverScopedContextHandler &sc) {
  +                auto handle = sc.get_handle(queue);
  +                cusolverStatus_t err;
  +                CUSOLVER_ERROR_FUNC_T_SYNC(func_name, func, err, handle, m, n, scratch_size);
  +            });
  +        })
  +        .wait();
  ```
  Because it looks too funny, I added `// clang-format off/on` for all the similar suggestions in this file.
- `tests/unit_tests/dft/include/{compute_out_of_place,compute_inplace}.hpp`
  Although I don't like the following formatting personally, it is probably still fine. I can turn the formatting off if there are objections. 
  ```diff
  -                EXPECT_TRUE(check_equal_strided<domain == oneapi::mkl::dft::domain::REAL>(
  -                    bwd_ptr + backward_distance * i, out_host_ref.data() + ref_distance * i, sizes,
  -                    strides_bwd_cpy, abs_error_margin, rel_error_margin, std::cout));
  +                EXPECT_TRUE(check_equal_strided < domain ==
  +                            oneapi::mkl::dft::domain::REAL >
  +                                (bwd_ptr + backward_distance * i,
  +                                 out_host_ref.data() + ref_distance * i, sizes, strides_bwd_cpy,
  +                                 abs_error_margin, rel_error_margin, std::cout));
  ```

# Checklist

## All Submissions

- Do all unit tests pass locally? Only code reformatting. The CPU tests in CI passed.


Rbiessy:
The diff looks good to me overall.
Regarding the changes in `src/lapack/backends/cusolver/cusolver_lapack.cpp`, could we try to change the clang-format configuration file to get a good enough formatting without `// clang-format off/on`?
Looking at the repository's `_clang-format` and the [documentation](https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormatStyleOptions.html) there are a bunch of deprecated configurations that we should update:

- `Standard: Cpp11` -> `Standard: c++17` (or `Standard: Latest`)
- `AllowAllConstructorInitializersOnNextLine` and `ConstructorInitializerAllOnOneLineOrOnePerLine` can be replaced with `PackConstructorInitializers: Never`
- `AlwaysBreakAfterDefinitionReturnType` can be removed
- `AlwaysBreakAfterReturnType ` -> `BreakAfterReturnType: Automatic`
- `AlwaysBreakTemplateDeclarations` -> `BreakTemplateDeclarations: Yes`
- `KeepEmptyLinesAtTheStartOfBlocks` ->
  ```
  KeepEmptyLines:
    AtEndOfFile: true
    AtStartOfBlock: false
    AtStartOfFile: false
  ```
- `SpaceInEmptyParentheses` -> `SpacesInParens: Never`

Maybe this will be enough to fix the issue. If not we could maybe add the configuration `PenaltyIndentedWhitespace: 500` for instance.

andrewtbarker:
I agree with @Rbiessy that we should try to improve the configuration file rather than using `// clang-format off`. This PR is a good opportunity to improve the config file.

dnhsieh-intel:
It looks like this is just how `clang-format` handles method chaining:
```c++
    queue
        .submit([&](sycl::handler& cgh) {
            // body
            });
        })
        .wait();
```
I didn't find an option for this unfortunately. One workaround is to write `queue.submit()` and `queue.wait()` separately, but this is against the purpose of using automatic formatting. I think we just let `clang-format` do its job...

The style configuration file is updated in the commit ff81d95. Most changes and clean-up are straightforward. However, one significant change is `DerivePointerAlignment: false`. [The description of this option](https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormatStyleOptions.html#derivepointeralignment) is
>If `true`, analyze the formatted file for the most common alignment of `&` and `*`. Pointer and reference alignment styles are going to be updated according to the preferences found in the file. `PointerAlignment` is then used only as fallback.

Because the alignment style is according to **the preferences found in the file**, the current code base has both left and right alignments, for example,
https://github.com/oneapi-src/oneMKL/blob/6923d402d5bccba9ae1966062bc5a277fc74776c/examples/blas/run_time_dispatching/level3/gemm_usm.cpp#L63
https://github.com/oneapi-src/oneMKL/blob/6923d402d5bccba9ae1966062bc5a277fc74776c/examples/blas/compile_time_dispatching/level3/gemm_usm_mklcpu_cublas.cpp#L64
I am open to either left or right alignment, but we must choose one in my opinion.

Rbiessy (approved):
The changes look good to me, thanks.
I much prefer the left alignment of pointers and reference!
I don't mind too much about the `queue.[...].wait()` pattern. As you said we can split these lines if needed later.

Rbiessy:
I didn't mean to approve on behalf of the other teams, I don't know of way to change this in GitHub so I have just requested for their approval again!

andrewtbarker (approved):
Thanks for doing this.

tests/unit_tests/main_test.cpp
```diff
@@ -114,42 +114,37 @@ int main(int argc, char** argv) {
                         unique_devices.insert(dev.get_info<sycl::info::device::name>());
 #if !defined(ONEMKL_ENABLE_MKLCPU_BACKEND) &&             \
     !defined(ONEMKL_ENABLE_PORTBLAS_BACKEND_INTEL_CPU) && \
-    !defined(ONEMKL_ENABLE_PORTFFT_BACKEND) && \
-    !defined(ONEMKL_ENABLE_NETLIB_BACKEND)
+    !defined(ONEMKL_ENABLE_PORTFFT_BACKEND) && !defined(ONEMKL_ENABLE_NETLIB_BACKEND)
```
sknepper, line 117: I have a slight preference for the previous way this was formatted (each clause on its own line), but I don't think we have too many places where this happens, and it's still understandable.

sknepper (approved):
Thank you for all your work here, @dnhsieh-intel !
This is wonderful - looking forward to the upcoming CI enablement as well!

mkrainiuk (approved):

tests/unit_tests/main_test.cpp
dnhsieh-intel: I think (almost) each clause on its own line was introduced in [PR #571](https://github.com/oneapi-src/oneMKL/pull/571/files#diff-555a5cfce78e81f8a579d1898a92141512b7d89140732263d0dc6ed8b444c221), and we haven't fixed its formatting. Actually `clang-format-9` and the old style file would give the same result.

dnhsieh-intel:
@oneapi-src/onemkl-rng-write, @oneapi-src/onemkl-sparse-write, @oneapi-src/onemkl-dft-write I plan to merge the PR on 10/22 EOD Pacific time if there are no objections.

tests/unit_tests/dft/include/compute_inplace.hpp
```diff
@@ -94,9 +94,12 @@ int DFT_Test<precision, domain>::test_in_place_buffer() {
             auto acc_host = inout_buf.get_host_access();
             auto ptr_host = reinterpret_cast<FwdOutputType*>(acc_host.get_pointer());
             for (std::int64_t i = 0; i < batches; i++) {
-                EXPECT_TRUE(check_equal_strided<domain == oneapi::mkl::dft::domain::REAL>(
```
anantsrivastava30, line 97: Looks good to me, this is the only point that I prefer the old syntax as it's slightly more readable, but as these are unit test its fine. 
Thank you for this effort.

PR #595, opened 2024-10-15
[CI] Enable clang-format check

### LINKED ISSUE #580: [CI] Add GitHub CI job running clang-format

dnhsieh-intel:
# Description

According to the discussion in issue #580, this PR adds a job in CI to check code formatting using `pre-commit` with clang-format v19.1.0.

This PR will be rebased after [PR #594](https://github.com/oneapi-src/oneMKL/pull/594) is merged. We should have a green clang-format check then.

Fixes #580

# Checklist

## All Submissions

- Do all unit tests pass locally? No changes that affect tests.
- Have you formatted the code using clang-format? N/A

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -14,6 +14,19 @@ env:
   PARALLEL: -j 2
 
 jobs:
+  format-checks:
+    runs-on: ubuntu-latest
+    name: clang-format check
+    steps:
+    - uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # v4.1.5
+    - uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
+      with:
+        python-version: '3.10'
+        cache: 'pip'
+    - name: Install pre-commit
+      run: pip install pre-commit
+    - name: Run clang-format check
+      run: pre-commit run --all-files --show-diff-on-failure --color always
   unit-tests:
     runs-on: ubuntu-latest
     # One runner for each domain
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -13,7 +13,6 @@ Fixes # (GitHub issue)
 ## All Submissions
 
 - [ ] Do all unit tests pass locally? Attach a log.
-- [ ] Have you formatted the code using clang-format?
 
 ## New interfaces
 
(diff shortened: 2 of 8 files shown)


rscohn2 (approved):
It looks good.

.github/workflows/pr.yml
```diff
@@ -14,6 +14,19 @@ env:
...
+    steps:
+    - uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # v4.1.5
+    - uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
+      with:
+        python-version: '3.10'
+        cache: 'pip'
+    - name: Install pre-commit
+      run: pip install pre-commit
+    - name: Run clang-format check
+      run: pre-commit run --all-files --show-diff-on-failure --color always
```
Rbiessy, line 29: Is it possible to run clang-format only on the changes of the PR with `pre-commit`? It could be useful to save a bit on the ressources used.
With `git-clang-format` it is quite easy but I don't know if that's usable with `pre-commit`.

rscohn2: You can omit the `--all-files` and it will only check the changed files.

This step in CI took 11 seconds so there is not much savings for this project. When you update the clang-format version, it is not smart enough to force checking on all files and the --all-files will catch that.

Rbiessy: Right, I didn't notice it was that fast with the GitHub CI. I suppose it's fine then.

mkrainiuk (approved):
Finally we won't need to run it manually, thank you for this improvement!

dnhsieh-intel:
@Rbiessy I am merging the PR in case other commits get between this one and the reformatting. I will open a new one if you have suggested changes.

Rbiessy:
I had no concerns with this PR, thanks!

PR #597, opened 2024-10-18
[cublas][curand][cusolver] Fix OpenCL include dir hint

npmiller:
In recent versions of DPC++ the CL headers go in `include` instead of `include/sycl`. So this patch fixes finding the OpenCL headers. Keep the hints for the previous path to still support older version.


Rbiessy:
Won't `FindcuSOLVER.cmake` and `FindcuRAND.cmake` run into the same issue?

npmiller:
Maybe, we just ran into this for cuBLAS, I'll see if the other ones need it too

Rbiessy (approved):

sknepper (approved):
This is great, thanks @npmiller !

Issue #567, opened 2024-09-05
Using MKL::MKL to resolve the -fsycl issue with AdaptiveCpp fails.


lrpablo:
# Summary
Trying to link the examples when compiling with AdaptiveCpp using MKL::MKL as the target library fails.

# Version
oneMKL version 0.2.0
githash = f4d6b1994eb110587ced0fc9777bac7550498874.

# Environment

* OS name and version=Ubuntu 22.04.4 LTS
* Compiler version=AdaptiveCpp version: 24.02.1+git.d1341108.20240718.branch.develop.dirty

* CMake output log
-- CMAKE_BUILD_TYPE: None, set to Release by default
-- The CXX compiler identification is Clang 18.1.8
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /home/pablo/usr/local/bin/acpp - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- TARGET_DOMAINS: blas
-- MKL_VERSION: 2025.0.0
-- MKL_ROOT: /opt/intel/oneapi/mkl/2025.0
-- MKL_SYCL_ARCH: intel64
-- MKL_ARCH: intel64
-- MKL_SYCL_LINK: dynamic
-- MKL_LINK: dynamic
-- MKL_SYCL_INTERFACE_FULL: intel_ilp64
-- MKL_INTERFACE_FULL: intel_ilp64
-- MKL_SYCL_THREADING: tbb_thread
-- MKL_THREADING: tbb_thread
-- MKL_MPI: None, set to ` intelmpi` by default
-- Found /opt/intel/oneapi/mkl/2025.0/lib/libmkl_scalapack_ilp64.so
-- Found /opt/intel/oneapi/mkl/2025.0/lib/libmkl_cdft_core.so
-- Found /opt/intel/oneapi/mkl/2025.0/lib/libmkl_intel_ilp64.so
-- Found /opt/intel/oneapi/mkl/2025.0/lib/libmkl_tbb_thread.so
-- Found /opt/intel/oneapi/mkl/2025.0/lib/libmkl_core.so
-- Found /opt/intel/oneapi/mkl/2025.0/lib/libmkl_blacs_intelmpi_ilp64.so
-- Found /opt/intel/oneapi/mkl/2025.0/lib/libmkl_sycl_blas.so
[...]

PR #609, opened 2024-10-30
[BLAS] Simplify CublasScopedContextHandler

keyradical:
Nowadays we only use the cuda primary context in DPC++, hence we can refactor `CublasScopedContextHandler`.
 
 - Since now we only use the primary context which is unique to a device, we can change the 
   `unordered_map<context, handle>` to be `unordered_map<device, handle>`,
- I believe that the call to `sycl::detail::contextSetExtendedDeleter` is not necessary. It provides additional feature tying the lifetime of cublasHandle to the corresponding sycl queue but it makes the code very complicated (with atomics etc.) and uses the detail namespace which is not ideal.
- `cublas_handle.hpp` was modified such that it is not templated anymore. Both DPC++ and AdaptiveCpp versions use the same mapping so that could be simplified. Also, we need to set the correct context in the custom destructor in order to destroy cublas handles so the native type was necessary there.

Side note:
We can definitely remove the dependency on UR headers just by changing the templated types to the native types. 

[test_main_blas_rt.txt](https://github.com/user-attachments/files/17577366/test_main_blas_rt.txt)
[test_main_blas_ct.txt](https://github.com/user-attachments/files/17577367/test_main_blas_ct.txt)

### DIFF
--- a/src/blas/backends/cublas/cublas_scope_handle.cpp
+++ b/src/blas/backends/cublas/cublas_scope_handle.cpp
@@ -39,32 +39,15 @@ thread_local cublas_handle<CUcontext> CublasScopedContextHandler::handle_helper 
     cublas_handle<CUcontext>{};
 
 CublasScopedContextHandler::CublasScopedContextHandler(sycl::queue queue, sycl::interop_handle& ih)
-        : ih(ih),
-          needToRecover_(false) {
+        : ih(ih) {
     placedContext_ = new sycl::context(queue.get_context());
     auto cudaDevice = ih.get_native_device<sycl::backend::ext_oneapi_cuda>();
     CUresult err;
     CUcontext desired;
-    CUDA_ERROR_FUNC(cuCtxGetCurrent, err, &original_);
     CUDA_ERROR_FUNC(cuDevicePrimaryCtxRetain, err, &desired, cudaDevice);
-    if (original_ != desired) {
-        // Sets the desired context as the active one for the thread
-        CUDA_ERROR_FUNC(cuCtxSetCurrent, err, desired);
-        // No context is installed and the suggested context is primary
-        // This is the most common case. We can activate the context in the
-        // thread and leave it there until all the PI context referring to the
-        // same underlying CUDA primary context are destroyed. This emulates
-        // the behaviour of the CUDA runtime api, and avoids costly context
-        // switches. No action is required on this side of the if.
-        needToRecover_ = !(original_ == nullptr);
-    }
 }
 
 CublasScopedContextHandler::~CublasScopedContextHandler() noexcept(false) {
-    if (needToRecover_) {
-        CUresult err;
-        CUDA_ERROR_FUNC(cuCtxSetCurrent, err, original_);
-    }
     delete placedContext_;
 }
 
--- a/src/blas/backends/cublas/cublas_scope_handle.cpp
+++ b/src/blas/backends/cublas/cublas_scope_handle.cpp
@@ -39,36 +39,7 @@ thread_local cublas_handle<CUcontext> CublasScopedContextHandler::handle_helper 
     cublas_handle<CUcontext>{};
 
 CublasScopedContextHandler::CublasScopedContextHandler(sycl::queue queue, sycl::interop_handle& ih)
-        : ih(ih) {
-    placedContext_ = new sycl::context(queue.get_context());
-    auto cudaDevice = ih.get_native_device<sycl::backend::ext_oneapi_cuda>();
-    CUresult err;
-    CUcontext desired;
-    CUDA_ERROR_FUNC(cuDevicePrimaryCtxRetain, err, &desired, cudaDevice);
-}
-
-CublasScopedContextHandler::~CublasScopedContextHandler() noexcept(false) {
-    delete placedContext_;
-}
-
-void ContextCallback(void* userData) {
-    auto* ptr = static_cast<std::atomic<cublasHandle_t>*>(userData);
-    if (!ptr) {
-        return;
-    }
-    auto handle = ptr->exchange(nullptr);
-    if (handle != nullptr) {
-        cublasStatus_t err1;
-        CUBLAS_ERROR_FUNC(cublasDestroy, err1, handle);
-        handle = nullptr;
-    }
-    else {
-        // if the handle is nullptr it means the handle was already destroyed by
-        // the cublas_handle destructor and we're free to delete the atomic
-        // object.
-        delete ptr;
-    }
-}
+        : ih(ih) {}
 
 cublasHandle_t CublasScopedContextHandler::get_handle(const sycl::queue& queue) {
     auto cudaDevice = ih.get_native_device<sycl::backend::ext_oneapi_cuda>();
@@ -105,9 +76,6 @@ cublasHandle_t CublasScopedContextHandler::get_handle(const sycl::queue& queue) 
 
     auto insert_iter = handle_helper.cublas_handle_mapper_.insert(
         std::make_pair(desiredCtx, new std::atomic<cublasHandle_t>(handle)));
-
-    sycl::detail::pi::contextSetExtendedDeleter(*placedContext_, ContextCallback,
-                                                insert_iter.first->second);
 
     return handle;
 }
--- a/src/blas/backends/cublas/cublas_scope_handle.cpp
+++ b/src/blas/backends/cublas/cublas_scope_handle.cpp
@@ -35,13 +35,8 @@ namespace cublas {
  * takes place if no other element in the container has a key equivalent to
  * the one being emplaced (keys in a map container are unique).
  */
-#ifdef ONEMKL_PI_INTERFACE_REMOVED
-thread_local cublas_handle<ur_context_handle_t> CublasScopedContextHandler::handle_helper =
-    cublas_handle<ur_context_handle_t>{};
-#else
-thread_local cublas_handle<pi_context> CublasScopedContextHandler::handle_helper =
-    cublas_handle<pi_context>{};
-#endif
+thread_local cublas_handle<CUcontext> CublasScopedContextHandler::handle_helper =
+    cublas_handle<CUcontext>{};
 
 CublasScopedContextHandler::CublasScopedContextHandler(sycl::queue queue, sycl::interop_handle& ih)
         : ih(ih),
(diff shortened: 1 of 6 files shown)


keyradical:
@oneapi-src/onemkl-blas-write any thoughts on this?

andrewtbarker:
I very much like the simplification here. If a user has two cuda devices available, how do they specify which one to use?

keyradical:
> I very much like the simplification here. If a user has two cuda devices available, how do they specify which one to use?

They would specify that at the creation of a `sycl::queue`. Each queue can be mapped only to a single device and `CublasScopedContextHandler` constructor takes a queue parameter. My understanding is that we want to use one `cublasHandle_t` per one cuda device.

andrewtbarker (approved):

keyradical:
Thanks for the review @andrewtbarker! In that case I will apply similar changes to other backends as well.

src/blas/backends/cublas/cublas_scope_handle.cpp
```diff
@@ -17,126 +17,28 @@
...
-
-void ContextCallback(void* userData) {
-    auto* ptr = static_cast<std::atomic<cublasHandle_t>*>(userData);
-    if (!ptr) {
-        return;
-    }
-    auto handle = ptr->exchange(nullptr);
-    if (handle != nullptr) {
-        cublasStatus_t err1;
-        CUBLAS_ERROR_FUNC(cublasDestroy, err1, handle);
```
Rbiessy, line 84: It seems we never destroy the cublas handle anymore. I get that the existing solution relies on `sycl::detail::pi::contextSetExtendedDeleter` which is not ideal, is there no other way we could destroy the cublas handle when the SYCL queue is destroyed? This seems like a useful feature.

src/blas/backends/cublas/cublas_scope_handle.cpp
```diff
@@ -17,126 +17,28 @@
...
-    }
-    else {
-        // if the handle is nullptr it means the handle was already destroyed by
-        // the cublas_handle destructor and we're free to delete the atomic
-        // object.
-        delete ptr;
-    }
+          nativeDevice(ih.get_native_device<sycl::backend::ext_oneapi_cuda>()) {
+    cublasStatus_t err;
+    CUBLAS_ERROR_FUNC(cublasCreate, err, &cublasHandle);
```
Rbiessy: This is a big change, the cublas handle used to be cached in the global map so it is created only once per device and host thread. Now cublas handles will be created in every host_task (and never destroyed, see my previous comment).
I think we should keep the ability to cache the cublas handles.

Rbiessy (commented):
Sorry for the late review. I agree that the `CublasScopedContextHandler` can be improved but I have 2 big concerns with the suggested changes.
Ideally we should have a closer look at the impact of this change on an application using oneMKL+cuBLAS or create an example that calls a few oneMKL functions.

src/blas/backends/cublas/cublas_scope_handle.cpp
keyradical: Ah, that's true. This is something I totally forgot about when I switched to this approach in one of the last commits. 

We definitely should destroy the cublas handle at some point. From what I understand, there  were 2 places in the existing solution when the handle was destroyed. One of them was using `contextSetExtendedDeleter` and the other was when the destructor of the `cublas_handle` wrapper was called - that would be invoked at the end of the program, when the current host thread was destroyed. 

In that case I'm in favor of the latter solution. Destroying the cublas handle at the end of the program. We can achieve that in my new version by wrapping the member `cublasHandle_t` in a `thread_local shared_ptr` and manually specifying its destructor so that it calls `cublasDestroy`.

Rbiessy: Right I forgot the handle could also be destroyed when destroying `cublas_handle`.
I'm not sure it is possible to destroy it at the end of the program with the current solution. You have to be very careful about the fact that `host_task` are used which spawn their own threads. This means every `host_task` creates its own `CublasScopedContextHandler`, have their own copy of `cublas_handle` as it is `thread_local` in the develop branch.
I am worried that your suggested solution would still create and destroy the cublas handle in every host task. I don't know how this could be improved, I am just very wary of any change in the scope handle!

src/blas/backends/cublas/cublas_scope_handle.cpp
keyradical: What do you mean that it will be created in every host_task? Does that mean that in a typical application which uses this, we would call several host_tasks in a single thread where each of those would instantiate `CublasScopedContextHandler`?

src/blas/backends/cublas/cublas_scope_handle.cpp
keyradical: Yes, it's been quite a ride trying to untangle all of this so far! :wink: 

I think if that's the case with the host task, we just need to make the member `cublasHandle_t` thread_local and shared ptr and add a destructor to it which destroys it at the end of the program. 

I'm actually off sick today so I will think about it more and get back to you! Thanks for the review tho, it's much appreciated.

src/blas/backends/cublas/cublas_scope_handle.cpp
Rbiessy: > What do you mean that it will be created in every host_task?

What I mean is that every compute function will launch a `host_task` which will call the constructor of `CublasScopedContextHandler` [here](https://github.com/oneapi-src/oneMKL/blob/develop/src/blas/backends/cublas/cublas_task.hpp#L75).

A typical application will have a main thread which will submit multiple `host_task`. From there it is hard to predict how many threads will be used, it depends on the SYCL runtime and on the graph dependencies. I believe DPC++ will try to re-use the same thread for 2 consecutive `host_task` with a linear dependencies. I'm not 100% sure what happens to the lifetime of the different objects when the thread is re-used.
I believe that with the version in upstream the application may end up having multiple cublas handle for each thread in the DPC++ thread pool (which is maybe not ideal but seems safer) so the number of cublas handle would end up bounded by the number of threads used by DPC++.
I believe that with your suggested solution every `host_task` submitted will simply create a new cublas handle i.e. the more functions from oneMKL BLAS are called the more cublas handle will be created.

keyradical: Right, I think I see what potentially could be a problem here and why the destruction of the cublas handle was also tied to the sycl queue.

Is it the worry that a given host_task thread will create `cublasHandle 0` tied to queue 0 or in another words: to device 0 but then that thread might be later reused for another queue, e.g. queue 1. In that case this thread might not be destroyed before this reusing and hence `cublasHandle 0` won't be destroyed. For that purpose we have it destroyed when queue 0 is destroyed?

keyradical: I think this causes problems if that thread can in fact be reused for another device. Otherwise I guess it is fine since we would still preserve the mapping of `unique cublasHandle -> unique device`

src/blas/backends/cublas/cublas_scope_handle.cpp
Rbiessy: The destruction issue is solved I think so resolving this, thanks

src/blas/backends/cublas/cublas_scope_handle.cpp
Rbiessy: Using the `static thread_local` cublasHandle solved my main concern. I don't know what were the intentions of destroying the cublasHandle with the SYCL queue originally. To me this is to make sure the user has a way to free these ressources not just when the program ends.
I think you make a good point with the multiple queues, there may be a risk that DPC++ uses the same thread for 2 host_task of 2 different queues. Can you investigate whether this is possible in DPC++?

src/blas/backends/cublas/cublas_scope_handle.hpp
```diff
@@ -23,27 +23,6 @@
...
-#if __has_include(<sycl/detail/ur.hpp>)
-#include <sycl/detail/ur.hpp>
-#ifndef ONEMKL_PI_INTERFACE_REMOVED
-#define ONEMKL_PI_INTERFACE_REMOVED
-#endif
-#elif __has_include(<sycl/detail/pi.hpp>)
-#include <sycl/detail/pi.hpp>
-#else
-#include <CL/sycl/detail/pi.hpp>
-#endif
```
Rbiessy, line 46: The header file `cublas_handle.hpp` below can be completely removed now.

src/blas/backends/cublas/cublas_scope_handle.hpp
```diff
@@ -67,39 +46,18 @@ According to NVIDIA:
...
     sycl::context get_context(const sycl::queue& queue);
 
 public:
-    CublasScopedContextHandler(sycl::queue queue, sycl::interop_handle& ih);
+    CublasScopedContextHandler(sycl::interop_handle& ih);
 
-    ~CublasScopedContextHandler() noexcept(false);
     /**
    * @brief get_handle: creates the handle by implicitly impose the advice
    * given by nvidia for creating a cublas_handle. (e.g. one cuStream per device
```
Rbiessy, line 74: This comment should be updated (i.e. `cublas_handle` does not exist anymore).

src/blas/backends/cublas/cublas_scope_handle.hpp
keyradical: `cublas_handle` is still [being used](https://github.com/oneapi-src/oneMKL/blob/8f4312ef966420b9b8b4b82b9d5c22e2c91a1fe7/src/blas/backends/cublas/cublas_scope_handle_hipsycl.cpp#L27) by the hipSYCL version, I'm not sure what to do about that.

Rbiessy: Right, there are some known issues with AdaptiveCpp and not enough workforce to fix them so let's just keep the file.

Rbiessy: I think `cublas_handle.hpp` doesn't need to be included in `cublas_scope_handle.hpp` though?

src/blas/backends/cublas/cublas_scope_handle.cpp
keyradical: The more I think about this the more issues I'm finding with the current version of this PR. 
You've mentioned that:
> A typical application will have a main thread which will submit multiple host_task. From there it is hard to predict how many threads will be used, it depends on the SYCL runtime and on the graph dependencies. I believe DPC++ will try to re-use the same thread for 2 consecutive host_task with a linear dependencies.

So just to confirm, the main reason for `static thread_local` cublasHandle, is that a thread will be *reused*? And at the point when it is being reused, the handle will still be alive and we won't have to create it again?

Rbiessy: That is correct. Just to be clear there is also another important reason to make it `thread_local` as mentioned in a comment of `cublas_scope_handle.hpp`
```
It is recommended (but not required, if care is taken) that different handles be used for different host threads
```
I think if it weren't `thread_local` there would be a risk that 2 host_task on different threads use the same cublasHandle.

keyradical:
I modified this PR such that the cublasHandle(s) are destroyed only in one place: at the end of the program. Because of all the threading shenanigans I believe the `static thread_local unordered_map` needs to stay. 

By removing the use of `sycl::detail::contextSetExtendedDeleter` we aren't allowing the possibility of `cublasHandle` being destroyed when `sycl::queue` is destroyed. This is useful feature but it makes this code very complicated and I'm not sure exactly how much it is actually used in practice. If it is really necessary I think we should provide a public API for it in DPC++ and not use the `detail` namespace. 

The good news is that those changes make AdaptiveCpp and DPC++ implementations identical (if that's of any benefit).

Rbiessy:
Alright, that looks fine to me then.

> The good news is that those changes make AdaptiveCpp and DPC++ implementations identical (if that's of any benefit).

Do you think it would be possible to remove `cublas_scope_handle_hipsycl.hpp` and `cublas_scope_handle_hipsycl.cpp` as part of this PR then? I'm not sure if the blas domain still compile with AdaptiveCpp so it may be too difficult to test right now.

keyradical:
Good point, I think those files can be removed but I don't have much experience with AdaptiveCpp so I wouldn't be able to test the build (at least right now).

Rbiessy:
Looking at the internal CI, it does not seem that the blas domain compiles with AdaptiveCpp anyway. One of the issue is https://github.com/oneapi-src/oneMKL/issues/567. I see that there are still some differences like `ih.get_native_device<sycl::backend::ext_oneapi_cuda>()` vs `interop_h.get_native_device<sycl::backend::cuda>()` so let's not bother trying to fix that today. I think we can merge this next week, thanks for the work!

Rbiessy (approved):

PR #620, opened 2024-12-11
Remove deprecated scripts and BLAS list files

vmalia:
As a result of https://github.com/uxlfoundation/oneMath/discussions/598 discussion, this PR removes the outdated scripts and updates the `create_new_backend` document accordingly.

Document preview: https://github.com/uxlfoundation/oneMath/blob/96e0d13ee2224619af396b892a58926236e36476/docs/create_new_backend.rst


scripts/generate_cmake.py
```diff
@@ -1,129 +0,0 @@
-#!/usr/bin/env python
```
Rbiessy, line 1: We should also remove this line: https://github.com/uxlfoundation/oneMath/blob/develop/.github/CODEOWNERS#L14

Rbiessy (commented):
LGTM, overall!

scripts/generate_cmake.py
vmalia: Good catch. Updated.

Rbiessy (approved):

spencerpatty (approved):
changes look fine to me.  nice to see the "+" and "-" highlighted in the examples ... makes it simpler to follow what is necessary.   
 The text changes and removal of python scripts is also good, to focus on most important pieces of changes needed.  Approved for merge.

PR #621, opened 2024-12-12
cmake: Only set deprecated property when supported

### LINKED ISSUE #619: [cmake] Update supported cmake version
# Summary
Version of supported cmake shall be updated.

# Observed behavior
When building with cmake 3.22.x build fails with error:
`CMake Error at src/CMakeLists.txt:47 (set_target_properties):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "DEPRECATION" is not allowed.`
Updating cmake to 3.25.x solves the issue.

Rbiessy:
# Description

CMake versions older than 3.22 do not supported setting the deprecated property on an `INTERFACE` target.
I have tested with CMake 3.22 and 3.13 locally.

Fixes https://github.com/uxlfoundation/oneMath/issues/619

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -44,7 +44,9 @@ function(add_deprecated_library target_name)
   # The INTERFACE IMPORTED target works like an alias which can have different properties
   add_library(deprecated_name INTERFACE IMPORTED)
   target_link_libraries(deprecated_name INTERFACE target_name)
-  set_target_properties(deprecated_name PROPERTIES DEPRECATION "${deprecated_name} target is deprecated, please use ${target_name} instead")
+  if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.22.1")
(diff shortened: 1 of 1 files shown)


s-Nick:
what happen if the cmake version is older than 3.22? Don't we want to manually leave a deprecation message or something?

Rbiessy:
> what happen if the cmake version is older than 3.22? Don't we want to manually leave a deprecation message or something?

The deprecation property is used to only add a message if a deprecated target is used. In our case we want the deprecated target to be an alias of the new target which does not seem to be possible before 3.22. Adding a `message()` would be printed even in valid usages so we can't do this.
Users using the minimum CMake version won't get a warning when using a deprecated target unfortunately. I don't think this is a good enough reason to bump the CMake version but that would be fine with me.

s-Nick (approved):
LGTM

Rbiessy:
Pinging @vmalia since you are around. Could your team help review this by the end of next week?

dnhsieh-intel (approved):

PR #632, opened 2025-02-20
[RNG][Device API] Added (u)int8_t and (u)int16_t types support for Uniform

egrabovskaya:
# Description

At the DPNP team request support for the following small types was added for Uniform distribution:
- int8_t
- uint8_t
- int16_t
- uint16_t

# Checklist

## All Submissions

### DIFF
--- a/include/oneapi/math/rng/device/detail/uniform_impl.hpp
+++ b/include/oneapi/math/rng/device/detail/uniform_impl.hpp
@@ -123,7 +123,8 @@ protected:
         OutType res;
         if constexpr (std::is_integral<Type>::value) {
             if constexpr (std::is_same_v<Type, std::int8_t> || std::is_same_v<Type, std::uint8_t> ||
-                          std::is_same_v<Type, std::int16_t> || std::is_same_v<Type, std::uint16_t>) {
+                          std::is_same_v<Type, std::int16_t> ||
+                          std::is_same_v<Type, std::uint16_t>) {
                 return generate_single_int<float, OutType>(engine);
             }
             else if constexpr (std::is_same_v<Type, std::int32_t> ||
@@ -248,7 +249,8 @@ protected:
         Type res;
         if constexpr (std::is_integral<Type>::value) {
             if constexpr (std::is_same_v<Type, std::int8_t> || std::is_same_v<Type, std::uint8_t> ||
-                          std::is_same_v<Type, std::int16_t> || std::is_same_v<Type, std::uint16_t>) {
+                          std::is_same_v<Type, std::int16_t> ||
+                          std::is_same_v<Type, std::uint16_t>) {
                 float res_fp =
                     engine.generate_single(static_cast<float>(a_), static_cast<float>(b_));
                 res_fp = sycl::floor(res_fp);
--- a/include/oneapi/math/rng/device/detail/uniform_impl.hpp
+++ b/include/oneapi/math/rng/device/detail/uniform_impl.hpp
@@ -122,8 +122,12 @@ protected:
                                       float>::type;
         OutType res;
         if constexpr (std::is_integral<Type>::value) {
-            if constexpr (std::is_same_v<Type, std::int32_t> ||
-                          std::is_same_v<Type, std::uint32_t>) {
+            if constexpr (std::is_same_v<Type, std::int8_t> || std::is_same_v<Type, std::uint8_t> ||
+                          std::is_same_v<Type, std::int16_t> || std::is_same_v<Type, std::uint16_t>) {
+                return generate_single_int<float, OutType>(engine);
+            }
+            else if constexpr (std::is_same_v<Type, std::int32_t> ||
+                               std::is_same_v<Type, std::uint32_t>) {
                 return generate_single_int<FpType, OutType>(engine);
             }
             else {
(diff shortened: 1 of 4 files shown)


include/oneapi/math/rng/device/detail/uniform_impl.hpp
```diff
@@ -122,8 +122,12 @@ class distribution_base<oneapi::math::rng::device::uniform<Type, Method>> {
                                       float>::type;
         OutType res;
         if constexpr (std::is_integral<Type>::value) {
-            if constexpr (std::is_same_v<Type, std::int32_t> ||
-                          std::is_same_v<Type, std::uint32_t>) {
+            if constexpr (std::is_same_v<Type, std::int8_t> || std::is_same_v<Type, std::uint8_t> ||
+                          std::is_same_v<Type, std::int16_t> || std::is_same_v<Type, std::uint16_t>) {
```
andreyfe1: I would suggest to move this dispatch above to have floating type definition in one place, e.g.
```cpp
std::conditional_t<
    !std::is_same_v<Method, uniform_method::accurate> ||
     std::is_same_v<Type, std::int8_t> ||
     std::is_same_v<Type, std::uint8_t> ||
     std::is_same_v<Type, std::int16_t> ||
     std::is_same_v<Type, std::uint16_t>,
float, double>
```

include/oneapi/math/rng/device/distributions.hpp
```diff
@@ -581,6 +595,10 @@ class poisson : detail::distribution_base<poisson<IntType, Method>> {
 // Supported types:
 //      std::uint32_t
 //      std::int32_t
+//      std::uint16_t
+//      std::int16_t
+//      std::uint8_t
+//      std::int8_t
```
andreyfe1, line 596: Above we have the following order: 
```
//      std::int8_t
//      std::uint8_t
//      std::int16_t
//      std::uint16_t
//      std::int32_t
//      std::uint32_t
//      std::int64_t
//      std::uint64_t
```
Could you please make it consistent with above order?

include/oneapi/math/rng/device/distributions.hpp
```diff
@@ -36,8 +36,14 @@ namespace oneapi::math::rng::device {
 // Supported types:
 //      float
 //      double
+//      std::int8_t>
```
andreyfe1: `>` is extra in the end of line

tests/unit_tests/rng/device/include/rng_device_test_common.hpp
```diff
@@ -187,6 +187,82 @@ struct statistics_device<oneapi::math::rng::device::uniform<Fp, Method>> {
...
+        float b = distr.b();
+
+        // Theoretical moments
+        tM = (a + b - 1.0) / 2.0;
+        tD = ((b - a) * (b - a) - 1.0) / 12.0;
+        tQ = (((b - a) * (b - a)) * ((1.0 / 80.0) * (b - a) * (b - a) - (1.0 / 24.0))) +
+             (7.0 / 240.0);
+
+        return compare_moments(r, tM, tD, tQ);
+    }
```
andreyfe1, line 237: The body of function is the same as for 3 new functions above. Could you please generalize it? It will make future support easier

andreyfe1 (commented):
Thank you very much for the PR!

include/oneapi/math/rng/device/distributions.hpp
egrabovskaya: Done

tests/unit_tests/rng/device/include/rng_device_test_common.hpp
egrabovskaya: Done

include/oneapi/math/rng/device/distributions.hpp
egrabovskaya: Fixed

tests/unit_tests/rng/device/include/rng_device_test_common.hpp
```diff
@@ -166,100 +166,111 @@ bool compare_moments(const std::vector<Fp, AllocType>& r, double tM, double tD,
...
-struct statistics_device<oneapi::math::rng::device::uniform<Fp, Method>> {
-    template <typename AllocType>
-    bool check(const std::vector<Fp, AllocType>& r,
-               const oneapi::math::rng::device::uniform<Fp, Method>& distr) {
-        double tM, tD, tQ;
-        Fp a = distr.a();
-        Fp b = distr.b();
+    double tM, tD, tQ;
+    ParamsType a = distr.a();
+    ParamsType b = distr.b();
```
egrabovskaya: Is it OK if a, b params are always double-precision floating point numbers for all small types?

include/oneapi/math/rng/device/detail/uniform_impl.hpp
egrabovskaya: I think it looks clearer with the extra condition. Furthermore, the existing condition relates to (u)int32, (u)int64 and floating point precision types. But with the option you suggested we will have less additional code.
@iMartyan, could you please share your opinion here?

tests/unit_tests/rng/device/include/rng_device_test_common.hpp
andreyfe1: Could you please check it? If tests are passed I don't see any objections to use `double` for all `Fp` values

egrabovskaya: I launched moments manually and they successfully passed

andreyfe1: So, I'm fine if we switch to `double` if there are no objections from somebody else

include/oneapi/math/rng/device/detail/uniform_impl.hpp
iMartyan: I also like Andrey's approach more because the generation method is the same. So, better not to split it into several branches.

tests/unit_tests/rng/device/include/rng_device_test_common.hpp
iMartyan: fine with me

iMartyan (approved):
lgtm

andreyfe1 (approved):
Looks good to me. Thank you for your efforts

PR #633, opened 2025-02-20
[LAPACK] Add Arm Performance Libraries backend

adegomme:
# Description

This adds the Lapack backend for ArmPl on aarch64 CPUs. 
Batch versions are not yet implemented. 
Due to the LAPACKE interface used in ArmPl not exactly matching oneMath, some _scratchpad sizes are not queried and allocated/deallocated on the go (heevg, hegvd, .. which would need multiple work arrays with different sizes)

# Checklist

## All Submissions

[log_lapack.txt](https://github.com/user-attachments/files/18891851/log_lapack.txt)
100% of the tests are reported as passing, which is an issue actually on the lapack backend (no skipped state, fix coming in another MR)


include/oneapi/math/lapack/detail/armpl/onemath_lapack_armpl.hpp
```diff
@@ -0,0 +1,23 @@
+//
```
sknepper: Please add a copyright

src/lapack/backends/armpl/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
+#
```
sknepper, line 3: Please add a copyright

src/lapack/backends/armpl/armpl_batch.cpp
```diff
@@ -0,0 +1,988 @@
...
+
+#undef POTRF_GROUP_LAUNCHER_SCRATCH
+
+#define POTRS_GROUP_LAUNCHER_SCRATCH(TYPE)                                                   \
+    template <>                                                                              \
+    std::int64_t potrs_batch_scratchpad_size<TYPE>(                                          \
+        sycl::queue & queue, oneapi::math::uplo * uplo, std::int64_t* n, std::int64_t* nrhs, \
+        std::int64_t* lda, std::int64_t* ldb, std::int64_t group_count,                      \
+        std::int64_t* group_sizes) {                                                         \
+        throw unimplemented("lapack", "potrfs_batch_scratchpad_size");                       \
```
sknepper: typo: potrs_ (not potrfs_)

src/lapack/backends/armpl/armpl_batch.cpp
```diff
@@ -0,0 +1,988 @@
...
+inline sycl::event potrf_batch(sycl::queue& queue, oneapi::math::uplo* uplo, std::int64_t* n, T** a,
+                               std::int64_t* lda, std::int64_t group_count,
+                               std::int64_t* group_sizes, T* scratchpad,
+                               std::int64_t scratchpad_size,
+                               const std::vector<sycl::event>& dependencies) {
+    throw unimplemented("lapack", "potrf_batch");
+}
+
+// Scratchpad memory not needed as parts of buffer a is used as workspace memory
+#define POTRF_BATCH_LAUNCHER_USM(TYPE)                                                            \
```
sknepper, line 506: What is this doing, given that all batched LAPACK is unimplemented?

src/lapack/backends/armpl/armpl_common.hpp
```diff
@@ -0,0 +1,162 @@
...
+* and limitations under the License.
+*
+*
+* SPDX-License-Identifier: Apache-2.0
+*******************************************************************************/
+
+#ifndef _ARMPL_COMMON_HPP_
+#define _ARMPL_COMMON_HPP_
+
+#define MKL_Complex8  std::complex<float>
```
sknepper: Unclear why MKL is used here...

Rbiessy (approved):
No concern from my side

src/lapack/backends/armpl/armpl_common.hpp
```diff
@@ -0,0 +1,162 @@
...
+    }
+}
+
+inline char get_jobsvd(oneapi::math::jobsvd job) {
+    switch (job) {
+        case oneapi::math::jobsvd::N: return 'N';
+        case oneapi::math::jobsvd::A: return 'A';
+        case oneapi::math::jobsvd::O: return 'O';
+        case oneapi::math::jobsvd::S: return 'S';
+    }
```
sknepper, line 107: Should there be a default that throws "Wrong job."?

src/lapack/backends/armpl/armpl_common.hpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+inline char get_diag(oneapi::math::diag diag) {
+    switch (diag) {
+        case oneapi::math::diag::N: return 'N';
+        case oneapi::math::diag::U: return 'U';
+    }
```
sknepper, line 115: Similar comment here

src/lapack/backends/armpl/armpl_common.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+template <typename T>
+constexpr bool is_complex = false;
+template <typename T>
+constexpr bool is_complex<std::complex<T>> = true;
+template <>
+inline constexpr bool is_complex<armpl_singlecomplex_t> = true;
+template <>
+inline constexpr bool is_complex<armpl_doublecomplex_t> = true;
+//static constexpr bool is_complex(armpl_singlecomplex_t> = true;
```
sknepper: Are these two commented out lines needed?

src/lapack/backends/armpl/armpl_wrappers.cpp
```diff
@@ -0,0 +1,3050 @@
...
+template <typename Func, typename T>
+void getrf(Func func, sycl::queue& queue, std::int64_t m, std::int64_t n, sycl::buffer<T>& a,
+           std::int64_t lda, sycl::buffer<std::int64_t>& ipiv) {
+    using ArmDataType = typename ArmEquivalentType<T>::Type;
+    queue.submit([&](sycl::handler& cgh) {
+        auto a_acc = a.template reinterpret<ArmDataType>()
+                         .template get_access<sycl::access::mode::read_write>(cgh);
+        auto ipiv_acc = ipiv.template get_access<sycl::access::mode::write>(cgh);
+        host_task<class armpl_kernel_getrf>(cgh, [=]() {
+            func(LAPACK_COL_MAJOR, m, n, a_acc.GET_MULTI_PTR, lda, ipiv_acc.GET_MULTI_PTR);
```
sknepper: Where are computational-time exceptions handled (e.g., matrix given to getrf is singular)?

src/lapack/backends/armpl/armpl_wrappers_table_dyn.cpp
```diff
@@ -0,0 +1,412 @@
+//
```
sknepper: Please add a copyright

sknepper (commented):
Thank you for this awesome contribution! A few questions/comments from my side

include/oneapi/math/lapack/detail/armpl/onemath_lapack_armpl.hpp
adegomme: done

src/lapack/backends/armpl/CMakeLists.txt
adegomme: done

src/lapack/backends/armpl/armpl_batch.cpp
adegomme: good catch, done

src/lapack/backends/armpl/armpl_batch.cpp
adegomme: comment was indeed useless, I removed it

src/lapack/backends/armpl/armpl_common.hpp
adegomme: Old version needed it, but that's no longer the case, I removed it, thanks

src/lapack/backends/armpl/armpl_common.hpp
adegomme: done

src/lapack/backends/armpl/armpl_common.hpp
adegomme: done

src/lapack/backends/armpl/armpl_common.hpp
adegomme: removed

src/lapack/backends/armpl/armpl_wrappers_table_dyn.cpp
adegomme: done

src/lapack/backends/armpl/armpl_wrappers.cpp
adegomme: This is a good point, we don't check outputs yet, as in our tests errors are caught inside the fortran lapack code and usually result in an exit() being triggered. The open dpc++ sycl with pocl backend I use actually fails to identify this exit inside the kernel, and hangs indefinitely after printing the error message, even with an exception handler added.
I don't have a good idea on how to prevent this, as the STOP in lapack's xerbla is a problem and can't be skipped. There is a comment saying implementers could remove it and handle return codes instead, but it seems Arm did not do this. 

At the C level (lapacke) though, a few errors can be caught (usually lapacke_xerbla is called and will print a message, but it will not trigger an exit). I will add some way of triggering an exception in these cases.

adegomme: Lapacke error handling added

src/lapack/backends/armpl/armpl_batch.cpp
sknepper: Actually, it was more a question about the POTRF_BATCH_LAUNCHER_USM rather than the comment. POTRF and POTRS batch launchers are defined, but no other batch functionality has them defined. But it seems like they just call potr{f,s}_batch, which throws an unimplemented exception, so it's unclear why they are needed in the first place.

sknepper (approved):
Thanks for all your work here! I just had one last question, but I don't think the extra code is harming anything.
We don't currently have any aarch64 systems to validate on, so thank you for validating on your side. Thanks again!

src/lapack/backends/armpl/armpl_batch.cpp
adegomme: as far as I remember this was probably adapted from another backend (probably rocsolver), which only implements one version of the two versions of the potrf_batch usm calls (same for potrs, one signature type has pointers, one has not). 
So in rocsolver it's just 4 unimplemented functions + a launcher macro for the 4 implemented ones with a template. And we replaced the implemented ones by a templated unimplemented one. These are just in the end 8 different unimplemented calls for each potrs/f, but with two different ways of doing it mixed up, which is not pretty indeed, but probably harmless.

PR #634, opened 2025-02-20
[RNG] add RNG armpl backend

adegomme:
# Description

Add RNG backend for ArmPl
OpenRNG from ARM, included in ArmPl, is a drop in replacement for OneMKL's VSL interface, most of the interface is the same. Some distributions are not yet implemented in OpenRNG (poisson lognormal).
4 tests (integer uniform with a<0 ) fail currently due to an issue in uniform generation for int32, reported to ARM.  A check has been added to disable them pending next release

# Checklist

## All Submissions

[log_rng.txt](https://github.com/user-attachments/files/18892411/log_rng.txt)
100% tests passed, 0 tests failed out of 154
28 tests skipped


Rbiessy (approved):
No concern from my side

src/rng/backends/CMakeLists.txt
```diff
@@ -36,3 +36,8 @@ if(ENABLE_ROCRAND_BACKEND AND UNIX)
   add_subdirectory(rocrand)
 endif()
 
+if(ENABLE_ARMPL_BACKEND AND UNIX)
```
aelizaro, line 39: As it is only supported for Linux, should https://github.com/uxlfoundation/oneMath/blob/develop/cmake/FindARMPL.cmake (or other cmake files) handle the case when `ENABLE_ARMPL_BACKEND` is specified on Windows to have a clear error message on the configuration stage?

adegomme: I actually discovered recently that there is an ArmPl version available for Windows/Arm platform, so it may actually work. I don't think I'll be able to try the setup with oneMath anytime soon though, so I'm fine adding a message here.

include/oneapi/math/rng/detail/armpl/onemath_rng_armpl.hpp
```diff
@@ -0,0 +1,56 @@
+/*******************************************************************************
+* Copyright 2025 SiPearl
+* Copyright 2020-2021 Intel Corporation
```
andreyfe1, line 3: Could you please clarify why the Intel's copyright is added?

andreyfe1 (commented):
Thanks. Looks good to me overall

include/oneapi/math/rng/detail/armpl/onemath_rng_armpl.hpp
adegomme: hello, most of these files have been duplicated from oneMKL backend, as implementation is roughly the same (OpenRNG from ArmPl is a drop-in replacement for Intel VSL, using same API. Some calls are not supported, others are added, so minor modifications were still necessary). So initial copyright was retained in these files. I can remove it if necessary, but I see this as derivative, so apache license says to retain them.

andreyfe1: ok. Let us take a closer look to determine whether this copyright is necessary or not

andreyfe1: We see no issues with that. Let's keep 2 copyrights in this and similar files

andreyfe1 (approved):
Thanks for the contribution!

PR #635, opened 2025-02-21
[BLAS] Let iamax/iamin accept index_base argument to align to the spec

### LINKED ISSUE #591: blas iamax/iamin cannot accept a oneapi::mkl::index_base argument, which is not align to the document
In the document https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onemkl/source/domains/blas/iamax#onemkl-blas-iamax and https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onemkl/source/domains/blas/iamin, these 2 APIs can accept an oneapi::mkl::index_base argument.

But in the header file, there is no such parameter in the function declaration: https://github.com/oneapi-src/oneMKL/blob/develop/include/oneapi/mkl/blas.hxx#L815 and https://github.com/oneapi-src/oneMKL/blob/develop/include/oneapi/mkl/blas.hxx#L3022

zhiweij1:
# Description

In the spec [iamax](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.4-rev-1/elements/onemath/source/domains/blas/iamax) and [iamin](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.4-rev-1/elements/onemath/source/domains/blas/iamin) can accept an oneapi::mkl::index_base argument.
But in the current impl, there is no such parameter in the function declaration.

Fixes #591

# Checklist

## All Submissions

Only run related tests by `ctest -R Iam`
[cublas.txt](https://github.com/user-attachments/files/18902194/cublas.txt)
[generic.txt](https://github.com/user-attachments/files/18902195/generic.txt)
[mklcpu.txt](https://github.com/user-attachments/files/18902196/mklcpu.txt)
[netlib.txt](https://github.com/user-attachments/files/18902197/netlib.txt)
No env to test Arm and Roc backends.

## Bug fixes

- [X] Have you added relevant regression tests?
      GitHub issue or in this PR)?

### DIFF
--- a/src/blas/backends/armpl/armpl_level1.cxx
+++ b/src/blas/backends/armpl/armpl_level1.cxx
@@ -172,21 +172,24 @@ DOTU_LAUNCHER(std::complex<double>, ::cblas_zdotu_sub)
 
 template <typename T, typename CBLAS_FUNC>
 void iamin(sycl::queue& queue, int64_t n, sycl::buffer<T, 1>& x, int64_t incx,
-           sycl::buffer<int64_t, 1>& result, CBLAS_FUNC cblas_func) {
+           sycl::buffer<int64_t, 1>& result, oneapi::math::index_base base, CBLAS_FUNC cblas_func) {
     queue.submit([&](sycl::handler& cgh) {
         auto accessor_x = x.template get_access<sycl::access::mode::read>(cgh);
         auto accessor_result = result.template get_access<sycl::access::mode::write>(cgh);
         host_task<class armpl_kernel_iamin>(cgh, [=]() {
             accessor_result[0] =
-                cblas_func((armpl_int_t)n, accessor_x.GET_MULTI_PTR, (armpl_int_t)incx);
+                cblas_func((armpl_int_t)n, accessor_x.GET_MULTI_PTR, (armpl_int_t)incx) + base ==
+                        oneapi::math::index_base::zero
+                    ? 0
+                    : 1;
         });
     });
 }
 
 #define IAMIN_LAUNCHER(TYPE, ROUTINE)                                                 \
     void iamin(sycl::queue& queue, int64_t n, sycl::buffer<TYPE, 1>& x, int64_t incx, \
-               sycl::buffer<int64_t, 1>& result) {                                    \
-        iamin(queue, n, x, incx, result, ROUTINE);                                    \
+               sycl::buffer<int64_t, 1>& result, oneapi::math::index_base base) {     \
+        iamin(queue, n, x, incx, result, base, ROUTINE);                              \
     }
 
 IAMIN_LAUNCHER(float, ::cblas_isamin)
@@ -196,21 +199,24 @@ IAMIN_LAUNCHER(std::complex<double>, ::cblas_izamin)
 
 template <typename T, typename CBLAS_FUNC>
 void iamax(sycl::queue& queue, int64_t n, sycl::buffer<T, 1>& x, int64_t incx,
-           sycl::buffer<int64_t, 1>& result, CBLAS_FUNC cblas_func) {
+           sycl::buffer<int64_t, 1>& result, oneapi::math::index_base base, CBLAS_FUNC cblas_func) {
     queue.submit([&](sycl::handler& cgh) {
         auto accessor_x = x.template get_access<sycl::access::mode::read>(cgh);
         auto accessor_result = result.template get_access<sycl::access::mode::write>(cgh);
         host_task<class armpl_kernel_iamax>(cgh, [=]() {
             accessor_result[0] =
-                cblas_func((armpl_int_t)n, accessor_x.GET_MULTI_PTR, (armpl_int_t)incx);
+                cblas_func((armpl_int_t)n, accessor_x.GET_MULTI_PTR, (armpl_int_t)incx) + base ==
+                        oneapi::math::index_base::zero
+                    ? 0
+                    : 1;
         });
     });
 }
 
 #define IAMAX_LAUNCHER(TYPE, ROUTINE)                                                 \
     void iamax(sycl::queue& queue, int64_t n, sycl::buffer<TYPE, 1>& x, int64_t incx, \
-               sycl::buffer<int64_t, 1>& result) {                                    \
-        iamax(queue, n, x, incx, result, ROUTINE);                                    \
+               sycl::buffer<int64_t, 1>& result, oneapi::math::index_base base) {     \
+        iamax(queue, n, x, incx, result, base, ROUTINE);                              \
     }
 
 IAMAX_LAUNCHER(float, ::cblas_isamax)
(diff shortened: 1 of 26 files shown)


zhiweij1:
@Rbiessy Please help to review when you have time. Thanks.

src/blas/backends/mkl_common/mkl_level1.cxx
```diff
@@ -283,43 +283,71 @@ void swap(sycl::queue& queue, std::int64_t n, sycl::buffer<std::complex<double>,
 }
 
 void iamax(sycl::queue& queue, std::int64_t n, sycl::buffer<float, 1>& x, std::int64_t incx,
-           sycl::buffer<std::int64_t, 1>& result) {
-    RETHROW_ONEMKL_EXCEPTIONS(blas_major::iamax(queue, n, x, incx, result));
+           sycl::buffer<std::int64_t, 1>& result, oneapi::math::index_base base) {
+    RETHROW_ONEMKL_EXCEPTIONS(blas_major::iamax(queue, n, x, incx, result,
+                                                base == oneapi::math::index_base::zero
```
andrewtbarker: For the `base` parameter throughout this file, I think it would be better to add a `get_onemkl_base` function in [common_onemkl_conversion.hpp](https://github.com/uxlfoundation/oneMath/blob/develop/src/include/common_onemkl_conversion.hpp#L36) and use it here (the same way `get_onemkl_transpose` is used [here](https://github.com/uxlfoundation/oneMath/blob/develop/src/blas/backends/mkl_common/mkl_level2.cxx#L25)).

src/blas/backends/cublas/cublas_level1.cpp
```diff
@@ -477,15 +478,18 @@ inline void iamax(const char* func_name, Func func, sycl::queue& queue, int64_t
     queue.submit([&](sycl::handler& cgh) {
         auto int_res_acc = int_res_buff.template get_access<sycl::access::mode::read>(cgh);
         auto result_acc = result.template get_access<sycl::access::mode::write>(cgh);
-        cgh.single_task(
-            [=]() { result_acc[0] = std::max((int64_t)int_res_acc[0] - 1, (int64_t)0); });
+        cgh.single_task([=]() {
```
andrewtbarker, line 481: I think it would be better to skip the `single_task` entirely if `base == index_base::one`, ie
```c++
if (base == oneapi::math::index_base::zero) {
    cgh.single_task(
                [=]() { result_acc[0] = std::max((int64_t)int_res_acc[0] - 1, (int64_t)0); });
}
```

src/blas/backends/cublas/cublas_level1.cpp
```diff
@@ -1230,22 +1248,30 @@ inline sycl::event iamin(const char* func_name, Func func, sycl::queue& queue, i
     done.wait();
     if (result_on_device) {
         auto last_ev = queue.submit([&](sycl::handler& cgh) {
```
andrewtbarker, line 1250: Similar to the buffer case, it would be better to skip this queue submission entirely if `base == index_base::one`.

src/blas/backends/rocblas/rocblas_level1.cpp
```diff
@@ -504,15 +505,18 @@ inline void iamax(Func func, sycl::queue& queue, int64_t n, sycl::buffer<T, 1>&
     queue.submit([&](sycl::handler& cgh) {
         auto int_res_acc = int_res_buff.template get_access<sycl::access::mode::read>(cgh);
         auto result_acc = result.template get_access<sycl::access::mode::write>(cgh);
-        cgh.single_task(
-            [=]() { result_acc[0] = std::max((int64_t)int_res_acc[0] - 1, (int64_t)0); });
+        cgh.single_task([=]() {
```
andrewtbarker, line 508: Same comments as for the cublas backend.

andrewtbarker (commented):
Thanks for this improvement!

src/blas/backends/generic/generic_level1.cxx
```diff
@@ -32,22 +32,38 @@ void dotu(sycl::queue& queue, std::int64_t n, sycl::buffer<std::complex<real_t>,
 }
 
 void iamax(sycl::queue& queue, std::int64_t n, sycl::buffer<real_t, 1>& x, std::int64_t incx,
-           sycl::buffer<std::int64_t, 1>& result) {
+           sycl::buffer<std::int64_t, 1>& result, oneapi::math::index_base base) {
     CALL_GENERIC_BLAS_FN(::blas::_iamax, queue, n, x, incx, result);
+    queue.submit([&](sycl::handler& cgh) {
```
Rbiessy: It's not ideal to submit another kernel for this. Ideally it should be done in the same kernel provided by https://github.com/uxlfoundation/generic-sycl-components/tree/main.
I understand this is more work. For now I would suggest to only submit that kernel if the `index_base` is `one` and add a TODO comment that this could be improved. Same comment for the other similar places in the generic backend.

tests/unit_tests/blas/level1/iamax.cpp
```diff
@@ -129,35 +130,77 @@ int test(device* dev, oneapi::math::layout layout, int N, int incx) {
...
+TEST_P(IamaxTests, RealSinglePrecisionBaseZero) {
+    EXPECT_TRUEORSKIP(test<float>(std::get<0>(GetParam()), std::get<1>(GetParam()), 1357, 2,
+                                  oneapi::math::index_base::zero));
+    EXPECT_TRUEORSKIP(test<float>(std::get<0>(GetParam()), std::get<1>(GetParam()), 1357, 1,
+                                  oneapi::math::index_base::zero));
+    EXPECT_TRUEORSKIP(test<float>(std::get<0>(GetParam()), std::get<1>(GetParam()), 1357, -3,
+                                  oneapi::math::index_base::zero));
 }
-TEST_P(IamaxTests, RealDoublePrecision) {
+TEST_P(IamaxTests, RealSinglePrecisionBaseOne) {
```
Rbiessy: I don't think we need to duplicate all the tests to test the parameter `index_base`. If the same logic were applied to other parameters and operations the number of tests would not be manageable. I would suggest to only have `RealSinglePrecision` and one call to the test function with `index_base::one`. Same for other precisions and for iamin.

zhiweij1: fixed in 10760de

src/blas/backends/generic/generic_level1.cxx
zhiweij1: fixed in 10760de

src/blas/backends/mkl_common/mkl_level1.cxx
zhiweij1: fixed in 10760de

src/blas/backends/cublas/cublas_level1.cpp
zhiweij1: I think we cannot skip the entire kernel. Besides "-1" for `index_base::zero`, it also does `int` to `int64_t` casting and assigns `int_res_acc`(int) to `result_acc`(int64_t).

tests/unit_tests/blas/level1/iamax.cpp
Rbiessy: I think it would be more consistent with the rest of the tests in the blas domain if the tests in `RealSinglePrecisionBaseOne` were merged in `RealSinglePrecision` (and only one of the sizes with an index_base one would be enough IMO).

zhiweij1: Hi @Rbiessy, in 4c21de8, I still kept 2 cases for index_base::one. One's inc is positive (normal case), and another's is negative (we need to return zero. There is no need to +1 or -1).

Rbiessy (approved):

andrewtbarker (approved):

src/blas/backends/armpl/armpl_level1.cxx
```diff
@@ -561,22 +565,27 @@ DOTU_USM_LAUNCHER(std::complex<double>, ::cblas_zdotu_sub)
 
 template <typename T, typename CBLAS_FUNC>
 sycl::event iamin(sycl::queue& queue, int64_t n, const T* x, int64_t incx, int64_t* result,
-                  const std::vector<sycl::event>& dependencies, CBLAS_FUNC cblas_func) {
+                  const std::vector<sycl::event>& dependencies, oneapi::math::index_base base,
```
rafbiels, line 568: Hi @zhiweij1,
there's a bug on this line - the new argument was inserted after `dependencies` so the order is `dependencies, base` but the spec defines it should be before, i.e. `base, dependencies`. This function is correctly called with `base, dependencies` in L588 below so this doesn't compile:
```
/work/oneMath/src/blas/backends/armpl/armpl_level1.cxx:592:1: error: no matching function for call to 'iamin'                                                                                                      
  592 | IAMIN_USM_LAUNCHER(double, ::cblas_idamin)                                                                                                                                                                 
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                 
/work/oneMath/src/blas/backends/armpl/armpl_level1.cxx:588:16: note: expanded from macro 'IAMIN_USM_LAUNCHER'                                                                                                      
  588 |         return iamin(queue, n, x, incx, result, base, dependencies, ROUTINE);                      \                                                                                                       
      |                ^~~~~                                                                                                                                                                                       
/work/oneMath/src/blas/backends/armpl/armpl_level1.cxx:567:13: note: candidate function template not viable: no known conversion from 'oneapi::math::index_base' to 'const std::vector<sycl::event>' for 6th argument
  567 | sycl::event iamin(sycl::queue& queue, int64_t n, const T* x, int64_t incx, int64_t* result,
      |             ^
  568 |                   const std::vector<sycl::event>& dependencies, oneapi::math::index_base base,                                                                                                             
      |                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                             

```

Would you be able to submit a fix for this?

zhiweij1: Hi @rafbiels, I have submitted https://github.com/uxlfoundation/oneMath/pull/659, but I don't have an ARM env to test. Could you help check if that PR can fix your issue?

rafbiels: Thank you, I can confirm it compiles with the change

PR #644, opened 2025-03-05
[RNG] Added random engine adaptor: count_engine_adaptor

andreyfe1:
# Description

New engine adaptor is added. It wraps a random engine to add extra functionality: to query how many random numbers were taken from an engine during the generate call.

Spec changes: https://github.com/uxlfoundation/oneAPI-spec/pull/616

# Checklist

## All Submissions

## New features


include/oneapi/math/rng/device/engines.hpp
```diff
@@ -165,6 +165,73 @@ class mcg59 : detail::engine_base<mcg59<VecSize>> {
...
+
+    template <typename RealType>
+    RealType generate_single(RealType a, RealType b) {
+        counted_++;
+        return engine_.generate_single(a, b);
+    }
+
+    template <typename UIntType>
+    auto generate_uniform_bits() {
+        counted_ += Engine::vec_size;
```
iMartyan: Could you please check uniform bits engine function? I believe it may call internal generate 2 times depending on output type

include/oneapi/math/rng/device/engines.hpp
```diff
@@ -165,6 +165,73 @@ class mcg59 : detail::engine_base<mcg59<VecSize>> {
...
+        return engine_.template generate_uniform_bits<UIntType>();
+    }
+
+    template <typename UIntType>
+    auto generate_single_uniform_bits() {
+        counted_++;
```
iMartyan: same for several internal generates for 64 bit type

tests/unit_tests/rng/device/include/count_adaptor_test.hpp
```diff
@@ -0,0 +1,128 @@
...
+                auto res = oneapi::math::rng::device::generate(distr, engine);
+
+                Engine base_engine = engine.base();
+                res = oneapi::math::rng::device::generate(distr, base_engine);
+                if constexpr (Engine::vec_size == 1) {
+                    r[id] = res;
+                }
+                else {
+                    for (int j = 0; j < Engine::vec_size; ++j) {
+                        r[id * Engine::vec_size + j] = res[j];
```
iMartyan, line 79: `vec.store` here could be simpler

tests/unit_tests/rng/device/include/count_adaptor_test.hpp
```diff
@@ -0,0 +1,128 @@
...
+                    }
+                }
+
+                if (id == 42)
```
iMartyan: I suggest to use computed id number, not hardcoded just in case N_GEN would change

iMartyan (commented):
Thanks Andrey!
Just a general thought - do we really need to extend moments test with engine_adaptor as it doesn't change any computation? Also, correctness is checked in separate adaptor test against non-adaptored engine

include/oneapi/math/rng/device/engines.hpp
andreyfe1: Good catch! Done

include/oneapi/math/rng/device/engines.hpp
andreyfe1: Done

tests/unit_tests/rng/device/include/count_adaptor_test.hpp
andreyfe1: Done

tests/unit_tests/rng/device/include/count_adaptor_test.hpp
andreyfe1: Unfortunately, it's a bit painful since we work with USM pointers. So, it will look like
```cpp
res.store(id, sycl::address_space_cast<sycl::access::address_space::global_space, sycl::access::decorated::yes>(r));
```
And I'm not sure that all compilers support this combination. I'd keep it as-is

iMartyan: Agree

iMartyan (approved):
lgtm

aelizaro (approved):

PR #650, opened 2025-03-18
[blas][dft][generic] Fix macro for generic backend to enable generic device

s-Nick:
# Description

This PR fixes a bug with enablement macros for generic backend devices which vendor is not known to us yet.
It occurs after renaming to oneMath since these macros weren't renamed.

# Checklist

## All Submissions

    - [intel_igpu_generic_log.txt](https://github.com/user-attachments/files/19319216/intel_gpu_log.txt)
    -  [nvidia_generic_log.txt](https://github.com/user-attachments/files/19319222/nvidia_generic_log.txt)

## Bug fixes

      GitHub issue or in this PR)?
To test it properly you need a GPU that is neither Intel, NVIDIA or AMD

### DIFF
--- a/include/oneapi/math/detail/backends_table.hpp
+++ b/include/oneapi/math/detail/backends_table.hpp
@@ -95,7 +95,7 @@ static std::map<domain, std::map<device, std::vector<const char*>>> libraries = 
           } },
         { device::generic_device,
           {
-#ifdef ENABLE_GENERIC_BLAS_BACKEND
+#ifdef ONEMATH_ENABLE_GENERIC_BLAS_BACKEND
               LIB_NAME("blas_generic"),
 #endif
           } } } },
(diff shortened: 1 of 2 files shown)


Rbiessy (approved):

Rbiessy:
Nit: The PR title should also have the [dft] tag for the portFFT change. That's what will be used when merging I think.

Pinging @uxlfoundation/onemath-arch-write to help us get a second approval.

andreyfe1 (commented):
Does it make sense to make changes for CMake files or we keep it as-is, e.g. https://github.com/uxlfoundation/oneMath/blob/develop/CMakeLists.txt#L53 ?

s-Nick:
We added the `ONEMATH_` prefix to all our internal defined macros to comply with [ES.33 in C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-MACROS) as it was highlighted and requested in issue #554  , so I don't think the CMake definition should be changed.

Rbiessy:
Pinging @andreyfe1 in case you missed the answer above.

andreyfe1 (approved):

andreyfe1:
> Pinging @andreyfe1 in case you missed the answer above.

Thanks! I'm fine with the answer

PR #655, opened 2025-03-24
[BLAS][generic] Fix compilation with AdaptiveCpp

rafbiels:
# Description

The generic BLAS backend currently offers limited support for AdaptiveCpp, where:
* complex data type is not supported
* USM API is not supported

Add the required protections to make the generic BLAS backend compile and run correctly in the capacity it offers with AdaptiveCpp. That is, make the buffer API with real/integer data work fine. Throw the `unimplemented` exception for the unsupported features.

This change relies on the update of the generic blas backend to version 0.2.0 in: https://github.com/uxlfoundation/generic-sycl-components/pull/7

# Checklist

   * Tested on 12th gen Intel integrated GPU with SPIRV JIT compilation.
   * DPC++ before this PR: 1961 tests, 1800 skipped, 161 ran, 136 failed, 25 passed - [ut-dpcpp-igpu-spirv-ref.log.gz](https://github.com/user-attachments/files/19542325/ut-dpcpp-igpu-spirv-ref.log.gz)
   * DPC++ with this PR: 1961 tests, 1800 skipped, 161 ran, 136 failed, 25 passed - [ut-dpcpp-igpu-spirv.log.gz](https://github.com/user-attachments/files/19542328/ut-dpcpp-igpu-spirv.log.gz)
   * AdaptiveCpp before this PR: doesn't compile
   * AdaptiveCpp with this PR: 1961 tests, 1802 skipped, 159 ran, 117 failed, 42 passed - [ut-acpp-igpu.log.gz](https://github.com/user-attachments/files/19542331/ut-acpp-igpu.log.gz)
- [x] Have you added relevant regression tests? No AdaptiveCpp-specific tests are needed, existing tests provide enough coverage and should work with both DPC++ and AdaptiveCpp.
- [x] Have you included information on how to reproduce the issue (either in a GitHub issue or in this PR)? Compiled with
  `-DONEMATH_SYCL_IMPLEMENTATION=hipsycl` `-DENABLE_MKLCPU_BACKEND=False`
  `-DHIPSYCL_TARGETS='generic` `-DENABLE_MKLGPU_BACKEND=False`
  `-DENABLE_MKLCPU_BACKEND=False` `-DENABLE_GENERIC_BLAS_BACKEND=True`
  `-DBUILD_FUNCTIONAL_TESTS=True` `-DBUILD_EXAMPLES=True`


src/blas/backends/generic/generic_common.hpp
```diff
@@ -40,9 +40,11 @@ using handle_t = ::blas::SB_Handle;
 template <typename ElemT>
 using buffer_iterator_t = ::blas::BufferIterator<ElemT>;
 
+#ifdef BLAS_ENABLE_COMPLEX
```
Rbiessy: `BLAS_ENABLE_COMPLEX` is only a CMake variable set when compiling the generic blas component. A CMake variable is not set as a preprocessor definition so this will always disable complex.

To add a preprocessor definition please have a look at https://github.com/uxlfoundation/oneMath/blob/develop/src/CMakeLists.txt#L92 and https://github.com/uxlfoundation/oneMath/blob/develop/src/config.hpp.in. We want the preprocessor definition to start with the `ONEMATH_` prefix to avoid conflict, I would suggest to add one with the name `ONEMATH_GENERIC_BLAS_ENABLE_COMPLEX` that depends on the value of the CMake variable `GENERIC_BLAS_ENABLE_COMPLEX`. Then `GENERIC_BLAS_ENABLE_COMPLEX` could be set to true by default and set to false if `BLAS_ENABLE_COMPLEX` is found to be false (in the `find_package` path of generic blas). I think this is what you want to achieve?
It may look cumbersome but I think it's best to decouple the variable names from oneMath and from the generic blas component if possible.

Rbiessy (commented):
Thanks for the fix, I can see why this can be needed.
I don't get why do you say that no unit tests are available? Do you mean there is another issue with AdaptiveCpp preventing you to run the tests?

src/blas/backends/generic/generic_common.hpp
Rbiessy: There may be some logic needed to export the CMake variable `BLAS_ENABLE_COMPLEX ` in generic blas component so that it is visible in oneMath, I'm not sure.

rafbiels: It is already exported correctly with `target_compile_definitions(onemath_sycl_blas INTERFACE)`. If I use DPC++ and set `-DCMAKE_VERBOSE_MAKEFILE=ON` like this:
```
cmake -Bbuild-dpcpp -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$PWD/install-dpcpp -DCMAKE_CXX_COMPILER=icpx -DCMAKE_C_COMPILER=icx -DENABLE_MKLGPU_BACKEND=False -DENABLE_MKLCPU_BACKEND=False -DBUILD_FUNCTIONAL_TESTS=True -DBUILD_EXAMPLES=True -G Ninja -DCMAKE_VERBOSE_MAKEFILE=ON -DENABLE_GENERIC_BLAS_BACKEND=True -DONEMATH_SYCL_BLAS_DIR=$PWD/../generic-sycl-components/onemath/sycl/blas/install-dpcpp
```
I see both `-DBLAS_ENABLE_COMPLEX` and `-DSB_ENABLE_USM ` in the compilation commands, e.g.
```
/opt/intel/oneapi/compiler/2025.1/bin/icpx -DBLAS_ENABLE_COMPLEX -DSB_ENABLE_USM -I/home/rafal/work/OAS-670/oneMath/include -I/home/rafal/work/OAS-670/oneMath/build-dpcpp/bin -I/home/rafal/work/OAS-670/oneMath/src/include -I/home/rafal/work/OAS-670/oneMath/src -isystem /home/rafal/work/OAS-670/generic-sycl-components/onemath/sycl/blas/install-dpcpp/include -isystem /home/rafal/work/OAS-670/generic-sycl-components/onemath/sycl/blas/install-dpcpp/src -O3 -DNDEBUG -fPIC -fsycl -mllvm=-enable-global-offset=false -Wno-deprecated-declarations -Wno-deprecated-copy-with-user-provided-copy -MD -MT bin/blas/backends/generic/CMakeFiles/onemath_blas_generic_obj.dir/generic_level1_double.cpp.o -MF bin/blas/backends/generic/CMakeFiles/onemath_blas_generic_obj.dir/generic_level1_double.cpp.o.d -o bin/blas/backends/generic/CMakeFiles/onemath_blas_generic_obj.dir/generic_level1_double.cpp.o -c /home/rafal/work/OAS-670/oneMath/src/blas/backends/generic/generic_level1_double.cpp
```
so these code blocks are enabled when compiling with DPC++.

rafbiels:
> I don't get why do you say that no unit tests are available? Do you mean there is another issue with AdaptiveCpp preventing you to run the tests?

I missed I needed NETLIB to build the tests. I installed it now and ran the tests, however, with the tip of the `develop` branch (as reference) when compiling with DPC++ for NVIDIA GPU, I'm getting very poor results. Out of 1961 tests 1636 are skipped and 277 fail, so only 48 pass. Is this expected? I'm getting the same results with the changes in this PR and DPC++, so there is no regression. I can upload the output, but I must be doing something wrong, so should perhaps rerun in another configuration?

My commands were:
```
source /opt/intel/oneapi/setvars.sh
git clone https://github.com/uxlfoundation/oneMath.git
cd oneMath
cmake -Bbuild-dpcpp -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$PWD/install-dpcpp -DCMAKE_CXX_COMPILER=icpx -DCMAKE_C_COMPILER=icx -DENABLE_MKLGPU_BACKEND=False -DENABLE_MKLCPU_BACKEND=False -DBUILD_FUNCTIONAL_TESTS=True -DBUILD_EXAMPLES=True -G Ninja -DENABLE_GENERIC_BLAS_BACKEND=True -DGENERIC_BLAS_TUNING_TARGET=NVIDIA_GPU
cmake --build build-dpcpp
ctest --test-dir build-dpcpp --output-on-failure -j16
```

src/blas/backends/generic/generic_common.hpp
Rbiessy: I see. I think it can look confusing for future developers searching where and how these macros are enabled. It's not obvious they come from the generic blas component. I'd still prefer my suggested solution to clarify how the macros in generic blas impact oneMath and to align your changes with the existing oneMath defines.

Rbiessy:
The tests being marked as skipped is expected. One CTest runs multiple configurations and if one of them is not supported the test is marked as skipped. It's not great but changing that would be a large change I think.
The failures are not expected. The internal CI is working fine. This could have something to do with the CUDA version used, the CI uses 12.2. Or maybe with running the tests with `-j16`? (that would be bad). In case case for the scope of this PR I can run the CI once the changes are ready.
I'd suggest you attach a test log showing the complex tests are indeed skipped when disabled and if you have any tests failing with your configuration attach another one showing this patch does make more tests fail. You could also run the tests on an integrated GPU if that's easier.

src/blas/backends/generic/generic_common.hpp
rafbiels: I agree it might be confusing that they're set by the external project and happy to redefine them in oneMath. There will be changes needed for this in the external project first, because `SB_ENABLE_USM` is only a [preprocessor macro](https://github.com/uxlfoundation/generic-sycl-components/blob/f09710d77b607e9142d57aef6dafb463829c245e/onemath/sycl/blas/CMakeLists.txt#L82) and not a cmake variable in there, so oneMath cmake has no way to get its value (other than parsing cxx flags).

Which design would you prefer?
Option 1:
* oneMath calls find_package/FetchContent
* generic blas cmake sets the variables
* oneMath cmake reads the variables and sets its own variables based on them

Option 2:
* oneMath cmake sets the variables
* oneMath calls find_package/FetchContent
* generic blas cmake reads the variables and fails if the configuration is unsupported (e.g. complex enabled with acpp)

In my view, the second option where the parent project decides would be clearer. In the first (current) setup it's obfuscated to a oneMath developer where these values come from. On the other hand, second option requires duplicating the compiler support logic - first in oneMath to set the variables, then in generic blas to validate.

Rbiessy: I had the option 1 in mind when describing my suggested solution.
I think with option 2 it would not work when generic blas is included via find_package. generic sycl blas is already built at this point so oneMath setting a CMake variable would not change anything, unless I'm misunderstanding something.

rafbiels: You're right, the options are fixed in the `onemath_sycl_blas-config.cmake` file and there's no good way of exporting the logic into the config file. I'll work on option 1 then.

rafbiels: @Rbiessy option 1 requires exporting arbitrary target properties which requires CMake version >=3.19 as mentioned in:
https://cmake.org/cmake/help/latest/prop_tgt/EXPORT_PROPERTIES.html

Would you be happy increasing `cmake_minimum_required` to 3.19 for both projects?

The only other ways to make this work are:
* for kind-of option 1: parse `onemath_sycl_blas` target compile definitions and search for `BLAS_ENABLE_COMPLEX`
* for kind-of option 2: make generic blas project create extra header-only libraries with dummy headers and corresponding compile definitions, and use them as components, like:
```
find_package(ONEMATH_SYCL_BLAS COMPONENTS ENABLE_COMPLEX ENABLE_USM)`
...
target_link_libraries(${LIB_OBJ} PUBLIC ONEMATH::SYCL::SYCL onemath_sycl_blas ONEMATH_SYCL_BLAS::ENABLE_COMPLEX ONEMATH_SYCL_BLAS::ENABLE_USM)
```

My preference would be to require CMake >=3.19 and use exported target properties to inform the parent project of the complex/usm options set in the installed generic blas.

Rbiessy: Changing the minimum CMake version required has a big impact for users, I don't think this is a good enough reason unfortunately. The note you are referring to is about target properties (i.e. variables attached to a target). In our case we just need to read a global variable. We've had a similar case [here](https://github.com/ggml-org/llama.cpp/pull/12097/files#diff-123d14cf628b04694c9022d4210d80c14a29409095635150b944f8dfd1c70b37R39) where we read the value of `DNNL_GPU_VENDOR` coming from the DNNL package. I don't think oneDNN does anything special to export that variable but I could miss something.

Also a third option if this is too difficult would be to just add an option to oneMath to disable USM and complex for the generic blas backend. It's not ideal but easier.

rafbiels: > I don't think oneDNN does anything special to export that variable but I could miss something.

It has a package config file which explicitly exports that variable:
https://github.com/uxlfoundation/oneDNN/blob/762e317ce6b0dd72bb52efc9bbe072fa225b1ee7/cmake/config.cmake.in#L24

and uses `configure_package_config_file` from CMakePackageConfigHelpers to create the config. The generic BLAS project, on the other hand, simply uses `export(EXPORT ...)` to create the config file.

I can refactor the generic BLAS cmake to use a package config file exporting the variables, that seems to be a good solution - thanks for the pointer to oneDNN.

rafbiels: I updated https://github.com/uxlfoundation/generic-sycl-components/pull/7 with these changes

rafbiels:
@Rbiessy this is ready, but please rerun CI here after https://github.com/uxlfoundation/generic-sycl-components/pull/7 is merged as oneMath pulls the main branch. Before the merge, the variables are not exported so complex and USM support gets disabled in oneMath.

I uploaded new unit test outputs in the description which confirm there's no regression, but I still feel I'm doing something wrong on my system or the tests are unreliable as I'm getting very few tests passing even with Intel iGPU or CPU with the head of `develop` branch.

Rbiessy (approved):
Thanks @rafbiels for the work, this looks good to me! We need 2 approvals on this PR so can someone from @uxlfoundation/onemath-arch-write review this one? Then we can merge the generic-blas PR, re-run the CI here and ensure the CI passes before merging.

src/blas/backends/generic/CMakeLists.txt
```diff
@@ -161,13 +160,21 @@ if (NOT ONEMATH_SYCL_BLAS_FOUND)
...
+  )
   message(STATUS "Looking for onemath_sycl_blas - downloaded")
 
 else()
   message(STATUS "Looking for oneMath blas kernels - found")
   add_library(onemath_sycl_blas ALIAS ONEMATH_SYCL_BLAS::onemath_sycl_blas)
 endif()
 
+# Read cmake options exported by the onemath_sycl_blas project into oneMath variables
+set(ONEMATH_GENERIC_BLAS_ENABLE_COMPLEX ${BLAS_ENABLE_COMPLEX} CACHE INTERNAL  "Enable complex data support")
```
andrewtbarker:
```suggestion
set(ONEMATH_GENERIC_BLAS_ENABLE_COMPLEX ${BLAS_ENABLE_COMPLEX} CACHE INTERNAL  "Enable complex precision support")
```

andrewtbarker: or maybe "data" -> "data type"

andrewtbarker (approved):
Thansk for this fix! I have a very minor suggestion but it is not required for merging.

src/blas/backends/generic/CMakeLists.txt
rafbiels: Thanks, I opted for "complex data type" which is the same as the [backend docs](https://github.com/uxlfoundation/generic-sycl-components/blob/main/onemath/sycl/blas/README.md#cmake-options) call it

rafbiels:
CI failing due to https://github.com/uxlfoundation/oneMath/issues/658 - this needs fixing first

PR #656, opened 2025-03-24
[BLAS][generic] Support CMake option to provide local source dir

rafbiels:
Add cmake option `-DONEMATH_SYCL_BLAS_SOURCE_DIR=<path>` to provide a local checkout of the generic SYCL BLAS backend instead of always pulling from github. This facilitates local development of the generic SYCL BLAS code while testing within oneMath.

The search now works as follows (only the second step is new):
1. Try `find_package` to locate precompiled library and headers (as far as I'm aware no FindModule exists for onemath_sycl_blas).
2. If the option `ONEMATH_SYCL_BLAS_SOURCE_DIR` is used, pass the path to `add_subdirectory`.
3. Use FetchContent to check out the main branch from GitHub.

Additionally, do not set `BLAS_ENABLE_COMPLEX=ON` explicitly - this is already the default value.

### DIFF
--- a/docs/building_the_project_with_dpcpp.rst
+++ b/docs/building_the_project_with_dpcpp.rst
@@ -275,7 +275,10 @@ Building for oneMath generic SYCL BLAS
 --------------------------------------
 
 `onemath generic SYCL BLAS <https://github.com/uxlfoundation/generic-sycl-components/tree/main/onemath/sycl/blas>`_
-is enabled by setting ``-DENABLE_GENERIC_BLAS_BACKEND=True``.
+is enabled by setting ``-DENABLE_GENERIC_BLAS_BACKEND=True``. By default, the
+source code is downloaded automatically from GitHub during CMake configuration.
+To use a local checkout of the source code instead, use the
+``-DONEMATH_SYCL_BLAS_SOURCE_DIR=<path>`` option.
 
 By default, the generic BLAS backend is not tuned for any specific device.
 This tuning is required to achieve best performance.
--- a/src/blas/backends/generic/CMakeLists.txt
+++ b/src/blas/backends/generic/CMakeLists.txt
@@ -138,34 +138,44 @@ else()
   message(STATUS "generic BLAS is not tuned for any device which can impact performance")
 endif()
 
-# If find_package doesn't work, download onemath_sycl_blas from Github. This is
-# intended to make oneMath easier to use.
+# Try the following ways of finding onemath_sycl_blas:
+# 1. find_package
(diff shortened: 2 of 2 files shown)


docs/building_the_project_with_dpcpp.rst
```diff
@@ -275,7 +275,10 @@ Building for oneMath generic SYCL BLAS
 --------------------------------------
 
 `onemath generic SYCL BLAS <https://github.com/uxlfoundation/generic-sycl-components/tree/main/onemath/sycl/blas>`_
-is enabled by setting ``-DENABLE_GENERIC_BLAS_BACKEND=True``.
+is enabled by setting ``-DENABLE_GENERIC_BLAS_BACKEND=True``. By default, the
+source code is downloaded automatically from GitHub during CMake configuration.
+To use a local checkout of the source code instead, use the
+``-DONEMATH_SYCL_BLAS_SOURCE_DIR=<path>`` option.
```
Rbiessy, line 281: Could you detail why do you suggest this change? I'm curious to know in which scenario using `ONEMATH_SYCL_BLAS_SOURCE_DIR` is preferrable over compiling and installing the generic blas component and specifying it with `-DONEMATH_SYCL_BLAS_Dir`? That would allow CMake to find the package without any extra CMake logic needed here.

rafbiels: I actually missed how this was supposed to work as there were no instructions on how to do this either in oneMath or in onemath_sycl_blas.

I just tried now "building" and installing the onemath_sycl_blas project, and then pointed oneMath to the installation directory with `-DONEMATH_SYCL_BLAS_DIR`. This indeed worked as it exports .cmake configuration files.

I'm happy to close this PR and instead contribute an update to the instructions that explains this procedure, if that makes sense?

https://github.com/uxlfoundation/generic-sycl-components/tree/main/onemath/sycl/blas#Usage-in-oneMath says:
> The documentation on how to build and use it is provided in [oneMath repository](https://uxlfoundation.github.io/oneMath/building_the_project_with_dpcpp.html#building-for-onemath-generic-sycl-blas)

but I don't see where this is explained.

Rbiessy: Yes we missed that in the documentation. It's a feature of CMake [find_package](https://cmake.org/cmake/help/latest/command/find_package.html) itself but it is obviously obscure if you're not familiar  enough with CMake.
I would prefer to close this PR as you suggest if using `_DIR` works well enough for you, thanks!

PR #667, opened 2025-05-01
Lapack fixes

adegomme:
# Description

Recent armpl uses std::complex types, so if I'm not mistaken return value for the size here should be in real().


andrewtbarker (approved):

sknepper (approved):
Very nice, thanks for the fixes @adegomme !

PR #602, opened 2024-10-21
Rename oneMKL Interface to oneMath

Rbiessy:
# Description

Related RFC: https://github.com/oneapi-src/oneMKL/pull/564 and related spec PR: https://github.com/uxlfoundation/oneAPI-spec/pull/596
As a reminder the plan is to have this PR and the spec PR approved before we move this repository to https://github.com/uxlfoundation and merge the 2 PRs.

- The PR has the following implications on the MKLCPU and MKLGPU backends:
  - Had to introduce [`src/include/common_onemkl_conversion.hpp`](https://github.com/oneapi-src/oneMKL/pull/602/files#diff-1bd466e9a241e032faf9da384f46900bbc322e8258b086802c6f34d3a50ad381R36) to convert oneMath common types to oneMKL types. This wasn't needed before as the backends relied on the 2 projects to use the same namespace. The file also provide helper macros to catch oneMKL exceptions and rethrow them as oneMath exceptions.
  - The DFT MKLGPU backend used to catch and rethrow some but not all `oneapi::mkl` exceptions when committing a descriptor. We discussed this issue by email recently. With the changes needed by the renaming, we now catch and rethrow all the exceptions as `oneapi::math` exceptions and I was able to remove a related workaround in the tests (FYI @oneapi-src/onemkl-dft-write). Also see related discussion below: https://github.com/oneapi-src/oneMKL/pull/602#discussion_r1814235856
  - The RNG MKLGPU backend relied heavily on the 2 libraries using the same namespace. I introduced [`onemkl_distribution_conversion.hpp`](https://github.com/oneapi-src/oneMKL/pull/602/files#diff-6687d28034219fe472148c98659e2b98f391007028e2811f732b5823695cb33fR36) to convert the RNG types and call the backend's free functions such as `oneapi::mkl::rng::generate`, `oneapi::mkl::rng::skip_ahead`. @oneapi-src/onemkl-rng-write you may want to carefully review `src/rng/backends/mklgpu/` as I am not familiar with the RNG domain.
  - The renaming made it clear that some header files were duplicating function declarations for no reason in the BLAS and LAPACK domains. They are removed in https://github.com/oneapi-src/oneMKL/commit/baeb476f92b6dd7a0d137ab87c09fde1ec706634.
[...]

PR #668, opened 2025-05-02
Add CI testing for ArmPL backend

rafbiels:
Add CI testing for the ArmPL backend using the GitHub runner `ubuntu-24.04-arm`. Keep the CI strategy in line with the existing x86_64 pipeline splitting the jobs by domain.

There is no DPC++ binary release for aarch64 host, so the CI job builds it from source and caches the result. The build takes a lot of resources, but should almost always be loaded from the cache.

DPC++ tag `nightly-2025-04-01` is used instead of the latest release `v6.0.0` because the release is not compatible with the new-ish aarch64 glibc installed on the runner. This was fixed (presumably through upstream LLVM) later. We can move to the next open-source release once it becomes available.


adegomme:
lgtm, even if I can't formally approve it. thanks a lot !

.github/workflows/pr-arm.yml
```diff
@@ -0,0 +1,102 @@
...
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-24.04-arm
+    # One runner for each domain
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - config: oneMath BLAS
```
dnhsieh-intel:
```suggestion
        - config: ArmPL BLAS
```
I think we should also rename `oneMath BLAS` in `pr.yml` to `oneMKL BLAS` (accidentally changed in PR #602).

dnhsieh-intel (commented):
Thanks for adding new testing!

Does it make sense to also rename `pr.yml` to, for example, `pr-x86.yml`?

Not the scope of this PR, can be done later: We might want to separate out the clang-format check to something like "PR Linters". I personally feel a little awkward that the clang-format check is under "x86-64".

.github/workflows/pr-arm.yml
```diff
@@ -0,0 +1,102 @@
...
+    - name: Restore DPC++ from cache
+      id: cache-dpcpp
+      uses: actions/cache@d4323d4df104b026a6aa633fdb11d772146be0bf # v4.2.2
+      with:
+        path: dpcpp/install
+        key: dpcpp-${{ env.DPCPP_VERSION }}-arm
+    - name: Build DPC++
+      if: steps.domain_check.outputs.result == 'true' && steps.cache-dpcpp.outputs.cache-hit != 'true'
+      run: |
+        wget -P dpcpp https://github.com/intel/llvm/archive/refs/tags/${{ env.DPCPP_VERSION }}.tar.gz
```
dnhsieh-intel:
```suggestion
        wget --progress=dot:giga -P dpcpp https://github.com/intel/llvm/archive/refs/tags/${{ env.DPCPP_VERSION }}.tar.gz
```

.github/workflows/pr-arm.yml
```diff
@@ -0,0 +1,102 @@
...
+        tar -xzf ${{ env.DPCPP_VERSION }}.tar.gz
+        python llvm-${{ env.DPCPP_VERSION }}/buildbot/configure.py -o build -t Release --cmake-gen Ninja --native_cpu --cmake-opt="-DSYCL_ENABLE_BACKENDS=native_cpu" --cmake-opt="-DCMAKE_INSTALL_PREFIX=$PWD/install" --llvm-external-projects=openmp
+        cd build
+        ninja deploy-sycl-toolchain omp install
+    - name: Install ArmPL
+      if: steps.domain_check.outputs.result == 'true'
+      run: |
+        mkdir armpl
+        cd armpl
+        wget https://developer.arm.com/-/cdn-downloads/permalink/Arm-Performance-Libraries/Version_${{ env.ARMPL_VERSION }}/arm-performance-libraries_${{ env.ARMPL_VERSION }}_deb_gcc.tar
```
dnhsieh-intel:
```suggestion
        wget --progress=dot:giga https://developer.arm.com/-/cdn-downloads/permalink/Arm-Performance-Libraries/Version_${{ env.ARMPL_VERSION }}/arm-performance-libraries_${{ env.ARMPL_VERSION }}_deb_gcc.tar
```

.github/workflows/pr-arm.yml
```diff
@@ -0,0 +1,102 @@
...
+      with:
+        script: |
+         const domainCheck = require('.github/scripts/domain-check.js')
+         return domainCheck({github, context, domain: "${{ matrix.domain }}"})
+    - name: Restore netlib from cache
+      id: cache-lapack
+      uses: actions/cache@d4323d4df104b026a6aa633fdb11d772146be0bf # v4.2.2
+      with:
+        path: lapack/install
+        key: lapack-${{ env.LAPACK_VERSION }}-arm
```
dnhsieh-intel, line 49: Have you checked if the netlib and compiler caches are working as expected?

rafbiels: Yes, I also just pushed an insignificant change to this PR to re-check this (https://github.com/uxlfoundation/oneMath/pull/668/commits/8f0bac66a43eba34771391e0039263c3e7e17168). Indeed, the previous job built netlib and the compiler:
https://github.com/uxlfoundation/oneMath/actions/runs/14968116030/job/42042637818
and the new job only loaded them from cache:
https://github.com/uxlfoundation/oneMath/actions/runs/14970915727/job/42051453141

rafbiels:
Many thanks for the review  @dnhsieh-intel, these are great suggestions! I applied all of them in https://github.com/uxlfoundation/oneMath/pull/668/commits/131b4e06c8adf7c5a4eba29fd0cd74ec5dd245ce

dnhsieh-intel (approved):
Thank you!

One last thing: I believe `px-x86.yml` is a typo. 😉

rafbiels:
Fixed!

Rbiessy (approved):

PR #669, opened 2025-05-07
[BLAS] SYCL-Graph integration for native-command

EwanC:
In order to support applications calling the library with a sycl queue recording to a SYCL-Graph, check if the `ext_codeplay_enqueue_native_command` command-group is being recorded to a graph object. If so use the native stream recording APIs to add the blas calls as nodes in the graph.

In particular this fixes the llama.cpp `MUL_MAT` unit tests on CUDA with SYCL-Graph enabled. Previously this would throw an error:

```sh
$ GGML_SYCL_DISABLE_GRAPH=0 ./bin/test-backend-ops -b SYCL0 -o MUL_MAT -p type_a=f16,type_b=f32,m=16,n=1,k=256,bs=\\[1,1\\],nr=\\[2

UR CUDA ERROR:
        Value:           700
        Name:            CUDA_ERROR_ILLEGAL_ADDRESS
        Description:     an illegal memory access was encountered
        Function:        operator()
        Source Location: $HOME/dpcpp/unified-runtime/source/adapters/cuda/queue.cpp:154

Native API failed. Native API returns: 2147483646 (UR_RESULT_ERROR_UNKNOWN)
Exception caught at file:$HOME/llama.cpp/ggml/src/ggml-sycl/ggml-sycl.cpp, line:3598, func:operator()
SYCL error: CHECK_TRY_ERROR((stream)->wait()): Meet error in this line code!
  in function ggml_backend_sycl_synchronize at $HOME/llama.cpp/ggml/src/ggml-sycl/ggml-sycl.cpp:3598
$HOME/llama.cpp/ggml/src/ggml-sycl/../ggml-sycl/common.hpp:118: SYCL error
Could not attach to process.  If your uid matches the uid of the target
process, check the setting of /proc/sys/kernel/yama/ptrace_scope, or try
again as the root user.  For more details, see /etc/sysctl.d/10-ptrace.conf
ptrace: Operation not permitted.
No stack.
The program is not being run.
```

With this patch we can successfully record the blas native-command command-groups used as part of the llama.cpp oneMath calls. In particular USM `gemm` and `gemm_batch` which I've extended the tests to verify the correctness of.


src/blas/backends/cublas/cublas_batch.cpp
```diff
@@ -722,8 +722,9 @@ inline sycl::event gemm_batch_usm_impl(sycl::queue& queue, transpose* transa, tr
             auto handle = sc.get_handle(queue);
             int64_t offset = 0;
             cublasStatus_t err;
-            for (int64_t i = 0; i < group_count; i++) {
 #ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
+            sc.begin_recording_if_graph(queue);
```
Rbiessy: Why is this added only for `gemm_batch`? I understand llama only uses oneMath for batch gemm (if oneDNN is found) but from oneMath perspective it would look odd. I think this could be supported for all operations by calling `begin_recording_if_graph` and `end_recording_if_graph` in https://github.com/uxlfoundation/oneMath/blob/1694befacc9f251117fa8dc70c901086834bd160/src/blas/backends/cublas/cublas_task.hpp#L57

Also note that there is a llama PR to start using oneDNN for batch gemm too which could be another way to avoid the issue? See https://github.com/ggml-org/llama.cpp/pull/12972.

Rbiessy (commented):
To answer your questions:
1. Yes we need to ensure this can build with the latest public oneAPI release
2. Yes it would be best to have a test with SYCL-Graph if this is something we want to support. I'm thinking it could be a separate test file. I don't think we would need to test every operation, some sort of example using a single oneMath operation could be enough?

Let me know if you think you will still need this oneMath change. The llama PR using oneDNN looks promising and if it works well we could remove the dependency on oneMath.

src/blas/backends/cublas/cublas_batch.cpp
EwanC: Thanks for the feedback, I've reworked this patch so it part of that generic helper. I discovered that I would need to record the `gemm` operations as well as `gemm_batch` to get the llama.cpp MUL_MAT tests working, so this definitely helped with that.

I extended the existing tests to have a graph variant rather than creating a separate SYCL-Graph test file. Partly because i'm not familiar enough with BLAS to be able to write tests from scratch effectively, and also because having the graph tests fully integrated gives me confidence to say to stakeholders that we support recording gemm/gemm_batch operations to a graph. If you have reservations about this test setup we can revisit it though. 

Note the last thing blocking me taking this PR out of draft now is https://github.com/uxlfoundation/oneMath/pull/681 which is required for the new graph gemm tests I added to pass. Edit: That's now merged, so this PR is unblocked.

Rbiessy: > I extended the existing tests to have a graph variant rather than creating a separate SYCL-Graph test file. Partly because i'm not familiar enough with BLAS to be able to write tests from scratch effectively, and also because having the graph tests fully integrated gives me confidence to say to stakeholders that we support recording gemm/gemm_batch operations to a graph. If you have reservations about this test setup we can revisit it though.

Originally the PR was only affecting `gemm` but we made the change more generic (for the best IMO). I think it's odd that SYCL-Graph is only tested in the gemm and gemm_batch test files. I think a separate test file would give you more freedom to what we want to test with SYCL-Graph and how. For instance I don't think it is needed to duplicate the SYCL-Graph tests for all data types and for different configurations of gemm. I'm worried this increases the testing time more than necessary. I can help a bit more with how to set this up if you want.

tests/unit_tests/blas/sycl-graph/gemm_batch_usm.cpp
EwanC, line 1: @Rbiessy I've added some tests in a new graph directory that create a `float` linear multi-node graph, and execute it.

I think `gemm_usm.cpp` is decent, the test creates a 4 node graph with the possible transpose/non-transpose combinations of matrix A & B. I'm less confident about how i've done `gemm_batch_usm.cpp`, it's similar to the original which is quite verbose. So if you've ideas on how to reduce that further it would be good to hear.

tests/unit_tests/blas/sycl-graph/gemm_usm.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include "test_helper.hpp"
+
+#include <gtest/gtest.h>
+
+using namespace sycl;
+
+extern std::vector<sycl::device*> devices;
+
+namespace {
+#if defined(CALL_RT_API) && defined(SYCL_EXT_ONEAPI_GRAPH)
```
Rbiessy: I think it's best to also enable the 2 tests for the compiler-time dispatcher (when `CALL_RT_API` is not defined), just to make it clear it works fine with SYCL-Graph.

tests/unit_tests/blas/sycl-graph/gemm_batch_usm.cpp
Rbiessy: It looks fine to me. Sure we could introduce more helper functions, I don't mind either way.

EwanC: cool, I'll just keep it as-is then

tests/unit_tests/blas/sycl-graph/gemm_usm.cpp
EwanC: sure, I wasn't sure if the compile-time test path added extra testing value, but added it back now.

Rbiessy (approved):

EwanC:
@andrewtbarker Would you be able to do the other review on this PR? Picking on you since you've reviewed my other PRs, but if there are other codeowners that could do the review then feel free to tag them instead.

src/blas/backends/cublas/cublas_scope_handle.cpp
```diff
@@ -32,36 +32,80 @@ namespace cublas {
...
+        return;
+    }
+
+    CUresult err;
+#if CUDA_VERSION >= 12030
+    // After CUDA 12.3 we can use cuStreamBeginCaptureToGraph to capture
+    // the stream directly in the native graph, rather than needing to
+    // instantiate the stream capture as a new graph.
+    auto graph = ih.ext_codeplay_get_native_graph<sycl::backend::ext_oneapi_cuda>();
+    CUDA_ERROR_FUNC(cuStreamBeginCaptureToGraph, err, streamId, graph, nullptr, nullptr, 0,
```
andrewtbarker, line 76: I'm a bit confused about if this runs when we are not doing graph recording. I don't see any guard/check that keeps it from running, so if we haven't called `graph.begin_recording(main_queue);` does this just do nothing?

andrewtbarker (approved):
This generally looks good, thanks especially for the nice comments. I have one question for my own understanding but it doesn't block merging.

src/blas/backends/cublas/cublas_scope_handle.cpp
EwanC: good question, this is the purpose of the `ih.ext_codeplay_has_graph()` check on line 66. It will return true if the queue is recording to a graph and false otherwise. So if `graph.begin_recording(main_queue);` hasn't been called `CublasScopedContextHandler::begin_recording_if_graph()`  will early return there.

Rbiessy:
> /intelci: run

I think the internal CI is down for some time. @EwanC would you be able to attach a log of the tests running on Nvidia HW? That's the best alternative.

src/blas/backends/cublas/cublas_scope_handle.cpp
andrewtbarker: Thanks! The name `ext_codeplay_has_graph()` initially seems to me it should mean something slightly different, but this also makes sense.

EwanC:
> I think the internal CI is down for some time. @EwanC would you be able to attach a log of the tests running on Nvidia HW? That's the best alternative.

Unfortunately the results I see locally are non-deterministic and I see fails (which vary in number on each run) on the develop branch and this feature branch. I've inserted the logs below however.

<details>
<summary> Tip DPC++ with CUDA 12.8 used for below testing</summary>

```sh

$ git log | head
commit 4b3ed9b13632bc31eb35c34a3ef37e17c80ffe05
Author: Wu Yingcong <yingcong.wu@intel.com>
Date:   Thu Jun 12 16:21:08 2025 +0800

    [DeviceSanitizer] Remove not needed test fix. (#18946)

    The fix is no longer need and not working at the moment since it does
    not come with `%{run}`.
	
$ ./bin/sycl-ls
[cuda:gpu][cuda:0] NVIDIA CUDA BACKEND, NVIDIA GeForce GT 1030 6.1 [CUDA 12.8]	
```
</details>

**Logs** 
[develop_test_main_blas_ct.txt](https://github.com/user-attachments/files/20707122/develop_test_main_blas_ct.txt) - 19 FAILED TESTS
[develop_test_main_blas_rt.txt](https://github.com/user-attachments/files/20707123/develop_test_main_blas_rt.txt) - 12 FAILED TESTS
[native-command_test_main_blas_ct.txt](https://github.com/user-attachments/files/20707124/native-command_test_main_blas_ct.txt) - 12 FAILED TESTS
[native-command_test_main_blas_rt.txt](https://github.com/user-attachments/files/20707125/native-command_test_main_blas_rt.txt) - 8 FAILED TESTS

Rbiessy:
Thanks for sharing. Yeah I'm not sure why you see inconsistent tests failing. It may have to do with this device that we're not using for tests usually.
I think this is enough to show this PR should not introduce more failures so merging now.

PR #671, opened 2025-05-08
Fix deprecation warnings for oneMKL >= 2025.0

uphoffc:
Compiling oneMath with oneMKL >= 2025.0 gives a lot of annoying deprecation warnings at compile time and at run time. This PR fixes the deprecation warnings and includes backwards compability switches for oneAPI 2024.x.

### DIFF
--- a/src/dft/backends/mklgpu/commit.cpp
+++ b/src/dft/backends/mklgpu/commit.cpp
@@ -40,8 +40,20 @@
 #include <mkl_version.h>
 #if INTEL_MKL_VERSION < 20250000
 #include <mkl/dfti.hpp>
+namespace oneapi::math::dft::mklgpu::detail {
+template <typename D, typename C, typename V>
+void set_vector_value(D& desc, C p, V const& vec) {
+    desc.set_value(p, vec.data());
+}
+}
 #else
 #include <mkl/dft.hpp>
+namespace oneapi::math::dft::mklgpu::detail {
+template <typename D, typename C, typename V>
+void set_vector_value(D& desc, C p, V const& vec) {
+    desc.set_value(p, vec);
+}
+}
 #endif
 
 // Intel oneMKL 2024.1 deprecates input/output strides.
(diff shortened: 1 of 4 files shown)


src/dft/backends/mklgpu/commit.cpp
```diff
@@ -159,14 +171,18 @@ class mklgpu_commit final : public dft::detail::commit_impl<prec, dom> {
         desc.set_value(backend_param::FORWARD_SCALE, config.fwd_scale);
         desc.set_value(backend_param::BACKWARD_SCALE, config.bwd_scale);
         desc.set_value(backend_param::NUMBER_OF_TRANSFORMS, config.number_of_transforms);
-        desc.set_value(backend_param::COMPLEX_STORAGE,
-                       to_mklgpu<onemath_param::COMPLEX_STORAGE>(config.complex_storage));
+        if (config.complex_storage != dft::detail::config_value::COMPLEX_COMPLEX) {
+            throw math::unimplemented("dft/backends/mklgpu", "commit",
+                                      "MKLGPU only supports complex-complex complex storage.");
```
lhuot, line 176: I believe config_value::REAL_REAL is still a valid option if config_param::COMPLEX_STORAGE is used. Is it not working?

src/dft/backends/mklgpu/mklgpu_helpers.hpp
```diff
@@ -82,95 +90,66 @@ inline constexpr oneapi::mkl::dft::config_param to_mklgpu(dft::detail::config_pa
                                          "Invalid config param.");
             return static_cast<oparam>(0);
     }
-}
+}*/
```
lhuot: Should it be removed instead of commented?

lhuot:
Thanks @uphoffc for this cleanup! Could you please attach log of successful build and tests? Thanks!

uphoffc:
Log file: [unit_tests.txt](https://github.com/user-attachments/files/20119060/unit_tests.txt)

src/dft/backends/mklgpu/mklgpu_helpers.hpp
uphoffc: Up to you. It's in the history, so it can always be restored.

src/dft/backends/mklgpu/commit.cpp
uphoffc: In the [documentation](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2025-1/configuring-data-layouts.html) it says

> Complex descriptor with config_value::REAL_REAL set for config_param::COMPLEX_STORAGE
> **(not implemented via the DPC++ interface of oneMKL)**

Moreover, the old to_mklgpu function was as following:

> template <>
> inline constexpr int to_mklgpu<dft::detail::config_param::COMPLEX_STORAGE>(
>     dft::detail::config_value value) {
>     if (value == dft::detail::config_value::COMPLEX_COMPLEX) {
>         return DFTI_COMPLEX_COMPLEX;
>     }
>     else {
>         throw math::unimplemented("dft", "MKLGPU descriptor set_value()",
>                                   "MKLGPU only supports complex-complex for complex storage.");
>         return 0;
>     }
> }

I concluded that REAL_REAL is effectively not supported even though one might set the option.

uphoffc:
Build logs: 
[build_2024_2.txt](https://github.com/user-attachments/files/20119260/build_2024_2.txt)
[build_2025_1.txt](https://github.com/user-attachments/files/20119262/build_2025_1.txt)
[build_2025_0.txt](https://github.com/user-attachments/files/20119263/build_2025_0.txt)

src/dft/backends/mklgpu/commit.cpp
lhuot: I see thanks! I was looking at another place in the doc and missed that note. I guess this highlighted an opportunity to make the Intel oneMKL doc clearer :-) Thanks!

src/dft/backends/mklgpu/mklgpu_helpers.hpp
lhuot: I think it is best to remove.

lhuot (approved):
Changes look good to me and precommit as well (failed steps unrelated to these changes). Thanks for the patch @uphoffc!

Rbiessy (approved):
LGTM!

PR #673, opened 2025-05-13
[RFC] Introduce backend maintainer roles

sknepper:
# Description

Link to rendered README: https://github.com/sknepper/oneMath/blob/rfcs_archowners/rfcs/20250509-backend-maintainers/README.md

The RFC proposes to introduce Backend maintainer roles.

Related PR updating the documentation per this RFC: #674

PR #674, opened 2025-05-13
[Docs] Introduce backend maintainer roles

sknepper:
# Description

Related RFC: #673 

Updates the CONTRIBUTING.md and MAINTAINERS.md files to introduce the new backend maintainer roles as described in the corresponding RFC, as well as CODEOWNERS file. These roles were created based on the approved RFC:

@uxlfoundation/onemath-cpu-aarch64
@uxlfoundation/onemath-cpu-x64
@uxlfoundation/onemath-gpu-amd
@uxlfoundation/onemath-gpu-intel
@uxlfoundation/onemath-gpu-nvidia

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -37,6 +37,13 @@
 /include/oneapi/math/blas.hxx @uxlfoundation/onemath-blas-write
 /include/oneapi/math/bfloat16.hpp @uxlfoundation/onemath-blas-write
 /include/oneapi/mkl/blas.hpp @uxlfoundation/onemath-blas-write
+/src/blas/backends/armpl/ @uxlfoundation/onemath-cpu-aarch64 @uxlfoundation/onemath-blas-write
+/src/blas/backends/cublas/ @uxlfoundation/onemath-gpu-nvidia @uxlfoundation/onemath-blas-write
+/src/blas/backends/mkl_common/ @uxlfoundation/onemath-cpu-x64 @uxlfoundation/onemath-gpu-intel @uxlfoundation/onemath-blas-write
+/src/blas/backends/mklcpu/ @uxlfoundation/onemath-cpu-x64 @uxlfoundation/onemath-blas-write
+/src/blas/backends/mklgpu/ @uxlfoundation/onemath-gpu-intel @uxlfoundation/onemath-blas-write
+/src/blas/backends/netlib/ @uxlfoundation/onemath-cpu-x64 @uxlfoundation/onemath-blas-write
+/src/blas/backends/rocblas/ @uxlfoundation/onemath-gpu-amd @uxlfoundation/onemath-blas-write
 /src/blas/ @uxlfoundation/onemath-blas-write
 /tests/unit_tests/blas/ @uxlfoundation/onemath-blas-write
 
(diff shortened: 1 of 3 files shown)


MAINTAINERS.md
```diff
@@ -79,6 +80,53 @@ The process of becoming a Domain maintainer is:
...
+    bug fixes, implementing features, and answering backend specific questions
+    in [#onemath](https://uxlfoundation.slack.com/archives/onemath) Slack channel.
+  * Review pull requests pertaining to their specific backend.
+  * Monitor testing results and flag issues in their specific areas of
+    responsibility.
+  * Support and guide Contributors.
+
+Requirements:
+  * Experience as Contributor in the specific backend for at least 6 months.
+  * Commit at least 25% of working time to the project.
```
rodburns: Should we include this requirement, or just use the one below that says "Commits to being responsible for that specific backend" ?

Rbiessy (commented):
Is the plan to also update [CODEOWNERS](https://github.com/uxlfoundation/oneMath/blob/develop/.github/CODEOWNERS) to automatically add backend maintainers to PRs review?
I think ideally we would want to ping both backend maintainers and domain owners? I think this could be done if we add patterns such as:
```
/src/blas/backends/cublas/ @uxlfoundation/onemath-gpu-nvidia @uxlfoundation/onemath-blas-write
```
and similar for other backends and domains.

MAINTAINERS.md
sknepper: Good point! Removed the 25% commitment as likely there would not be as much work required for backend maintainers.

sknepper:
> Is the plan to also update [CODEOWNERS](https://github.com/uxlfoundation/oneMath/blob/develop/.github/CODEOWNERS) to automatically add backend maintainers to PRs review? I think ideally we would want to ping both backend maintainers and domain owners? I think this could be done if we add patterns such as:
> 
> ```
> /src/blas/backends/cublas/ @uxlfoundation/onemath-gpu-nvidia @uxlfoundation/onemath-blas-write
> ```
> 
> and similar for other backends and domains.

Good point! Yes, this absolutely should be done. I've made the updates to CODEOWNERS (and also updated the RFC to reflect this). Thanks!

Rbiessy (approved):
Nice, LGTM!

.github/CODEOWNERS
```diff
@@ -37,6 +37,13 @@
 /include/oneapi/math/blas.hxx @uxlfoundation/onemath-blas-write
 /include/oneapi/math/bfloat16.hpp @uxlfoundation/onemath-blas-write
 /include/oneapi/mkl/blas.hpp @uxlfoundation/onemath-blas-write
+/src/blas/backends/armpl/ @uxlfoundation/onemath-cpu-aarch64 @uxlfoundation/onemath-blas-write
```
mkrainiuk, line 40: I see error in github web view in this file, like 
`Unknown owner on line 40: make sure the team @uxlfoundation/onemath-cpu-aarch64 exists, is publicly visible, and has write access to the repository`

Could it be that the teams are not yet created? I guess we need them before updating CODEOWNERS

sknepper: Nice find! I did wonder what would happen since these haven't been created yet...didn't realize GitHub was that smart in their web view. Thanks!! =)
If the RFC is approved, then I can request these GitHub teams to be created, and we can wait to merge this PR (with the documentation and codeowner changes) until that happens. Alternatively, I can separate out the codeowners changes to another PR.  Unless there would be a long delay with creating the GitHub teams, my preference would be the former.

mkrainiuk: It makes sense to merge all together after RFC is approved and new teams are created.

mkrainiuk: BTW the RFC is already approved, isn't it? Or are we waiting for more feedback?

sknepper: > BTW the RFC is already approved, isn't it? Or are we waiting for more feedback?

I had put a decision date of May 31 for the RFC. So far only positive responses have been received. =)

sknepper:
The new teams were created, made publicly visible, and write access to the oneMath repository was granted. The errors in the github web view of codeowners are no longer there.
I'll merge the approved RFC (#673) and once this PR gets a second approver, it can also be merged.

dnhsieh-intel (approved):

PR #678, opened 2025-05-16
[BLAS] Remove queue from `host_task_internal` API

EwanC:
A `sycl::queue` object is passed to the `host_task_internal` function as a parameter and used when interacting with a object `CublasScopedContextHandler`. However, the `interop_handle` parameter that is also passed to this entry-point and `CublasScopedContextHandler` constructor is already able to get this information via `interop_handle::get_native_queue()`.

This patch removes the superfluous queue parameters and changes the `CublasScopedContextHandler:get_stream()` implementation to use the `interop_handle` method instead. The unused `get_context()` and `wait_stream` methods are also removed from `CublasScopedContextHandler`

### DIFF
--- a/src/blas/backends/cublas/cublas_task.hpp
+++ b/src/blas/backends/cublas/cublas_task.hpp
@@ -46,19 +46,19 @@ namespace cublas {
 
 #ifdef __HIPSYCL__
 template <typename H, typename F>
-static inline void host_task_internal(H& cgh, sycl::queue queue, F f) {
-    cgh.hipSYCL_enqueue_custom_operation([f, queue](sycl::interop_handle ih) {
-        auto sc = CublasScopedContextHandler(queue, ih);
+static inline void host_task_internal(H& cgh, F f) {
+    cgh.hipSYCL_enqueue_custom_operation([f](sycl::interop_handle ih) {
+        auto sc = CublasScopedContextHandler(ih);
         f(sc);
     });
 }
 #else
 template <typename H, typename F>
-static inline void host_task_internal(H& cgh, sycl::queue queue, F f) {
+static inline void host_task_internal(H& cgh, F f) {
 #ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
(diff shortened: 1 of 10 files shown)


Rbiessy (approved):
LGTM, thanks!

src/blas/backends/cublas/cublas_task.hpp
```diff
@@ -46,28 +46,28 @@ namespace cublas {
 
 #ifdef __HIPSYCL__
 template <typename H, typename F>
-static inline void host_task_internal(H& cgh, sycl::queue queue, F f) {
-    cgh.hipSYCL_enqueue_custom_operation([f, queue](sycl::interop_handle ih) {
-        auto sc = CublasScopedContextHandler(queue, ih);
+static inline void host_task_internal(H& cgh, F f) {
+    cgh.hipSYCL_enqueue_custom_operation([f](sycl::interop_handle ih) {
+        auto sc = CublasScopedContextHandler(ih);
```
rafbiels, line 51: This will not compile with AdaptiveCpp as the `CublasScopedContextHandler` here is the one from `cublas_scope_handle_hipsycl.hpp` (see the conditional include at the top). Could you apply the same modifications in `cublas_scope_handle_hipsycl.[ch]pp` as in the DPC++ version?

rafbiels (changes requested):
Nice, I like this change! Could you do the same for the AdaptiveCpp code path? I can help with testing if needed, let me know.

src/blas/backends/cublas/cublas_task.hpp
EwanC: Thanks for spotting! I've made those changes. I see the same pass rate running `./bin/test_main_blas_rt` with an AdaptiveCpp build (using omp) before and after these oneMath tests. I'm not that familiar with this setup though, so if you have an established testing setup for AdaptiveCpp then please feel free to double check the patch.

rafbiels: Thanks! I think that's sufficient validation for these changes

rafbiels (approved):

andrewtbarker (approved):
This is a nice simplification, thanks for the changes.

PR #679, opened 2025-05-22
Fix preprocessor macros protecting DPC++ bfloat16 extension

### LINKED ISSUE #677: Error in types.hpp
# Summary
Build fail due to an error in `types.hpp`, where it references a `sycl::ext::oneapi::bfloat16` type that may not be defined or included properly and also adaptivecpp does not have ext namespace.

# Version
Report oneMath version and githash.
If it is a regression, report githash for the last known good revision.

# Environment
Compiler : clang / clang++ - 18.1.8
OS name and version - Centos 8
ARMPL backend - enabled (v_24.10, build using gcc)

# Steps to reproduce
1. Clone the oneMath repository and configure CMake with the following flags:
cmake .. -DONEMATH_SYCL_IMPLEMENTATION=hipsycl -DENABLE_MKLGPU_BACKEND=False -DENABLE_MKLCPU_BACKEND=False -DENABLE_ARMPL_BACKEND=True -DENABLE_MKLCPU_THREAD_TBB=False -DHIPSYCL_TARGETS=omp\;generic -DBUILD_FUNCTIONAL_TESTS=False -DBUILD_EXAMPLES=False -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang

cmake –build .

# Observed behavior
2385 |   template <typename T = dataT, std::enable_if<std::is_same_v<T, dataT> && !std::is_const_v<T>, bool> = false>
      |                                                                                                         ^~~~~
In file included from /path/to/oneMath/include/oneapi/math/lapack/detail/lapack_loader.hpp:31:
/path/to/test/oneMath/include/oneapi/math/types.hpp:37:24: error: ‘ext’ in namespace ‘sycl’ does not name a type
   37 | using bfloat16 = sycl::ext::oneapi::bfloat16;
      |                        ^~~

# Expected behavior
OneMath must build successfully

rafbiels:
Use the [SYCL2020-standard](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_requirements_for_an_extension) macros `SYCL_IMPLEMENTATION_<vendor>` to decide whether the extension is available or not.

Previous version assumed that if the compiler is not AdaptiveCpp, then it is DPC++. Although these are the only currently supported compilers, such assumptions block the development of support for other SYCL implementations. There is also no good reason to block non-SYCL compilers from parsing the oneMath headers in such a way.

Ideally, the use of an extension should be protected by a specific feature-test macro, but this is currently missing for the DPC++ bfloat16 extension (https://github.com/intel/llvm/issues/18638).

Note DPC++ provides two vendor identifiers `SYCL_IMPLEMENTATION_ONEAPI` and `SYCL_IMPLEMENTATION_INTEL`. The latter is used here to avoid confusion with the name "oneAPI" referring to "oneAPI Math Library" which could mean an internal implementation of a feature in this project.

Fixes #677

### DIFF
--- a/include/oneapi/math/types.hpp
+++ b/include/oneapi/math/types.hpp
@@ -20,20 +20,20 @@
 #ifndef _ONEMATH_TYPES_HPP_
 #define _ONEMATH_TYPES_HPP_
 
-#ifdef __HIPSYCL__
-#include "oneapi/math/bfloat16.hpp"
-#endif
-
 #if __has_include(<sycl/sycl.hpp>)
 #include <sycl/sycl.hpp>
 #else
 #include <CL/sycl.hpp>
 #endif
 
+#ifndef SYCL_IMPLEMENTATION_INTEL
+#include "oneapi/math/bfloat16.hpp"
+#endif
+
 namespace oneapi {
 namespace math {
 
-#ifndef __HIPSYCL__
+#ifdef SYCL_IMPLEMENTATION_INTEL
 using bfloat16 = sycl::ext::oneapi::bfloat16;
 #endif
 


sknepper (approved):
Thanks for the fix and the detailed explanation, @rafbiels !

Rbiessy (approved):

sknepper:
The failures in the pre-commit check are unrelated to this PR - thanks again, @rafbiels !

PR #681, opened 2025-05-23
[BLAS] Avoid blocking wait inside native-command callable

EwanC:
The `CUBLAS_ERROR_FUNC_SYNC` macro calls `cuStreamSynchronize(currentStreamId);` to synchronize the stream associated with the cuBLAS handle after making the API call, which `CUBLAS_ERROR_FUNC` doesn't.

When using the [sycl_ext_codeplay_enqueue_native_command](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_codeplay_enqueue_native_command.asciidoc) extension is advised by the specification not to use commands that block

> The function object interopCallable is invoked to enqueue commands to a native queue or graph and therefore, APIs which block or synchronize could prolong or interfere with other commands being enqueued to the backend.

As these will run when the command is submitted before the command-group dependencies have been satisfied.

Instead use the sync variant for the host-task path rather than native enqueue path, which aligns with how
[cublas_batch.cpp](https://github.com/uxlfoundation/oneMath/blob/f09d6cbe2aed9ffef47a6b36b7598322e7cb3104/src/blas/backends/cublas/cublas_batch.cpp#L170) is using a sync variant on the host-task path and not on the native enqueue path.

```cpp
#ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
            CUBLAS_ERROR_FUNC_T("cublasGemmStridedBatchedEx", cublasGemmStridedBatchedEx, err,
                                handle, get_cublas_operation(transa), get_cublas_operation(transb),
                                m, n, k, &alpha, a_, get_cublas_datatype<cuTypeA>(), lda, stride_a,
                                b_, get_cublas_datatype<cuTypeB>(), ldb, stride_b, &beta, c_,
                                get_cublas_datatype<cuTypeC>(), ldc, stride_c, batch_size,
                                get_cublas_datatype<cuTypeS>(), cublas_gemm_algo);
#else
            CUBLAS_ERROR_FUNC_T_SYNC(
                "cublasGemmStridedBatchedEx", cublasGemmStridedBatchedEx, err, handle,
                get_cublas_operation(transa), get_cublas_operation(transb), m, n, k, &alpha, a_,
                get_cublas_datatype<cuTypeA>(), lda, stride_a, b_, get_cublas_datatype<cuTypeB>(),
                ldb, stride_b, &beta, c_, get_cublas_datatype<cuTypeC>(), ldc, stride_c, batch_size,
                get_cublas_datatype<cuTypeS>(), cublas_gemm_algo);
#endif
```

### DIFF
--- a/src/blas/backends/cublas/cublas_level3.cpp
+++ b/src/blas/backends/cublas/cublas_level3.cpp
@@ -95,15 +95,15 @@ inline void gemm_ex(DATATYPE_A DT_A, DATATYPE_B DT_B, DATATYPE_C DT_C, sycl::que
             auto c_ = sc.get_mem<cuDataType_C*>(c_acc);
             cublasStatus_t err;
 #ifdef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
+            CUBLAS_ERROR_FUNC(cublasGemmEx, err, handle, get_cublas_operation(transa),
+                              get_cublas_operation(transb), m, n, k, (cuDataType_C*)&alpha, a_,
+                              DT_A, lda, b_, DT_B, ldb, (cuDataType_C*)&beta, c_, DT_C, ldc, DT_C,
+                              CUBLAS_GEMM_DEFAULT);
+#else
             CUBLAS_ERROR_FUNC_SYNC(cublasGemmEx, err, handle, get_cublas_operation(transa),
                                    get_cublas_operation(transb), m, n, k, (cuDataType_C*)&alpha, a_,
                                    DT_A, lda, b_, DT_B, ldb, (cuDataType_C*)&beta, c_, DT_C, ldc,
                                    DT_C, CUBLAS_GEMM_DEFAULT);
-#else
-            CUBLAS_ERROR_FUNC(cublasGemmEx, err, handle, get_cublas_operation(transa),
-                                   get_cublas_operation(transb), m, n, k, (cuDataType_C *)&alpha,
-                                   a_, DT_A, lda, b_, DT_B, ldb, (cuDataType_C *)&beta, c_, DT_C,
-                                   ldc, DT_C, CUBLAS_GEMM_DEFAULT);
 #endif
         });
     });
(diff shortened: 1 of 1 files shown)


src/blas/backends/cublas/cublas_level3.cpp
```diff
@@ -95,12 +95,12 @@ inline void gemm_ex(DATATYPE_A DT_A, DATATYPE_B DT_B, DATATYPE_C DT_C, sycl::que
...
-                                   get_cublas_operation(transb), m, n, k, (cuDataType_C*)&alpha, a_,
-                                   DT_A, lda, b_, DT_B, ldb, (cuDataType_C*)&beta, c_, DT_C, ldc,
-                                   DT_C, CUBLAS_GEMM_DEFAULT);
-#else
             CUBLAS_ERROR_FUNC(cublasGemmEx, err, handle, get_cublas_operation(transa),
+                              get_cublas_operation(transb), m, n, k, (cuDataType_C*)&alpha, a_,
+                              DT_A, lda, b_, DT_B, ldb, (cuDataType_C*)&beta, c_, DT_C, ldc, DT_C,
+                              CUBLAS_GEMM_DEFAULT);
+#else
+            CUBLAS_ERROR_FUNC_SYCN(cublasGemmEx, err, handle, get_cublas_operation(transa),
```
andrewtbarker:
```suggestion
            CUBLAS_ERROR_FUNC_SYNC(cublasGemmEx, err, handle, get_cublas_operation(transa),
```

EwanC: woops, thanks for spotting that. I did an AdaptiveCpp build locally and it caught that this didn't compile, so I'll try to build with that to double check PRs going forward.

Rbiessy (approved):
It is very strange that the previous version did not call `CUBLAS_ERROR_FUNC_SYNC` when `host_task` are used, I guess this was a mistake when this was added?

EwanC:
> It is very strange that the previous version did not call `CUBLAS_ERROR_FUNC_SYNC` when `host_task` are used, I guess this was a mistake when this was added?

@JackAKirk Looks like you touched this in https://github.com/uxlfoundation/oneMath/commit/7adfbccee91c00a2dfed4a9c92aba8dacd9fd5c4#diff-639b47d81bc0402ef50f1fe8d975d18c927f8a9d1b6d2b98c955dfdc79ccd904 Does this change make sense? Assuming was an oversight in your patch

JackAKirk:
> > It is very strange that the previous version did not call `CUBLAS_ERROR_FUNC_SYNC` when `host_task` are used, I guess this was a mistake when this was added?
> 
> @JackAKirk Looks like you touched this in [7adfbcc#diff-639b47d81bc0402ef50f1fe8d975d18c927f8a9d1b6d2b98c955dfdc79ccd904](https://github.com/uxlfoundation/oneMath/commit/7adfbccee91c00a2dfed4a9c92aba8dacd9fd5c4#diff-639b47d81bc0402ef50f1fe8d975d18c927f8a9d1b6d2b98c955dfdc79ccd904) Does this change make sense? Assuming was an oversight in your patch

Yes I've mixed up the sync and non-sync variants in the blas3 file only. Thanks for fixing my mistake!

src/blas/backends/cublas/cublas_level3.cpp
EwanC: Thanks for reviewing this patch @andrewtbarker , do you have any more feedback or can you approve the PR?

andrewtbarker (approved):

PR #701, opened 2025-07-17
[LAPACK] Dynamically load reference lapack library for tests. 

adegomme:
# Description

Lapack tests suffered same issue as BLAS tests which was reported in #204 and fixed in #251 . If a backend uses the same LAPACKE interface and symbols as the reference LAPACK, it will end up comparing only the first found library to itself.

This basically copies #251 solution by using dlopen each time the reference library symbols are needed instead of linking to the library directly.

In the process of testing this, I noticed that Lapack tests were not exactly working if only the lapack backend is built, as it depends on the BLAS library as well for accuracy checks. A second commit was added to make Lapack tests work alone, by checking that BLAS reference library can be found even if BLAS backend is not built. It also adds REF_LAPACK_ROOT in the search path for BLAS library, as in the reference design all blas/lapack libraries are in the same location (this avoids having to provide both REF_LAPACK_ROOT and REF_BLAS_ROOT in this case). 

100% tests passed, 0 tests failed out of 796

### DIFF
--- a/examples/lapack/compile_time_dispatching/CMakeLists.txt
+++ b/examples/lapack/compile_time_dispatching/CMakeLists.txt
@@ -31,7 +31,7 @@ set(EXAMPLE_TARGET example_lapack_getrs_usm_mklcpu_cusolver)
 add_executable(${EXAMPLE_TARGET} getrs_usm_mklcpu_cusolver.cpp)
 
 # May be needed to compile source files with SYCL kernels
-if (USE_ADD_SYCL_TO_TARGET_INTEGRATION)
+if(USE_ADD_SYCL_TO_TARGET_INTEGRATION)
   add_sycl_to_target(TARGET ${EXAMPLE_TARGET} SOURCES getrs_usm_mklcpu_cusolver.cpp)
 endif()
 
--- a/examples/lapack/compile_time_dispatching/CMakeLists.txt
+++ b/examples/lapack/compile_time_dispatching/CMakeLists.txt
@@ -30,6 +30,11 @@ set(EXAMPLE_TARGET example_lapack_getrs_usm_mklcpu_cusolver)
 # Create a CMake target with one source file
 add_executable(${EXAMPLE_TARGET} getrs_usm_mklcpu_cusolver.cpp)
 
+# May be needed to compile source files with SYCL kernels
+if (USE_ADD_SYCL_TO_TARGET_INTEGRATION)
+  add_sycl_to_target(TARGET ${EXAMPLE_TARGET} SOURCES getrs_usm_mklcpu_cusolver.cpp)
+endif()
+
 # Linking against onemath_lapack_mklcpu and onemath_lapack_cusolver in CMake will add the required include directories and dependencies.
 # One can also link against `onemath_lapack` to link against all the lapack backends built.
 # These targets should only be used for compile-time dispatching.
--- a/examples/lapack/run_time_dispatching/CMakeLists.txt
+++ b/examples/lapack/run_time_dispatching/CMakeLists.txt
@@ -46,7 +46,7 @@ set(EXAMPLE_TARGET example_lapack_getrs_usm)
 add_executable(${EXAMPLE_TARGET} getrs_usm.cpp)
 
 # May be needed to compile source files with SYCL kernels
-if (USE_ADD_SYCL_TO_TARGET_INTEGRATION)
+if(USE_ADD_SYCL_TO_TARGET_INTEGRATION)
   add_sycl_to_target(TARGET ${EXAMPLE_TARGET} SOURCES getrs_usm.cpp)
 endif()
 
--- a/examples/lapack/run_time_dispatching/CMakeLists.txt
+++ b/examples/lapack/run_time_dispatching/CMakeLists.txt
@@ -39,12 +39,16 @@ endif()
 message(STATUS "ONEAPI_DEVICE_SELECTOR will be set to the following value(s): [${DEVICE_FILTERS}] for run-time dispatching examples")
 
 set(EXAMPLE_TARGET example_lapack_getrs_usm)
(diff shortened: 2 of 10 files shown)


andrewtbarker (approved):
This all looks good. Thanks for the changes!

tests/unit_tests/CMakeLists.txt
```diff
@@ -19,12 +19,15 @@
...
 
-if("blas" IN_LIST TEST_TARGET_DOMAINS)
+if(("blas" IN_LIST TEST_TARGET_DOMAINS) OR ("lapack" IN_LIST TEST_TARGET_DOMAINS))
   find_package(CBLAS)
   if(NOT CBLAS_FOUND)
     # TODO: add list of tests without Netlib dependency
-    message(WARNING "Netlib CBLAS headers or libraries are not found, BLAS unit tests will be skipped")
+    message(WARNING "Netlib CBLAS headers or libraries are not found, BLAS/LAPACK unit tests will be skipped")
     list(REMOVE_ITEM TEST_TARGET_DOMAINS "blas")
+    if  ("lapack" IN_LIST TEST_TARGET_DOMAINS)
```
dnhsieh-intel: Could you please check if it can work without this condition?

dnhsieh-intel (commented):
Thanks for the changes! I only have some minor comments.

One more thing is that the original inconsistency (`reference_blas_templates.hpp` in BLAS and `lapack_reference_wrappers.hpp` in LAPACK) now leads to:
- `reference_blas_templates.hpp` includes `reference_blas_wrappers.hpp` in BLAS
- `lapack_reference_wrappers.hpp` includes `lapack_reference_templates.hpp` in LAPACK

I will leave this for @andrewtbarker to decide if we need alignment.

examples/lapack/compile_time_dispatching/CMakeLists.txt
```diff
@@ -30,6 +30,11 @@ set(EXAMPLE_TARGET example_lapack_getrs_usm_mklcpu_cusolver)
 # Create a CMake target with one source file
 add_executable(${EXAMPLE_TARGET} getrs_usm_mklcpu_cusolver.cpp)
 
+# May be needed to compile source files with SYCL kernels
+if (USE_ADD_SYCL_TO_TARGET_INTEGRATION)
```
dnhsieh-intel: nit
```suggestion
if(USE_ADD_SYCL_TO_TARGET_INTEGRATION)
```

examples/lapack/run_time_dispatching/CMakeLists.txt
```diff
@@ -39,13 +39,17 @@ endif()
...
 set(EXAMPLE_TARGET example_lapack_getrs_usm)
-
 # External applications should use find_package or FetchContent to include oneMath first.
 # See https://github.com/uxlfoundation/oneMath/blob/develop/docs/using_onemath_with_cmake.rst
 
 # Create a CMake target with one source file
 add_executable(${EXAMPLE_TARGET} getrs_usm.cpp)
 
+# May be needed to compile source files with SYCL kernels
+if (USE_ADD_SYCL_TO_TARGET_INTEGRATION)
```
dnhsieh-intel: nit
```suggestion
if(USE_ADD_SYCL_TO_TARGET_INTEGRATION)
```

tests/unit_tests/lapack/source/CMakeLists.txt
```diff
@@ -86,7 +86,11 @@ if(BUILD_SHARED_LIBS)
       PUBLIC ${CMAKE_BINARY_DIR}/bin
       $<$<BOOL:${LAPACKE_FOUND}>:${LAPACKE_INCLUDE}>
   )
-  target_link_libraries(lapack_source_rt PUBLIC ONEMATH::SYCL::SYCL)
+  if (USE_ADD_SYCL_TO_TARGET_INTEGRATION)
```
dnhsieh-intel: nit
```suggestion
  if(USE_ADD_SYCL_TO_TARGET_INTEGRATION)
```

tests/unit_tests/lapack/source/CMakeLists.txt
```diff
@@ -99,4 +103,8 @@ target_include_directories(lapack_source_ct
     PUBLIC ${CMAKE_BINARY_DIR}/bin
     $<$<BOOL:${LAPACKE_FOUND}>:${LAPACKE_INCLUDE}>
 )
-target_link_libraries(lapack_source_ct PUBLIC ONEMATH::SYCL::SYCL)
+if (USE_ADD_SYCL_TO_TARGET_INTEGRATION)
+	add_sycl_to_target(TARGET lapack_source_ct SOURCES ${LAPACK_SOURCES})
```
dnhsieh-intel: nit
```suggestion
if(USE_ADD_SYCL_TO_TARGET_INTEGRATION)
  add_sycl_to_target(TARGET lapack_source_ct SOURCES ${LAPACK_SOURCES})
```

andrewtbarker:
> Thanks for the changes! I only have some minor comments.
> 
> One more thing is that the original inconsistency (`reference_blas_templates.hpp` in BLAS and `lapack_reference_wrappers.hpp` in LAPACK) now leads to:
> 
> * `reference_blas_templates.hpp` includes `reference_blas_wrappers.hpp` in BLAS
> * `lapack_reference_wrappers.hpp` includes `lapack_reference_templates.hpp` in LAPACK
> 
> I will leave this for @andrewtbarker to decide if we need alignment.

I agree that the inconsistency is not ideal, but I don't feel strongly about the exact file names.

tests/unit_tests/CMakeLists.txt
adegomme: it does.

dnhsieh-intel (approved):

PR #702, opened 2025-07-18
[DFT] Add ArmPl DFT backend for aarch64 CPUs

### LINKED ISSUE #649: DFT: Scaling behaviour
When cuFFT was merged, it was noted in the discussion that it was not supporting forward/backward scaling, and some tests were adapted to remove BACKWARD_SCALE parameter and apply scaling during result checking only. 

Discussion originally here:
_Originally posted by @lhuot in https://github.com/uxlfoundation/oneMath/pull/284#discussion_r1171245931_

But this was not done everywhere, as cuFFT does not support real_real storage. Thus these tests have kept using BACKWARD_SCALE  and expect the backend to perform the scaling, or fail verification.

I'm currently adding support for ArmPl/FFTW, and the behaviour here is the same as in cuFFT, scaling is left to handle by the user after computations (rationale here https://www.fftw.org/faq/section3.html#whyscaled ).
But it also supports _split_ calls (real_real), and then I'm in the zone where I can't fix tests without breaking the others. So I would like to reopen the discussion to try to end up with a more unified solution:

- Either all tests ignore scaling, and backends not supporting it warn (like cuFFT), so we remove BACKWARD_SCALE references from compule_in(out_of_)place_real_real.hpp (but this means this goes untested and result actually vary depending on the backend (which is already the case))
- Either we force all backends to perform the scaling every time (performance wise it may not be a good idea). I started doing this but we have to be careful to actually do the full journey through the dimensions/strides/offsets which is normally done by the library.
- Or we find a way where the backends report if they are willing to do the scaling or not, and tests adapt ? but this may change external API or add a query.

adegomme:
# Description

This adds a DFT backend targeting aarch64 CPUs with ArmPl library.

This is basically a FFTW backend, as ArmPl uses FFTW, and can/will be generalized to other FFTW libraries.
A change has been made to tests to account for backends not supporting scaling during the DFT calls and do the scaling after, as is done in FFTW  ( fixes #649 ).

# Checklist

## All Submissions

- [ x] Do all unit tests pass locally? Attach a log.
- 100% tests passed, 0 tests failed out of 1878
1414 passed, 464 skipped, 
[out_dft_dpcpp.txt](https://github.com/user-attachments/files/21314031/out_dft_dpcpp.txt)


tests/unit_tests/dft/include/compute_inplace_real_real.hpp
```diff
@@ -44,7 +45,14 @@ int DFT_Test<precision, domain>::test_in_place_real_real_USM() {
         descriptor.set_value(oneapi::math::dft::config_param::BWD_DISTANCE, forward_elements);
         descriptor.set_value(oneapi::math::dft::config_param::BACKWARD_SCALE, backward_scale);
 
-        commit_descriptor(descriptor, sycl_queue);
+        try {
+            commit_descriptor(descriptor, sycl_queue);
+        }
+        catch (...) { //BACKWARD_SCALE not supported, go with 1 and scale at check
```
lhuot: commit can throw several exceptions including the one about scaling not supported so I would recommend checking the exception being caught here to make sure it was indeed because of the scaling and not something else.

tests/unit_tests/dft/include/compute_inplace_real_real.hpp
```diff
@@ -44,7 +45,14 @@ int DFT_Test<precision, domain>::test_in_place_real_real_USM() {
...
+            descriptor.set_value(oneapi::math::dft::config_param::BACKWARD_SCALE, 1.f);
+            commit_descriptor(descriptor, sycl_queue);
```
lhuot, line 54: It would probably be more robust to check again the exception here to make sure no unexpected error happened.

tests/unit_tests/dft/include/compute_inplace_real_real.hpp
```diff
@@ -104,8 +117,14 @@ int DFT_Test<precision, domain>::test_in_place_real_real_buffer() {
...
         descriptor.set_value(oneapi::math::dft::config_param::BWD_DISTANCE, forward_elements);
         descriptor.set_value(oneapi::math::dft::config_param::BACKWARD_SCALE, backward_scale);
-
-        commit_descriptor(descriptor, sycl_queue);
+        try {
+            commit_descriptor(descriptor, sycl_queue);
+        }
+        catch (...) { //BACKWARD_SCALE not supported, go with 1 and scale at check
+            descriptor.set_value(oneapi::math::dft::config_param::BACKWARD_SCALE, 1.f);
+            commit_descriptor(descriptor, sycl_queue);
```
lhuot: same comment here and in other files

tests/unit_tests/dft/include/compute_inplace_real_real.hpp
adegomme: indeed that's an oversight on my end, I will filter it, thanks

tests/unit_tests/dft/include/compute_inplace_real_real.hpp
adegomme: After some testing, this is already run in a more global handler at a higher level, so another unimplemented exception will trigger a skip (if a second feature is not implemented in commit, and maybe if we caught the wrong one first), and other exceptions will cause the test to fail, as intended
So this should be robust already (with the risk of catching the wrong "unimplemented" exception on the first attempt, and retrying once before failing/skipping the test as intended).

tests/unit_tests/dft/include/compute_inplace_real_real.hpp
adegomme: I made the change, but when I tried switching exception type from "unimplemented" to "invalid_argument", other tests started to fail. Indeed unimplemented exception is a non-failure case in some descriptor tests, while other exceptions would fail these tests. 
So i'm letting it as is for now, but if anyone prefers a "invalid_argument" exception, I'll have to modify more test files.

lhuot: I am fine with throwing unimplemented for scale != 1, I was not suggesting throwing "invalid_argument", just updating the catch in the test. So looks good to me.

lhuot:
Could you please update .github/workflows/pr-arm.yml to include DFT testing with ArmPL backend?

lhuot (approved):
Thank you for adding the ArmPL backend to the DFT domain and enabling testing in the CI! The changes look good to me.

lhuot:
@dnhsieh-intel could you please help review the PR for onemath-arch-write?

dnhsieh-intel (approved):
Thanks for the ping! I missed it because of "[DFT]" in the title. I need to update my mental filter...

Anyway, the changes look good to me. Thank you!

PR #707, opened 2025-08-07
OpenBLAS backend support for OneMath

CDAC-SSDG:
# Description

We aim to integrate OpenBLAS as a backend for the BLAS domain in the oneAPI Math Library (oneMath) to provide a high performance, open source alternative to vendor specific BLAS implementations. The integration will target CPU devices on x86 and ARM architectures, ensuring optimized performance across supported platforms using DPC++, ParaS and AdaptiveCPP compilers.

### DIFF
--- a/rfcs/20250806-OpenBLAS-Backend/README.md
+++ b/rfcs/20250806-OpenBLAS-Backend/README.md
@@ -22,7 +22,7 @@ Users will face reduced performance, limited portability, and continued reliance
 
 ### What is the timeline?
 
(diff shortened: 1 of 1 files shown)


sknepper (approved):
Thanks for this very detailed, thoughtful RFC! I agree that offering OpenBLAS as a backend aligns well with the goals of oneMath/UXL Foundation.
One thing to keep in mind, and I don't think it needs to be covered in the RFC, is how to handle any functionality that has oneMath APIs but no direct OpenBLAS APIs (e.g., some batched BLAS). The oneMath spec allows to throw an unimplemented exception, or the oneMath implementation could potentially still implement the functionality using existing OpenBLAS APIs (e.g., loops around non-batched BLAS).
Thank you for your contribution!

melonakos (approved):

PR #710, opened 2025-09-17
[BLAS][SPARSE][DPC++] Add CUDA_TARGETS option to pass for AoT compila…

abagusetty:
…tion

1. Adds an option to use `CUDA_TARGETS` similar to existing option for `HIP_TARGETS` for AoT builds of DPC++ compiler
2. `generic` `blas` already has stubs to use arch flags but was not set

### DIFF
--- a/src/blas/backends/cublas/CMakeLists.txt
+++ b/src/blas/backends/cublas/CMakeLists.txt
@@ -41,12 +41,21 @@ target_include_directories(${LIB_OBJ}
 )
 target_compile_options(${LIB_OBJ} PRIVATE ${ONEMATH_BUILD_COPT})
 
-if(NOT ${ONEMATH_SYCL_IMPLEMENTATION} STREQUAL "adaptivecpp")
-  if(DEFINED CUDA_TARGETS)
+if (NOT "${ONEMATH_SYCL_IMPLEMENTATION}" STREQUAL "adaptivecpp")
+  target_compile_options(ONEMATH::SYCL::SYCL INTERFACE
+    -fsycl-targets=nvptx64-nvidia-cuda -fsycl-unnamed-lambda
+  )
+  target_link_options(ONEMATH::SYCL::SYCL INTERFACE
+    -fsycl-targets=nvptx64-nvidia-cuda
+  )
+
+  if (DEFINED CUDA_TARGETS AND NOT "${CUDA_TARGETS}" STREQUAL "")
     target_compile_options(ONEMATH::SYCL::SYCL INTERFACE
-      -Xsycl-target-backend --cuda-gpu-arch=${CUDA_TARGETS})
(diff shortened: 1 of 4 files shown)


src/blas/backends/cublas/CMakeLists.txt
```diff
@@ -42,11 +42,14 @@ target_include_directories(${LIB_OBJ}
...
 
 if(NOT ${ONEMATH_SYCL_IMPLEMENTATION} STREQUAL "adaptivecpp")
+  if(DEFINED CUDA_TARGETS)
     target_compile_options(ONEMATH::SYCL::SYCL INTERFACE
-          -fsycl-targets=nvptx64-nvidia-cuda -fsycl-unnamed-lambda)
+      -Xsycl-target-backend --cuda-gpu-arch=${CUDA_TARGETS})
     target_link_options(ONEMATH::SYCL::SYCL INTERFACE
-          -fsycl-targets=nvptx64-nvidia-cuda)
+      -Xsycl-target-backend --cuda-gpu-arch=${CUDA_TARGETS})
+  endif()
```
andrewtbarker, line 59: Why not have an `else` here and maintain the previous behavior in case `CUDA_TARGETS` is not defined?

abagusetty: Thanks for reviewing, addressed it.

andrewtbarker (approved):

src/sparse_blas/backends/cusparse/CMakeLists.txt
```diff
@@ -43,6 +43,24 @@ target_include_directories(${LIB_OBJ}
...
+  if (DEFINED CUDA_TARGETS AND NOT "${CUDA_TARGETS}" STREQUAL "")
+    target_compile_options(ONEMATH::SYCL::SYCL INTERFACE
+      -Xsycl-target-backend --cuda-gpu-arch=${CUDA_TARGETS}
+    )
+    target_link_options(ONEMATH::SYCL::SYCL INTERFACE
+      -Xsycl-target-backend --cuda-gpu-arch=${CUDA_TARGETS}
+    )
+  endif()
+endif()
+
```
spencerpatty, line 63: this is fine, it is same update as for BLAS, for CUDA + SYCL linking, so no reason to object

abagusetty: Correct, was just adding it for consistency between the domains. Having a way to add AoT support too and to avoid issues with newer CUDA versions
Context: CUDA 13 dropped support for several older NVIDIA GPU architectures, specifically Maxwell (compute capability 5.x), Pascal (compute capability 6.x), and Volta (compute capability 7.0, 7.2) architectures.

spencerpatty (approved):
approved from Sparse BLAS perspective

PR #718, opened 2026-02-18
[mklcpu][mklgpu] use dynamic link for oneMKL, add MKLCPU_THREAD_OMP

### LINKED ISSUE #654: oneMath links Intel oneMKL statically on Windows
# Summary
The CMake logics set `MKL_LINK` to `static` when building for Windows:
https://github.com/uxlfoundation/oneMath/blob/c3c85db9f0946b1bd95e5bee17ee5cfaf90f086f/CMakeLists.txt#L231-L235

This creates a large library on Windows of around 100MB, whereas on Linux with dynamic linking it is around 3MB for the blas domain for instance. This is not practical to integrate in other applications.
I am wondering if someone from @uxlfoundation/onemath-arch-write know why this is the case? Would there be any risk to switch to linking Intel oneMKL dynamically today?

# Version
It looks like this was introduced in [c3c85db9](https://github.com/uxlfoundation/oneMath/commit/c3c85db9#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR231-R235)
Pinging @mkrainiuk in case you have more context.

# Environment
The issue happens on Windows with the Intel oneMKL backends.

mkrainiuk:
# Description

PR switches MKLCPU and MKLGPU backend to always build with dynamic Intel oneMKL on both Linux and Windows (before it was static only for Windows), because of static oneMKL SYCL library deprecation.

PR also introduces one more build switch `ENABLE_MKLCPU_THREAD_OMP` to enable building MKLCPU with OpenMP threading instead of TBB. It can be enabled with `-DENABLE_MKLCPU_THREAD_OMP=ON -DENABLE_MKLCPU_THREAD_TBB=OFF`, since TBB is still default one.

+ some fixes for missed BLAS dependency from LAPACK on Linux and incorrect Netlib LAPACK lib names used for loading them in Unit tests on Windows (should be *.dll, not `LAPACK64_file` which contains *.lib for linking)

Fixes #654 (but it's not really a bug, so nothing to add as a regression test)

# Checklist

## All Submissions

- [x] Do all unit tests pass locally? Attach a log. There are several failed LAPACK tests, but they look unrelated to the changes, most likely GPU stack instability. Same for Linux, all tests passed except for 2 failed BLAS tests because of accuracy, I see these failures without the changes, so not related. 
[log_shared_off.log](https://github.com/user-attachments/files/25399041/log_shared_off.log)
[log_shared_on.log](https://github.com/user-attachments/files/25399046/log_shared_on.log)

There are more accuracy issues in ArmPL BLAS/LAPACK tests but it is also not related to the PR.


andrewtbarker (approved):
Thanks for the changes!

vmalia (approved):
LGTM!

PR #722, opened 2026-04-13
Openblas backend implementation

CDAC-SSDG:
# Description

This PR adds support for using OpenBLAS as a new backend within oneMath for BLAS operations.

The implementation follows the design proposed in RFC#707 and integrates OpenBLAS as a CPU backend through the existing backend abstraction layer. BLAS operations are mapped to the OpenBLAS interface while preserving the oneMath API and execution model.

The implementation provides coverage across Level 1, Level 2, and Level 3 BLAS routines, along with support for batch operations and relevant BLAS extensions where applicable. Functionality is aligned with OpenBLAS capabilities.

Build system integration has been added to enable selection of the OpenBLAS backend through CMake configuration. The implementation has been validated with both DPC++ and AdaptiveCpp toolchains.

Documentation has been added describing how to configure and build the project with OpenBLAS support:
- `docs/building_the_project_with_openblas.rst`

## Motivation and Context

This change provides an additional backend option for BLAS operations using OpenBLAS, allowing users to build and run oneMath in environments where other backends may not be available or desired. It improves flexibility in backend selection while remaining consistent with the modular backend design of oneMath.

## Testing
The implementation has been built and tested, test logs are attached with this PR.
[OpenBlas with DPC++ unit tests log.txt](https://github.com/user-attachments/files/26678372/OpenBlas.with.DPC%2B%2B.unit.tests.log.txt)


sknepper:
Thank you much for your contribution, @CDAC-SSDG !
A few questions/comments:
1. More domains than just BLAS had changes. Could you please provide an explanation for those changes? Ideally, separating commits can also help in the review process, but that's probably too challenging at this point.
2. The clang-format check is failing - could you please apply the changes it suggests so that check passes?
3. Some of the ArmPL tests are failing - I'm guessing this isn't due to your changes, but could you please check, particularly for the BLAS domain?
4. As a next step after this PR is merged, having github runners testing the OpenBLAS backend would be a nice check to add (but not required).

cmake/mkl/MKLConfig.cmake
```diff
@@ -980,6 +980,11 @@ if(NOT MKL_THREADING STREQUAL "tbb_thread" AND MKL_THREADING MATCHES "_thread")
              "windows/compiler/lib/${MKL_ARCH}_win"
              "../compiler/lib/${MKL_ARCH}_lin" "../compiler/lib/${MKL_ARCH}_win"
              "../compiler/lib/${MKL_ARCH}" "../compiler/lib" "compiler/lib"
+             "../../compiler/latest/lib"
+             "../../compiler/latest/linux/compiler/lib/${MKL_ARCH}"
+             "../../compiler/latest/linux/compiler/lib/${MKL_ARCH}_lin"
+             "../../compiler/latest/windows/compiler/lib/${MKL_ARCH}"
+             "../../compiler/latest/windows/compiler/lib/${MKL_ARCH}_win"
```
mkrainiuk, line 987: Thank you for the changes! 
Looks like this branch `CDAC-Bengaluru:openblas-backend-implementation` is not synced with the latest develop branch from the main repo, so this PR contains reverted changes for some recent changes in the develop, like this change is reverting https://github.com/uxlfoundation/oneMath/pull/719.
Could you please rebase your branch on the lates develop so these reverted changes will go away?

include/oneapi/math/detail/backends_table.hpp
```diff
@@ -65,6 +69,10 @@ static std::map<domain, std::map<device, std::vector<const char*>>> libraries =
 #ifdef ONEMATH_ENABLE_NETLIB_BACKEND
               LIB_NAME("blas_netlib")
```
ndingle-arm: This needs a comma after it now that there is a new #ifdef added after it (see line 67 for example). Otherwise if multiple ifdefs are defined we will get one concatenated library name instead of two separate libraries.

tests/unit_tests/CMakeLists.txt
```diff
@@ -20,20 +20,28 @@
...
 if(("blas" IN_LIST TEST_TARGET_DOMAINS) OR ("lapack" IN_LIST TEST_TARGET_DOMAINS))
-  find_package(CBLAS)
-  if(NOT CBLAS_FOUND)
-    # TODO: add list of tests without Netlib dependency
-    message(WARNING "Netlib CBLAS headers or libraries are not found, BLAS/LAPACK unit tests will be skipped")
-    list(REMOVE_ITEM TEST_TARGET_DOMAINS "blas" "lapack")
+  if(ENABLE_OPENBLAS_BACKEND)
+    message(STATUS "Using OpenBLAS for BLAS/LAPACK unit tests")
+    set(CBLAS_FOUND TRUE)
+    set(LAPACKE_FOUND TRUE)
```
ndingle-arm: How do we guarantee when ENABLE_OPENBLAS_BACKEND is passed in that we will definitely find CBLAS and LAPACKE? We use a reference Netlib for comparison in the tests, but this skips the check that it is actually available when using the OpenBLAS backend.

As a result, LAPACK unit-test sources guarded by LAPACKE_FOUND may be compiled without confirming it is actually available, and this can lead to failures in places like tests/unit_tests/lapack/include/lapack_reference_wrappers.hpp (missing cblas.h / lapacke.h). This can happen when no REF_* roots are provided to the CMake invocation.

The BLAS test path might also be unsafe if CMake is invoked without REF_* roots: with ENABLE_OPENBLAS_BACKEND=ON, the tests include OpenBLAS cblas.h, and if ONEMATH_REF_* is empty the wrapper layer can resolve cblas_* from the already-linked OpenBLAS-backed process image. That means BLAS reference checks might end up comparing OpenBLAS-backed oneMath against OpenBLAS itself.

sknepper:
Hi @CDAC-SSDG  - could you please apply the changes from the clang-format check so it passes? Thanks!

sknepper (approved):
@CDAC-SSDG - thank you very much for this valuable contribution! The changes look good to me.
@ndingle-arm  - could you please check the latest changes and verify that all your questions have been answered?

src/blas/backends/openblas/openblas_common.hpp
```diff
@@ -0,0 +1,103 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions
+* and limitations under the License.
+*
+*
+* SPDX-License-Identifier: Apache-2.0
+*******************************************************************************/
+
+#ifndef _NETLIB_COMMON_HPP_
+#define _NETLIB_COMMON_HPP_
```
ndingle-arm: I think this should be _OPENBLAS_COMMON_HPP_ , or else you are using the same guard as in src/blas/backends/netlib/netlib_common.hpp and therefore risk inadvertently excluding one of these two files.

src/blas/backends/openblas/openblas_extensions.cxx
```diff
@@ -0,0 +1,667 @@
...
+}
+
+void imatcopy(sycl::queue&, transpose trans, int64_t m, int64_t n, std::complex<float> alpha,
+              sycl::buffer<std::complex<float>, 1>& ab, int64_t lda, int64_t ldb) {
+    auto acc = ab.get_access<sycl::access::mode::read_write>();
+
+    float* A = reinterpret_cast<float*>(acc.get_pointer());
+    const float* alpha_ptr = reinterpret_cast<const float*>(&alpha);
+
+    CBLAS_TRANSPOSE t = (trans == transpose::nontrans) ? CblasNoTrans : CblasTrans;
```
ndingle-arm: This maps every non-notrans case to CblasTrans and does not differentiate between CblasTrans and CblasConjTrans like it should. Compare with complex omatcopy above, which does:

    CBLAS_TRANSPOSE t;
    if (trans == transpose::nontrans) {
        t = CblasNoTrans;
    }
    else if (trans == transpose::trans) {
        t = CblasTrans;
    }
    else {
        t = CblasConjTrans;
    }

src/blas/backends/openblas/openblas_extensions.cxx
```diff
@@ -0,0 +1,667 @@
...
+}
+
+void imatcopy(sycl::queue&, transpose trans, int64_t m, int64_t n, std::complex<double> alpha,
+              sycl::buffer<std::complex<double>, 1>& ab, int64_t lda, int64_t ldb) {
+    auto acc = ab.get_access<sycl::access::mode::read_write>();
+
+    double* A = reinterpret_cast<double*>(acc.get_pointer());
+    const double* alpha_ptr = reinterpret_cast<const double*>(&alpha);
+
+    CBLAS_TRANSPOSE t = (trans == transpose::nontrans) ? CblasNoTrans : CblasTrans;
```
ndingle-arm: Same comment as for cimatcopy -- it needs to consider conjugate-transposition as well.

src/blas/backends/openblas/openblas_level1.cxx
```diff
@@ -0,0 +1,1659 @@
...
+    auto y_acc = y.get_access<sycl::access::mode::read_write>();
+
+    const float* X = x_acc.get_pointer();
+    float* Y = y_acc.get_pointer();
+
+    blasint N = (blasint)n;
+    blasint incX = (blasint)incx;
+    blasint incY = (blasint)incy;
+
+    cblas_saxpby(N, alpha, X, incX, beta, Y, incY);
```
ndingle-arm: I am not sure why [cdsz]axpby calls the corresponding CBLAS routine directly, but [cdsz]axpy and all other functions here go via the queue, e.g. in axpy it does queue.submit([&](sycl::handler& cgh) {...} but here queue is not used at all (and in other axpby precisions too)?

src/blas/backends/openblas/openblas_batch.cxx
```diff
@@ -0,0 +1,1696 @@
...
+#ifdef ROW_MAJOR
+    constexpr CBLAS_ORDER order = CblasRowMajor;
+#endif
+
+    for (int64_t i = 0; i < batch_size; ++i) {
+        const float* A = A0 + i * stride_a;
+        const float* B = B0 + i * stride_b;
+        float* C = C0 + i * stride_c;
+
+        cblas_sgemm(order, tA, tB, (blasint)m, (blasint)n, (blasint)k, alpha, A, (blasint)lda, B,
```
ndingle-arm: I would expect [cdsz]gemm_batch to go via queue.submit like the other BLAS calls, rather than calling the CBLAS function directly (this comment occurs in several files).

src/blas/backends/openblas/openblas_extensions.cxx
```diff
@@ -0,0 +1,667 @@
...
+    float* B = b_acc.get_pointer();
+
+    CBLAS_TRANSPOSE t = (trans == transpose::nontrans) ? CblasNoTrans : CblasTrans;
+
+#ifdef COLUMN_MAJOR
+    cblas_somatcopy(CblasColMajor, t, m, n, alpha, A, lda, B, ldb);
+#endif
+
+#ifdef ROW_MAJOR
+    cblas_somatcopy(CblasRowMajor, t, m, n, alpha, A, lda, B, ldb);
```
ndingle-arm: As above, I would expect these calls to use queue.submit() rather than direct CBLAS calls, based on how other backends are implmented.

src/blas/backends/openblas/openblas_extensions.cxx
```diff
@@ -0,0 +1,667 @@
...
+
+void imatcopy(sycl::queue&, transpose trans, int64_t m, int64_t n, float alpha,
+              sycl::buffer<float, 1>& ab, int64_t lda, int64_t ldb) {
+    auto acc = ab.get_access<sycl::access::mode::read_write>();
+    float* A = acc.get_pointer();
+
+    CBLAS_TRANSPOSE t = (trans == transpose::nontrans) ? CblasNoTrans : CblasTrans;
+
+#ifdef COLUMN_MAJOR
+    cblas_simatcopy(CblasColMajor, t, m, n, alpha, A, lda, ldb);
```
ndingle-arm: As above, I would expect these calls to use queue.submit() rather than direct CBLAS calls, based on how other backends are implmented.

src/blas/backends/openblas/CMakeLists.txt
```diff
@@ -0,0 +1,97 @@
...
+add_dependencies(onemath_backend_libs_blas ${LIB_NAME})
+
+if (USE_ADD_SYCL_TO_TARGET_INTEGRATION)
+  add_sycl_to_target(TARGET ${LIB_OBJ} SOURCES ${SOURCES})
+endif()
+
+target_include_directories(${LIB_OBJ}
+  PUBLIC
+      ${ONEMATH_INCLUDE_DIRS}
+      ${OPENBLAS_INCLUDE}
```
ndingle-arm: Backend-specific include directories are exported as PRIVATE by other backends (e.g. src/blas/backends/netlib/CMakeLists.txt:39) . Should OpenBLAS do the same? Otherwise a build machine’s OpenBLAS header path will be exported as part of its public CMake interface.

ndingle-arm (changes requested):
Thank you for the change @CDAC-SSDG but I spotted a few more things reading through again.

ndingle-arm (approved):
Thank you for the changes, they look good to me.

mkrainiuk (approved):
Thanks for the changes! Looks good to me. I guess as the next step it will be very useful to enable this backend in [github workflows](https://github.com/uxlfoundation/oneMath/tree/develop/.github/workflows) to make sure this backend will be tested with any new changes.

PR #723, opened 2026-04-22
Fix failing ArmPL SCAL tests

ndingle-arm:
# Description

The BLAS SCAL function returns immediately if INCX<=0. The reference USM calls pass in ABS(INCX) but the ArmPL calls were just passing in INCX. We were therefore comparing functions called with two different inputs, and the tests were failing. The fix is to pass ABS(INCX) into the ArmPL calls as well, in common with other ArmPL calls to BLAS functions in this file (e.g. compare the ArmPL and USM nrm2 calls).

# Checklist

## All Submissions

[test.log](https://github.com/user-attachments/files/26964849/test.log)


ndingle-arm:
This fixes the failing SCAL tests, but there are still further unrelated BLAS rank-update failures that I will investigate separately.

adegomme:
Indeed I just noticed this when testing 26.01. It was passing these tests with previous versions, so something changed in armpl, no?
fix lgtm anyway, thanks a lot

mkrainiuk (approved):
Thanks for the fix! Looks good to me.

andrewtbarker (approved):

PR #745, opened 2026-06-09
OpenBLAS-LAPACK backend support for OneMath

CDAC-SSDG:
We aim to integrate OpenBLAS LAPACK support as a backend for the LAPACK domain in the oneAPI Math Library (oneMath) to provide a high-performance, open-source alternative to existing LAPACK implementations. The integration will target CPU devices on x86 and ARM architectures, enabling portable and optimized dense linear algebra functionality across supported platforms. By leveraging the LAPACKE interface provided by OpenBLAS, this effort aims to enhance oneMath's backend flexibility, improve performance portability, and reduce dependence on vendor-specific libraries while maintaining compatibility with existing DPC++ and AdaptiveCpp compiler ecosystems.

### DIFF
--- a/rfcs/20260609-OpenBLAS-lapack-backend/README.md
+++ b/rfcs/20260609-OpenBLAS-lapack-backend/README.md
@@ -1,0 +1,175 @@
+# Design Document
+
+### Revision
+
+| Date     | Revision | Comments        |
+| -------- | -------- | --------------- |
+| 20260609 | 1.0      | Initial version |
+
+---
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


sknepper (approved):
This looks like a wonderful proposal, @CDAC-SSDG  - happily approved!
As you stated, this will provide an additional backend for oneMath users, supporting both x86 and ARM platforms.

sknepper:
@ndingle-arm - your review would be much appreciated!

ndingle-arm:
This sounds like a good proposal, and builds on the existing work to add an OpenBLAS BLAS backend.

ndingle-arm (approved):

PR #747, opened 2026-07-28
[DFT][cuFFT] Fix "Invalid strides" for shapes with a unit last dimension (#631)

### LINKED ISSUE #631: [DFT][cuFFT] Invalid strides error on CUDA
# Summary
When performing `DPF transform` with `oneapi::mkl::dft` on NVIDIA GPU for multidimensional arrays >=2D where the `last dimension is 1` [(2,1), (5,1), (2,3,1)....]
The computation fails with `oneMKL: dft/backends/cufft/commit: Invalid strides`

This problem does not occur when running on Intel.

# Version
I am using the latest release version [v0.6](https://github.com/uxlfoundation/oneMath/releases/tag/v0.6) 

# Environment
* HW you use :  NVIDIA A100 80GB PCIe
* Backend library version: Build cuda_12.4.r12.4/compiler.33961263_0
* OS name and version : Linux Ubuntu 24.04 LTS 6.8.0-51-generic
* Compiler version : Intel(R) oneAPI DPC++/C++ Compiler 2025.0.4 (2025.0.4.20241205) 
```
$ sycl-ls
[opencl:cpu][opencl:0] Intel(R) OpenCL, Intel(R) Xeon(R) Platinum 8480+ OpenCL 3.0 (Build 0) [2024.18.12.0.05_160000]
[cuda:gpu][cuda:0] NVIDIA CUDA BACKEND, NVIDIA A100 80GB PCIe 8.0 [CUDA 12.7]

```

# Steps to reproduce

Building:
```
cmake .. -DCMAKE_CXX_COMPILER=icpx -DCMAKE_C_COMPILER=icx -DMKL_ROOT=${MKLROOT} -DBUILD_FUNCTIONAL_TESTS=OFF              \
    -DENABLE_CUBLAS_BACKEND=True  -DENABLE_CUSOLVER_BACKEND=True -DENABLE_CURAND_BACKEND=True -DENABLE_CUFFT_BACKEND=True

# build
cmake --build .

# install
export MKL_INTERFACE_ROOT=$(pwd)/../install
cmake --install . --prefix=${MKL_INTERFACE_ROOT}
```

Compiling:

```
icpx -fsycl fft_repro.cpp -fsycl-targets=nvptx64-nvidia-cuda,spir64-unknown-unknown -I${MKL_INTERFACE_ROOT}/include
\
-L${MKL_INTERFACE_ROOT}/lib -lonemkl -Wl,-rpath,${MKL_INTERFACE_ROOT}/lib -o run_repro

```

Running:
```
ONEAPI_DEVICE_SELECTOR=cuda:gpu ./run_repro
ONEAPI_DEVICE_SELECTOR=opencl:cpu ./run_repro 
```

<details>
 <summary>Reproducer here</summary>

```
#include <iostream>
#include <vector>
#include <sycl/sycl.hpp>
#include <oneapi/mkl/dft.hpp>
#include <complex>

int main() {
    namespace dft = oneapi::mkl::dft;
    sycl::queue q(sycl::default_selector{});

    auto device = q.get_device();
    std::cout << "Device name: " << device.get_info<sycl::info::device::name>() << "\n";

    std::vector<std::int64_t> shape = {3, 1};
    const std::int64_t size = shape[0] * shape[1];

    std::vector<std::complex<float>> h_data(size, {1.0f, 0.0f});
    std::vector<std::complex<float>> h_output(size);

    std::complex<float>* d_data = sycl::malloc_device<std::complex<float>>(size, q);
    std::complex<float>* d_output = sycl::malloc_device<std::complex<float>>(size, q);

    q.memcpy(d_data, h_data.data(), size * sizeof(std::complex<float>)).wait();

    dft::descriptor<dft::precision::SINGLE, dft::domain::COMPLEX> desc(shape);

    std::vector<std::int64_t> fwd_strides = {0, 1, 1};
    std::vector<std::int64_t> bwd_strides = {0, 1, 1};

    desc.set_value(dft::config_param::PLACEMENT, dft::config_value::NOT_INPLACE);
    desc.set_value(dft::config_param::NUMBER_OF_TRANSFORMS, 1);
    // desc.set_value(dft::config_param::FWD_STRIDES, fwd_strides.data());
    // desc.set_value(dft::config_param::BWD_STRIDES, bwd_strides.data());

    desc.commit(q);

    sycl::event fft_event = dft::compute_forward(desc, d_data, d_output);
    fft_event.wait();

    q.memcpy(h_output.data(), d_output, size * sizeof(std::complex<float>)).wait();

    std::cout << "FFT Result:\n";
    for (std::int64_t i = 0; i < size; i++) {
        std::cout << h_output[i] << "\n";
    }

    sycl::free(d_data, q);
    sycl::free(d_output, q);

    return 0;
}
```
</details>

# Observed behavior
When executing the attached code on NVIDIA the program fails with an error

Example output
```
$ ONEAPI_DEVICE_SELECTOR=cuda:gpu ./run_repro 

Device name: NVIDIA A100 80GB PCIe
terminate called after throwing an instance of 'oneapi::mkl::exception'
  what():  oneMKL: dft/backends/cufft/commit: Invalid strides.
Aborted (core dumped)
```

However, running the same code on Intel works as expected
```
$ ONEAPI_DEVICE_SELECTOR=opencl:cpu ./run_repro 

Device name: Intel(R) Xeon(R) Platinum 8480+
FFT Result:
(3,0)
(0,0)
(0,0)
```

# Expected behavior
Similar implementation that uses cuFFT directly works correctly

```
$ ./fft_cuda 
FFT Result:
(3, 0)
(0, 0)
(0, 0)
```

<details>
 <summary>Reproducer here</summary>

```
nvcc -o fft_cuda fft_cuda.cpp -lcufft
```

```
#include <iostream>
#include <vector>
#include <cuda_runtime.h>
#include <cufft.h>
#include <complex>


int main() {
    const int NX = 1;
    const int NY = 3;
    const int size = NX * NY;

    std::vector<std::complex<float>> h_data(size, {1.0f, 0.0f});

    cufftComplex* d_data;
    cudaMalloc(reinterpret_cast<void**>(&d_data), size * sizeof(cufftComplex));

    std::vector<cufftComplex> h_cufft_data(size);
    for (int i = 0; i < size; i++) {
        h_cufft_data[i].x = std::real(h_data[i]);
        h_cufft_data[i].y = std::imag(h_data[i]);
    }

    cudaMemcpy(d_data, h_cufft_data.data(), size * sizeof(cufftComplex), cudaMemcpyHostToDevice);

    cufftHandle plan;
    cufftPlan2d(&plan, NY, NX, CUFFT_C2C);

    cufftExecC2C(plan, d_data, d_data, CUFFT_FORWARD);

    cudaMemcpy(h_cufft_data.data(), d_data, size * sizeof(cufftComplex), cudaMemcpyDeviceToHost);

    std::cout << "FFT Result:\n";
    for (int i = 0; i < size; i++) {
        std::cout << "(" << h_cufft_data[i].x << ", " << h_cufft_data[i].y << ")\n";
    }

    cufftDestroy(plan);
    cudaFree(d_data);

    return 0;
}
```

</details>

zjin-lcf:
## Summary

Fixes #631.

`oneapi::math::dft` on the cuFFT backend throws `dft/backends/cufft/commit: Invalid strides` for multi-dimensional transforms whose **last dimension has extent 1** (e.g. `{3,1}`, `{2,1}`, `{5,1}`, `{2,3,1}`). The same descriptors work on the Intel (MKL) backends, and the equivalent transform works when calling cuFFT directly.

### Root cause

In `src/dft/backends/cufft/commit.cpp`, the backend uses `std::min_element` to find the dimension with the smallest stride (which becomes the cuFFT `istride`/`ostride`, i.e. the innermost dimension).

When a dimension has extent 1, its default stride ties with its neighbour's stride. `std::min_element` returns the **first** element among ties, so an **outer** dimension is mis-identified as the innermost one. This triggers the dimension-swap heuristic, producing a dimension/stride ordering that the `check_stride_validity()` pre-check then rejects — computing `inner_n (e.g. 3) <= inner_stride (1)` as false for both directions and throwing `Invalid strides`. The underlying `cufftPlanMany` call that would have been issued is actually valid.

There is also latent **undefined behaviour**: the swap logic computes the selected index from `a_min` *after* `vec_a.erase(a_min)` has invalidated that iterator. This is why the failure is compiler-dependent (it reproduces with `icpx`, but a recent open-source DPC++/clang build happens to take a different, passing path).

### Fix

- Search for the smallest stride in **reverse**, so ties resolve to the **innermost (highest-index)** dimension. This avoids the spurious dimension swap and lets valid transforms through.
- Capture the selected stride index **before** `erase()`, removing the invalidated-iterator UB.

No behavioural change for shapes without unit dimensions: for strictly decreasing default strides the reverse search selects the same (last) element as before.

## Test plan

Validated end-to-end on an NVIDIA sm_103 device, CUDA 13.2 / cuFFT 12.2, using an open-source DPC++ toolchain (intel/llvm nightly with the CUDA backend). Built oneMath with `-DENABLE_CUFFT_BACKEND=ON -DTARGET_DOMAINS=dft`.

With the fix, the reproducer from #631 and related shapes all pass:

```
shape {3,1}   : OK -> [(3,0) (0,0) (0,0)]
shape {2,1}   : OK -> [(2,0) (0,0)]
shape {5,1}   : OK -> [(5,0) (0,0) (0,0) (0,0) (0,0)]
shape {2,3,1} : OK -> [(6,0) ...]
shape {3,4}   : OK   (regression check)
shape {2,3,4} : OK   (regression check)
shape {8}     : OK   (regression check)
```

- [x] Confirmed direct cuFFT (`cufftPlan2d(3,1)` and the exact `cufftPlanMany` params oneMath generates) succeed on the B300 — the failure is in the pre-check, not cuFFT.
- [x] Standalone comparison of the stride algorithm: original logic REJECTS `{3,1}`, `{2,1}`, `{5,1}`, `{2,3,1}`; fixed logic ACCEPTS them; regression shapes unchanged.
- [x] End-to-end build + run on B300: all shapes above pass.
- [ ] CI: DFT cuFFT functional tests.

Suggestion for maintainers: consider adding a DFT test case covering shapes with a unit last dimension to guard against regressions.

### DIFF
--- a/src/dft/backends/cufft/commit.cpp
+++ b/src/dft/backends/cufft/commit.cpp
@@ -25,6 +25,7 @@
 
 #include <array>
 #include <algorithm>
+#include <iterator>
 #include <optional>
 
 #include "oneapi/math/exceptions.hpp"
@@ -160,9 +161,20 @@ public:
         offset_bwd_in = stride_vecs.offset_bwd_in;
         offset_bwd_out = stride_vecs.offset_bwd_out;
 
-        // cufft ignores the first value in inembed and onembed, so there is no harm in putting offset there
-        auto a_min = std::min_element(stride_vecs.vec_a.begin() + 1, stride_vecs.vec_a.end());
-        auto b_min = std::min_element(stride_vecs.vec_b.begin() + 1, stride_vecs.vec_b.end());
+        // cufft ignores the first value in inembed and onembed, so there is no harm in putting offset there.
+        // Find the dimension with the smallest stride (used as the cuFFT istride/ostride). When several
+        // dimensions share the smallest stride - which happens when a dimension has extent 1 and hence the
+        // same stride as its neighbour - prefer the innermost (last) dimension. Searching in reverse makes
+        // std::min_element resolve such ties to the highest index, avoiding a spurious dimension swap below
+        // that would otherwise reject valid transforms (e.g. shapes like {3,1}). See issue #631.
+        auto a_min = std::min_element(std::make_reverse_iterator(stride_vecs.vec_a.end()),
+                                      std::make_reverse_iterator(stride_vecs.vec_a.begin() + 1))
+                         .base() -
+                     1;
+        auto b_min = std::min_element(std::make_reverse_iterator(stride_vecs.vec_b.end()),
+                                      std::make_reverse_iterator(stride_vecs.vec_b.begin() + 1))
+                         .base() -
+                     1;
         if constexpr (dom == dft::domain::REAL) {
             if ((a_min != stride_vecs.vec_a.begin() + rank) ||
                 (b_min != stride_vecs.vec_b.begin() + rank)) {
@@ -180,6 +192,8 @@ public:
         }
         const int a_stride = static_cast<int>(*a_min);
         const int b_stride = static_cast<int>(*b_min);
+        // Capture the position before erasing, as the iterator is invalidated by erase.
+        const auto a_min_idx = a_min - stride_vecs.vec_a.begin();
         stride_vecs.vec_a.erase(a_min);
         stride_vecs.vec_b.erase(b_min);
         int fwd_istride = a_stride;
@@ -188,9 +202,9 @@ public:
             stride_api_choice == dft::detail::stride_api::FB_STRIDES ? b_stride : a_stride;
         int bwd_ostride =
             stride_api_choice == dft::detail::stride_api::FB_STRIDES ? a_stride : b_stride;
-        if (a_min - stride_vecs.vec_a.begin() != rank) {
+        if (a_min_idx != rank) {
             // swap dimensions to have the last one have the smallest stride
-            std::swap(n_copy[a_min - stride_vecs.vec_a.begin() - 1], n_copy[rank - 1]);
+            std::swap(n_copy[a_min_idx - 1], n_copy[rank - 1]);
         }
         for (int i = 1; i < rank; i++) {
             if ((stride_vecs.vec_a[i] % a_stride != 0) || (stride_vecs.vec_b[i] % b_stride != 0)) {


melonakos (commented):
The #631 diagnosis here is correct and the fix is a good one, but half of this PR has already landed on `develop` via a different route — which is almost certainly why it now conflicts. A rebase should shrink it to something small and easy to approve.

## What's already merged

#750, "[DFT][cuFFT] Fix use of invalidated iterator in commit.cpp", merged to `develop` and did the same thing as your second hunk. `develop` currently reads:

```cpp
// erase() invalidates a_min (and any iterator after it).
const auto a_min_pos = a_min - stride_vecs.vec_a.begin();
stride_vecs.vec_a.erase(a_min);
stride_vecs.vec_b.erase(b_min);
...
if (a_min_pos != rank) {
    std::swap(n_copy[a_min_pos - 1], n_copy[rank - 1]);
}
```

That's your `a_min_idx` change under a different name. You were right about the bug — using `a_min` after `erase(a_min)` is UB — you just got there in parallel with whoever filed #750. This PR was opened 2026-07-28 and hasn't been rebased since, so it's carrying a now-redundant copy.

## What's still genuinely needed

The #631 fix itself. `develop` still has the plain forward search:

```cpp
auto a_min = std::min_element(stride_vecs.vec_a.begin() + 1, stride_vecs.vec_a.end());
auto b_min = std::min_element(stride_vecs.vec_b.begin() + 1, stride_vecs.vec_b.end());
```

`std::min_element` returns the *first* minimum, so when two dimensions share the smallest stride — which is exactly what happens when a dimension has extent 1 — the outer index wins, and the `a_min - begin() != rank` check below then performs a spurious dimension swap and rejects a perfectly valid transform. Shape `{3,1}` is the reproducer, and your comment explaining it is clear and worth keeping.

Your reverse-iterator search resolves ties to the highest index, which is the fix. I checked the iterator arithmetic: `min_element` over `[make_reverse_iterator(end()), make_reverse_iterator(begin()+1))` walks from the last element down to `begin()+1`, and `.base() - 1` maps the result back to the forward iterator for the same element. Correct.

## So: please rebase onto `develop` and drop the duplicated hunk

That leaves just the `min_element` change plus its comment — a small, focused PR that I'd expect to go through quickly. It should also clear the conflict, and since no CI has ever run on this branch (GitHub can't build a merge commit for a conflicting PR), the rebase is what will finally get it tested.

## Two suggestions while you're in there

**Factor out the tie-breaking search.** The expression is dense and appears twice verbatim. Giving it a name makes the intent legible at the call site:

```cpp
// Smallest stride, preferring the innermost dimension on ties (see issue #631).
auto innermost_min = [](auto& v) {
    return std::min_element(std::make_reverse_iterator(v.end()),
                            std::make_reverse_iterator(v.begin() + 1)).base() - 1;
};
auto a_min = innermost_min(stride_vecs.vec_a);
auto b_min = innermost_min(stride_vecs.vec_b);
```

**Add a `{3,1}` regression test.** You've got a concrete failing shape from the issue — that's the cheapest possible test to write, and it's the thing that stops #631 coming back.

## Related

#760 changes the corresponding validity logic on the rocFFT side, and I've approved it. Between these two the DFT backends should end up much closer to each other, which is a good outcome — the rocFFT file still had a `"dft/backends/cufft"` string in one of its throws, so they'd drifted a fair way apart.

PR #748, opened 2026-07-28
[LAPACK][cuSOLVER] Fix incorrect QR results on repeated runs (#626)

### LINKED ISSUE #626: [cuSOLVER] Incorrect results in QR decomposition on CUDA
# Summary
When performing QR decomposition on CUDA device the computed matrix `Q` contains incorrect elements during multiple runs  for matrices of size `256x256` or larger.
This problem does not occur when running on Intel.

# Version
I am using the latest release version [v0.6](https://github.com/uxlfoundation/oneMath/releases/tag/v0.6) 

# Environment
* HW you use :  NVIDIA A100 80GB PCIe
* Backend library version: Build cuda_12.4.r12.4/compiler.33961263_0
* OS name and version : Linux Ubuntu 24.04 LTS 6.8.0-51-generic
* Compiler version : Intel(R) oneAPI DPC++/C++ Compiler 2025.0.4 (2025.0.4.20241205) 
```
$ sycl-ls
[opencl:cpu][opencl:0] Intel(R) OpenCL, Intel(R) Xeon(R) Platinum 8480+ OpenCL 3.0 (Build 0) [2024.18.12.0.05_160000]
[cuda:gpu][cuda:0] NVIDIA CUDA BACKEND, NVIDIA A100 80GB PCIe 8.0 [CUDA 12.4]

```

# Steps to reproduce

Building:
```
cmake .. -DCMAKE_CXX_COMPILER=icpx -DCMAKE_C_COMPILER=icx -DMKL_ROOT=${MKLROOT} -DBUILD_FUNCTIONAL_TESTS=OFF              \
    -DENABLE_CUBLAS_BACKEND=True  -DENABLE_CUSOLVER_BACKEND=True -DENABLE_CURAND_BACKEND=True -DENABLE_CUFFT_BACKEND=True

# build
cmake --build .

# install
export MKL_INTERFACE_ROOT=$(pwd)/../install
cmake --install . --prefix=${MKL_INTERFACE_ROOT}
```

Compiling:

```
icpx -fsycl geqrf_orgqr_repro.cpp -fsycl-targets=nvptx64-nvidia-cuda,spir64-unknown-unknown -I${MKL_INTERFACE_ROOT}/include
\
-L${MKL_INTERFACE_ROOT}/lib -lonemkl -Wl,-rpath,${MKL_INTERFACE_ROOT}/lib -o run_repro

```

Running:
```
./run_repro 256 # set matrix size
```

<details>
 <summary>Reproducer here</summary>

```
#include <iostream>
#include <vector>
#include <sycl/sycl.hpp>
#include "oneapi/mkl.hpp"
#include <mkl_version.h>


void run_qr_decomposition(sycl::queue& q,
                          std::vector<float>& diagonal_elements,
                          std::int64_t n)
{
    using T = float;
    const std::int64_t lda = n;

    std::vector<T> h_A(n * n, 0.0f);
    for (std::int64_t i = 0; i < n; ++i) {
        h_A[i * n + i] = static_cast<T>(2);
    }

    T* d_A = sycl::malloc_device<T>(n * n, q);
    T* d_tau = sycl::malloc_device<T>(n, q);

    auto copy_event = q.memcpy(d_A, h_A.data(), h_A.size() * sizeof(T));

    const std::int64_t scratchpad_size_geqrf =
        oneapi::mkl::lapack::geqrf_scratchpad_size<T>(q, n, n, lda);
    T* scratchpad_geqrf = sycl::malloc_device<T>(scratchpad_size_geqrf, q);

    sycl::event geqrf_event;
    try {
        geqrf_event = oneapi::mkl::lapack::geqrf(q, n, n, d_A, lda, d_tau, scratchpad_geqrf, scratchpad_size_geqrf, {copy_event});

        const std::int64_t scratchpad_size_orgqr =
            oneapi::mkl::lapack::orgqr_scratchpad_size<T>(q, n, n, n, lda);
        T* scratchpad_orgqr = sycl::malloc_device<T>(scratchpad_size_orgqr, q);

        auto orgqr_event = oneapi::mkl::lapack::orgqr(q, n, n, n, d_A, lda, d_tau, scratchpad_orgqr, scratchpad_size_orgqr, {geqrf_event});

        orgqr_event.wait();

        sycl::free(scratchpad_orgqr, q);
    } catch (sycl::exception const& e) {
        std::cout << "SYCL exception: " << e.what() << std::endl;
        return;
    } catch (oneapi::mkl::lapack::exception const& e) {
        std::cout << "MKL exception: " << e.what() << ", info: " << e.info() << std::endl;
        return;
    } catch (...) {
        std::cout << "Unknown exception occurred." << std::endl;
        return;
    }

    std::vector<T> h_Q(n * n);
    q.memcpy(h_Q.data(), d_A, h_Q.size() * sizeof(T)).wait();

    for (std::int64_t i = 0; i < n; ++i) {
        diagonal_elements.push_back(h_Q[i * n + i]);
    }

    sycl::free(d_A, q);
    sycl::free(d_tau, q);
    sycl::free(scratchpad_geqrf, q);
}

int main(int argc, char* argv[]) {
    if (argc != 2) {
        std::cout << "Usage: " << argv[0] << " <matrix size n>\n";
        return 1;
    }

    std::int64_t n = std::atoi(argv[1]);
    if (n <= 0) {
        std::cout << "Matrix size n must be a positive integer.\n";
        return 1;
    }

    std::cout << "Matrix size: " << n << "x" << n << " with diagonal elements set to 2.\n";

    sycl::queue q_cpu(sycl::cpu_selector{});
    sycl::queue q_gpu(sycl::gpu_selector{});

    std::cout << "Intel OneMKL version: " << INTEL_MKL_VERSION << std::endl;

    const auto &d_cpu = q_cpu.get_device();
    const auto &name_cpu = d_cpu.get_info<sycl::info::device::name>();
    const auto &dver_cpu = d_cpu.get_info<sycl::info::device::driver_version>();

    const auto &d_gpu = q_gpu.get_device();
    const auto &name_gpu = d_gpu.get_info<sycl::info::device::name>();
    const auto &dver_gpu = d_gpu.get_info<sycl::info::device::driver_version>();

    std::cout << "CPU info: " << "Driver: " << name_cpu << " [" << dver_cpu << " ]" << std::endl;
    std::cout << "GPU info: " << "Driver: " << name_gpu << " [" << dver_gpu << " ]" << std::endl;

    std::vector<float> diag_cpu;
    std::vector<float> diag_gpu;

    for (int i = 0; i < 2; ++i) {
        std::cout << "Run " << i + 1 << " on CPU:" << std::endl;
        run_qr_decomposition(q_cpu, diag_cpu, n);
        std::cout << "Ok" << std::endl;
    }

    for (int i = 0; i < 2; ++i) {
        std::cout << "Run " << i + 1 << " on GPU:" << std::endl;
        run_qr_decomposition(q_gpu, diag_gpu, n);
        std::cout << "Ok" << std::endl;
    }

    std::cout << "Comparing diagonal elements CPU vs GPU: \n";
    for (std::size_t i = 0; i < diag_cpu.size(); ++i) {
        if (std::abs(diag_cpu[i] - diag_gpu[i]) > 1e-6) {
            std::cout << "Run " << (i / n + 1) << ", Row: " << (i % n)
                      << " CPU: " << diag_cpu[i]
                      << " GPU: " << diag_gpu[i]
                      << " Difference: " << std::abs(diag_cpu[i] - diag_gpu[i]) << std::endl;
        }
    }

    return 0;
}
```
</details>

# Observed behavior
For simplicity, a diagonal matrix with elements set to `2` is used as input. 
After the first QR decomposition the diagonal elements of the matrix `Q` are correct (`1` as expected). 
However on later runs the diagonal elements remain set to `2` resulting in incorrect results. 
This problem is not observed on Intel devices.

Example output
```
$ ./run_repro 256 

Matrix size: 256x256 with diagonal elements set to 2.
Intel OneMKL version: 20250001
CPU info: Driver: Intel(R) Xeon(R) Platinum 8480+ [2024.18.12.0.05_160000 ]
GPU info: Driver: NVIDIA A100 80GB PCIe [CUDA 12.4 ]
Run 1 on CPU:
Ok
Run 2 on CPU:
Ok
Run 1 on GPU:
Ok
Run 2 on GPU:
Ok
Comparing diagonal elements CPU vs GPU: 
Run 2, Row: 86 CPU: 1 GPU: 2 Difference: 1
Run 2, Row: 87 CPU: 1 GPU: 2 Difference: 1
Run 2, Row: 88 CPU: 1 GPU: 2 Difference: 1
Run 2, Row: 89 CPU: 1 GPU: 2 Difference: 1
Run 2, Row: 90 CPU: 1 GPU: 2 Difference: 1
Run 2, Row: 91 CPU: 1 GPU: 2 Difference: 1
Run 2, Row: 92 CPU: 1 GPU: 2 Difference: 1
Run 2, Row: 93 CPU: 1 GPU: 2 Difference: 1
Run 2, Row: 94 CPU: 1 GPU: 2 Difference: 1
Run 2, Row: 95 CPU: 1 GPU: 2 Difference: 1
Run 2, Row: 96 CPU: 1 GPU: 2 Difference: 1
...

```

# Expected behavior
The diagonal elements of the `Q` matrix should be `1` after QR decomposition for every run regardless of the device used as the input matrix is diagonal.

zjin-lcf:
## Summary

Fixes #626.

QR decomposition (`geqrf` + `orgqr`) on the cuSOLVER backend returned incorrect results on repeated runs: for a diagonal input matrix the diagonal of `Q` should be `1`, but from the second run onwards (for `n >= 256`) some entries stayed at the input value `2`. The corruption was nondeterministic and size-dependent, and did not occur on the Intel backends.

### Root cause

The Householder-family cuSOLVER routines (`geqrf`, `orgqr`, `ormqr`, `gebrd`, `orgbr`, `orgtr`, `ormtr`, and the complex `ung*`/`unm*` variants) passed `nullptr` for cuSOLVER's `devInfo` argument and skipped the `lapack_info_check` call that every other routine (`getrf`, `potrf`, …) performs.

Beyond losing all error reporting, this also skipped the **implicit synchronization** that `lapack_info_check` performs: it reads `devInfo` back through a blocking `queue.wait()`. Without that wait, the SYCL event returned from the native-command submission could signal before the cuSOLVER kernels had actually finished, so the subsequent `memcpy` read partially-computed data — producing the observed run-2+ corruption.

### Fix

- Allocate a real `devInfo` and call `lapack_info_check` in all of these routines (both buffer and USM paths), matching the established pattern used by the working routines. This restores error checking **and** removes the race.
- Align `CusolverScopedContextHandler::get_stream` with the cuBLAS backend by returning the interop handle's native queue (`ih.get_native_queue()`) instead of the queue's default stream.

## Test plan — validated end-to-end on NVIDIA device, CUDA 13.2

Built oneMath with `-DENABLE_CUSOLVER_BACKEND=ON -DENABLE_CUBLAS_BACKEND=ON -DTARGET_DOMAINS=lapack` using an open-source DPC++ toolchain (intel/llvm nightly, CUDA backend), and ran the #626 reproducer (QR of a diagonal matrix, checking `Q` diagonals across repeated runs):

| Build | n=512, 6 runs |
|---|---|
| `develop` (unpatched) | **FAIL** — run 2+ have dozens of wrong diagonal entries |
| This PR | **PASS** — all runs, all diagonals == 1 |

Also verified `n = 256`, `512`, `1024` all pass with the fix.

### Attribution note for reviewers

I isolated the two changes on the device:

- **`devInfo` / `lapack_info_check` change only** (stream change reverted): **PASS**.
- **`get_stream` change only** (devInfo change reverted): **FAIL** — still races.

So the `devInfo`/`lapack_info_check` addition is what actually resolves #626 (via the synchronization it introduces and by matching the working routines). The `get_stream` change is included as a correctness/consistency alignment with the cuBLAS backend, not as the fix itself; happy to drop it if maintainers prefer a minimal change.

- [x] End-to-end reproduce + fix (n=256/512/1024).
- [x] `devInfo`/`lapack_info_check` counts balanced across buffer and USM paths.
- [ ] CI: cuSOLVER LAPACK functional tests.

Suggestion: add a functional test that runs `geqrf`+`orgqr` repeatedly on a known-diagonal matrix to guard against regressions.

### DIFF
--- a/src/lapack/backends/cusolver/cusolver_lapack.cpp
+++ b/src/lapack/backends/cusolver/cusolver_lapack.cpp
@@ -41,6 +41,7 @@ inline void gebrd(const char* func_name, Func func, sycl::queue& queue, std::int
     if (m < n)
         throw unimplemented("lapack", "gebrd", "cusolver gebrd does not support m < n");
 
+    sycl::buffer<int> devInfo{ 1 };
     queue.submit([&](sycl::handler& cgh) {
         auto a_acc = a.template get_access<sycl::access::mode::read_write>(cgh);
         auto d_acc = d.template get_access<sycl::access::mode::write>(cgh);
@@ -48,6 +49,7 @@ inline void gebrd(const char* func_name, Func func, sycl::queue& queue, std::int
         auto tauq_acc = tauq.template get_access<sycl::access::mode::write>(cgh);
         auto taup_acc = taup.template get_access<sycl::access::mode::write>(cgh);
         auto scratch_acc = scratchpad.template get_access<sycl::access::mode::read_write>(cgh);
+        auto devInfo_acc = devInfo.get_access<sycl::access::mode::write>(cgh);
         onemath_cusolver_host_task(cgh, queue, [=](CusolverScopedContextHandler& sc) {
             auto handle = sc.get_handle(queue);
             auto a_ = sc.get_mem<cuDataType_A*>(a_acc);
@@ -56,11 +58,13 @@ inline void gebrd(const char* func_name, Func func, sycl::queue& queue, std::int
             auto tauq_ = sc.get_mem<cuDataType_A*>(tauq_acc);
             auto taup_ = sc.get_mem<cuDataType_A*>(taup_acc);
             auto scratch_ = sc.get_mem<cuDataType_A*>(scratch_acc);
+            auto devInfo_ = sc.get_mem<int*>(devInfo_acc);
             cusolverStatus_t err;
             cusolver_native_named_func(func_name, func, err, handle, m, n, a_, lda, d_, e_, tauq_,
-                                       taup_, scratch_, scratchpad_size, nullptr);
+                                       taup_, scratch_, scratchpad_size, devInfo_);
         });
     });
+    lapack_info_check(queue, devInfo, __func__, func_name);
 }
 
 #define GEBRD_LAUNCHER(TYPE_A, TYPE_B, CUSOLVER_ROUTINE)                                    \
@@ -107,20 +111,24 @@ inline void geqrf(const char* func_name, Func func, sycl::queue& queue, std::int
                   sycl::buffer<T>& scratchpad, std::int64_t scratchpad_size) {
     using cuDataType = typename CudaEquivalentType<T>::Type;
     overflow_check(m, n, lda, scratchpad_size);
+    sycl::buffer<int> devInfo{ 1 };
     queue.submit([&](sycl::handler& cgh) {
         auto a_acc = a.template get_access<sycl::access::mode::read_write>(cgh);
         auto tau_acc = tau.template get_access<sycl::access::mode::write>(cgh);
         auto scratch_acc = scratchpad.template get_access<sycl::access::mode::read_write>(cgh);
+        auto devInfo_acc = devInfo.get_access<sycl::access::mode::write>(cgh);
         onemath_cusolver_host_task(cgh, queue, [=](CusolverScopedContextHandler& sc) {
             auto handle = sc.get_handle(queue);
             auto a_ = sc.get_mem<cuDataType*>(a_acc);
             auto tau_ = sc.get_mem<cuDataType*>(tau_acc);
             auto scratch_ = sc.get_mem<cuDataType*>(scratch_acc);
+            auto devInfo_ = sc.get_mem<int*>(devInfo_acc);
             cusolverStatus_t err;
             cusolver_native_named_func(func_name, func, err, handle, m, n, a_, lda, tau_, scratch_,
-                                       scratchpad_size, nullptr);
+                                       scratchpad_size, devInfo_);
         });
     });
+    lapack_info_check(queue, devInfo, __func__, func_name);
 }
 
 #define GEQRF_LAUNCHER(TYPE, CUSOLVER_ROUTINE)                                            \
@@ -470,20 +478,24 @@ inline void orgbr(const char* func_name, Func func, sycl::queue& queue, oneapi::
                   std::int64_t scratchpad_size) {
     using cuDataType = typename CudaEquivalentType<T>::Type;
     overflow_check(m, n, k, lda, scratchpad_size);
+    sycl::buffer<int> devInfo{ 1 };
     queue.submit([&](sycl::handler& cgh) {
         auto a_acc = a.template get_access<sycl::access::mode::read_write>(cgh);
         auto tau_acc = tau.template get_access<sycl::access::mode::read>(cgh);
         auto scratch_acc = scratchpad.template get_access<sycl::access::mode::read_write>(cgh);
+        auto devInfo_acc = devInfo.get_access<sycl::access::mode::write>(cgh);
         onemath_cusolver_host_task(cgh, queue, [=](CusolverScopedContextHandler& sc) {
             auto handle = sc.get_handle(queue);
             auto a_ = sc.get_mem<cuDataType*>(a_acc);
             auto tau_ = sc.get_mem<cuDataType*>(tau_acc);
             auto scratch_ = sc.get_mem<cuDataType*>(scratch_acc);
+            auto devInfo_ = sc.get_mem<int*>(devInfo_acc);
             cusolverStatus_t err;
             cusolver_native_named_func(func_name, func, err, handle, get_cublas_generate(vec), m, n,
-                                       k, a_, lda, tau_, scratch_, scratchpad_size, nullptr);
+                                       k, a_, lda, tau_, scratch_, scratchpad_size, devInfo_);
         });
     });
+    lapack_info_check(queue, devInfo, __func__, func_name);
 }
 
 #define ORGBR_LAUNCHER(TYPE, CUSOLVER_ROUTINE)                                                   \
@@ -505,20 +517,24 @@ inline void orgqr(const char* func_name, Func func, sycl::queue& queue, std::int
                   sycl::buffer<T>& tau, sycl::buffer<T>& scratchpad, std::int64_t scratchpad_size) {
     using cuDataType = typename CudaEquivalentType<T>::Type;
     overflow_check(m, n, k, lda, scratchpad_size);
+    sycl::buffer<int> devInfo{ 1 };
     queue.submit([&](sycl::handler& cgh) {
         auto a_acc = a.template get_access<sycl::access::mode::read_write>(cgh);
         auto tau_acc = tau.template get_access<sycl::access::mode::read>(cgh);
         auto scratch_acc = scratchpad.template get_access<sycl::access::mode::read_write>(cgh);
+        auto devInfo_acc = devInfo.get_access<sycl::access::mode::write>(cgh);
         onemath_cusolver_host_task(cgh, queue, [=](CusolverScopedContextHandler& sc) {
             auto handle = sc.get_handle(queue);
             auto a_ = sc.get_mem<cuDataType*>(a_acc);
             auto tau_ = sc.get_mem<cuDataType*>(tau_acc);
             auto scratch_ = sc.get_mem<cuDataType*>(scratch_acc);
+            auto devInfo_ = sc.get_mem<int*>(devInfo_acc);
             cusolverStatus_t err;
             cusolver_native_named_func(func_name, func, err, handle, m, n, k, a_, lda, tau_,
-                                       scratch_, scratchpad_size, nullptr);
+                                       scratch_, scratchpad_size, devInfo_);
         });
     });
+    lapack_info_check(queue, devInfo, __func__, func_name);
 }
 
 #define ORGQR_LAUNCHER(TYPE, CUSOLVER_ROUTINE)                                              \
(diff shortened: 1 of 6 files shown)


andrewtbarker (approved):
This looks fine to me.

> Suggestion: add a functional test that runs geqrf+orgqr repeatedly on a known-diagonal matrix to guard against regressions.

Can you add this test either here or in another PR?

src/lapack/backends/cusolver/cusolver_lapack.cpp
```diff
@@ -1210,6 +1258,7 @@ inline sycl::event gebrd(const char* func_name, Func func, sycl::queue& queue, s
     if (m < n)
         throw unimplemented("lapack", "gebrd", "cusolver gebrd does not support m < n");
 
+    int* devInfo = (int*)malloc_device(sizeof(int), queue);
```
sknepper: Ideally we'd check that malloc was successful (devInfo not null), to avoid the situation we're in now where we are passing a null ptr but there are issues with repeated runs.

tests/unit_tests/lapack/source/geqrf_orgqr_diagonal.cpp
```diff
@@ -0,0 +1,151 @@
...
+        std::vector<fp> Q(lda * n);
+        for (int64_t run = 0; run < num_runs && result; ++run) {
+            /* Reset the input for every run and factor it in place. */
+            host_to_device_copy(queue, A_initial.data(), A_dev, A_initial.size());
+            queue.wait_and_throw();
+
+#ifdef CALL_RT_API
+            oneapi::math::lapack::geqrf(queue, m, n, A_dev, lda, tau_dev, scratchpad_dev,
+                                        geqrf_scratchpad_size);
+            oneapi::math::lapack::orgqr(queue, m, n, k, A_dev, lda, tau_dev, scratchpad_dev,
```
sknepper: I believe orgqr should either wait for geqrf to finish, or pass in the output event from geqrf

sknepper (commented):
Thanks for the fix and the new test! A couple of suggestions

melonakos (changes requested):
Thanks for chasing this one down, Zheming — #626 has been a bad bug and I think you've found the actual root cause. But I'd like to see this restructured before it goes in, and there's a defect in a helper it leans on. Details below.

## The real fix is excellent, and it's one line

The substance of #626 is this, in `cusolver_scope_handle.cpp`:

```cpp
-    return sycl::get_native<sycl::backend::ext_oneapi_cuda>(queue);
+    return ih.get_native_queue<sycl::backend::ext_oneapi_cuda>();
```

That's the bug. cuSOLVER work was going onto the queue's default stream while the SYCL runtime tracked completion via the native-command stream, so the returned event could signal before the computation finished — which is exactly why results looked correct once and wrong on repeated runs. Your comment explaining it is clear and I'd merge that change essentially as-is.

## Please split the PR

Right now this is three unrelated changes in one 324-line diff:

1. the stream fix above (~1 line),
2. `devInfo` error-check plumbing threaded through ~14 routines,
3. a change to LAPACK test gating in `tests/unit_tests/CMakeLists.txt`.

Reviewer time is the scarce resource on this repo — that's the whole reason this queue is long. Item 1 plus your new `geqrf_orgqr_diagonal.cpp` test is a small, obviously-correct PR that fixes a real user-facing bug and could land quickly. Item 2 is a behavior change across a lot of surface area that needs more scrutiny (see below). Bundling them means the fix waits on the plumbing.

Could you pull 1 + the test into its own PR? I'll review that one promptly.

## The devInfo plumbing sits on a broken helper

Before expanding `lapack_info_check` to ~14 more USM call sites, the USM overload it calls needs fixing. On `develop`:

```cpp
inline void get_cusolver_devinfo(sycl::queue& queue, const int* devInfo,
                                 std::vector<int>& dev_info_) {
    queue.wait();
    queue.memcpy(dev_info_.data(), devInfo, sizeof(int));
}
```

Two problems:

- **The `memcpy` is asynchronous and never waited on.** The returned event is discarded, and `lapack_info_check` reads `dev_info_` immediately afterward. That's a race — the value inspected may be whatever was in the vector before the copy landed. Needs `queue.memcpy(...).wait()`.
- **`sizeof(int)` ignores `dev_info_size`.** For any call with `dev_info_size > 1` only element 0 is copied and the rest stay zero-initialized, so errors on all but the first batch entry are silently dropped.

Both predate your PR, so this isn't your bug — but this PR takes a helper that doesn't reliably work and applies it in ~14 more places, which turns a latent problem into a real one. Worth fixing the helper first, in its own commit.

## devInfo leaks on the error path in the USM routines

Every USM routine now has this shape:

```cpp
int* devInfo = create_devinfo(queue, __func__);
auto done = queue.submit(...);
lapack_info_check(queue, devInfo, __func__, func_name);
free(devInfo, queue);
return done;
```

`lapack_info_check` throws `computation_error` when info > 0 — which is its entire purpose — and when it throws, `free(devInfo, queue)` never runs. So the device allocation leaks precisely when a decomposition fails, and a caller retrying in a loop leaks every iteration.

This is the same exception-safety shape you just fixed properly in #764: cleanup on the happy path only. A small RAII holder for the devInfo allocation would fix all ~14 sites at once, the same way the pointer-mode guard did.

## The USM entry points become synchronous

`get_cusolver_devinfo` calls `queue.wait()`, so every USM routine now blocks until all queued work completes before returning its event. The USM API's contract is asynchronous — callers get an event specifically so they can pipeline. Making `geqrf` and friends synchronous to check info is a real semantic and performance change, not just an internal detail.

I don't think there's a clean way to have both eager error reporting and async USM here, which is part of why I'd rather see this discussed in its own PR. One option is checking info lazily or only in the buffer path; another is gating it behind a build option. Worth deciding deliberately.

## The test CMake change needs a maintainer decision

Removing `list(REMOVE_ITEM TEST_TARGET_DOMAINS "lapack")` means LAPACK tests now build when LAPACKE is absent, rather than being skipped. I can see why you need it — your new test doesn't compare against a reference, so it should run without LAPACKE — and I think the direction is right. But it changes the build for every configuration without LAPACKE, so it wants explicit sign-off rather than riding along in a bug-fix PR. @sknepper, worth a look.

## Two mechanical blockers

- **This PR currently conflicts with `develop`** (`mergeable: CONFLICTING`). That's why no CI has ever run on it — GitHub can't construct the merge commit, so the `pull_request` workflows never trigger. This is the answer to the CI question you raised back in August: not a broken pipeline, just a conflict. A rebase should light CI up.
- **#758 overlaps this PR entirely.** It contains these same three cuSOLVER files with identical line counts, plus the rocSOLVER side. Once this lands, please rebase #758 down to just the rocSOLVER delta. Also worth removing the committed `qr_diag_repro` binary and `qr_diag_repro.cpp` from that branch.

Happy to review the split-out stream fix as soon as it's up.

PR #750, opened 2026-08-11
[DFT][cuFFT] Fix use of invalidated iterator in commit.cpp

### LINKED ISSUE #726: Using invalidated iterator in cufft-commit.cpp
# Summary
The problem is that the `a_min` iterator is being used after it got invalidated through the `stride_vecs.vec_a.erase(a_min);` call (see: https://github.com/uxlfoundation/oneMath/blob/809b2ba34e75f11d130f39f3a09820c1d1723b67/src/dft/backends/cufft/commit.cpp#L191)

# Version
v0.9 git tag

# Environment
Windows 11.

# Steps to reproduce
Code review

# Observed behavior
Observed assertion during debugging.

# Expected behavior
Document behavior you expect.

Image

zjin-lcf:
## Summary

Fixes #726.

`stride_vecs.vec_a` / `vec_b` are `std::vector<int>`, so `std::vector::erase` invalidates the erased iterator and every iterator after it. In the cuFFT `commit.cpp` stride-handling code, `a_min` was reused *after* `stride_vecs.vec_a.erase(a_min)` to recompute its index:

```cpp
stride_vecs.vec_a.erase(a_min);           // invalidates a_min
...
if (a_min - stride_vecs.vec_a.begin() != rank) {                  // UB
    std::swap(n_copy[a_min - stride_vecs.vec_a.begin() - 1], ...); // UB
}
```

`erase` on a vector shifts elements without reallocating, so on non-checked builds `a_min - begin()` happened to still yield the original index. On MSVC debug builds (the reporter's environment) the checked-iterator machinery detects the invalidated iterator and asserts.

## Fix

Save the position of the smallest stride *before* erasing and use the saved index afterward. This is behavior-preserving relative to the previous release-build behavior and removes the undefined behavior.

## Test plan

- [x] No functional change on release builds (identical index computation, just without touching the invalidated iterator).
- [ ] CI: cuFFT DFT functional tests.

### DIFF
--- a/src/dft/backends/cufft/commit.cpp
+++ b/src/dft/backends/cufft/commit.cpp
@@ -180,6 +180,9 @@ public:
         }
         const int a_stride = static_cast<int>(*a_min);
         const int b_stride = static_cast<int>(*b_min);
+        // Record the position of the smallest stride before erasing, since
+        // erase() invalidates a_min (and any iterator after it).
+        const auto a_min_pos = a_min - stride_vecs.vec_a.begin();
         stride_vecs.vec_a.erase(a_min);
         stride_vecs.vec_b.erase(b_min);
         int fwd_istride = a_stride;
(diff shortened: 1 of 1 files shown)


sknepper (approved):
Looks great - thank you for fixing this issue, @zjin-lcf !

ndingle-arm (approved):

Issue #711, opened 2025-09-25
CMakeList doesnt support adaptivecpp as a SCYL Implementation, contrary to docs


LowPolygons:
Very simply, the provided link to build instructions takes you to a website which recommends the cmake flag

`DONEMATH_SYCL_IMPLEMENTATION=adaptivecpp` (https://uxlfoundation.github.io/oneMath/building_the_project_with_adaptivecpp.html)

This fails to compile, and checking the CMakeLists file in the current version reveals no support for the recommended 'adaptivecpp'

```
  if(ONEMATH_SYCL_IMPLEMENTATION)
    string( TOLOWER "${ONEMATH_SYCL_IMPLEMENTATION}" ONEMATH_SYCL_IMPLEMENTATION)
    if (ONEMATH_SYCL_IMPLEMENTATION STREQUAL "hipsycl")
      message(STATUS "Looking for hipSYCL")
      find_package(hipSYCL CONFIG REQUIRED)
      set(USE_ADD_SYCL_TO_TARGET_INTEGRATION true)
      set (CMAKE_CXX_STANDARD 17)
      add_library(ONEMATH::SYCL::SYCL INTERFACE IMPORTED)
    elseif(ONEMATH_SYCL_IMPLEMENTATION STREQUAL "dpc++")
      message(STATUS "Looking for dpc++")
      set(USE_ADD_SYCL_TO_TARGET_INTEGRATION false)
      find_package(Compiler REQUIRED)
    else()
      message(FATAL_ERROR "SYCL implementation ${ONEMATH_SYCL_IMPLEMENTATION} is not known")
    endif()
  else()
    message(STATUS "Looking for dpc++")
    set(USE_ADD_SYCL_TO_TARGET_INTEGRATION false)
    find_package(Compiler REQUIRED)
  endif()
```

zjin-lcf:
@LowPolygons   Can you review the PR ?  Thanks.

PR #699, opened 2025-07-11
Require AdaptiveCpp >=24.06 and rename all hipSYCL references to AdaptiveCpp

rafbiels:
# Description

This is a rebase of https://github.com/uxlfoundation/oneMath/pull/543 on top of the current `develop` branch + an update to replace the remaining occurrences of hipSYCL to AdaptiveCpp. This is also addressing the review comment rightly pointing out that oneMath already implicitly requires AdaptiveCpp version >=24.06 by calling `AdaptiveCpp_enqueue_custom_operation`, so the requirement should be made explicit.

Fixes the second of two issues in #661

# Checklist
- [x] Do all unit tests pass locally? Attach a log.
 - AdaptiveCpp with generic BLAS backend on NVIDIA GPU: [ut-acpp-generic-blas-nvidia-gpu.log.gz](https://github.com/user-attachments/files/21184039/ut-acpp-generic-blas-nvidia-gpu.log.gz)
 - AdaptiveCpp with cuBLAS, cuRAND, cuFFT backends on NVIDIA GPU: [ut-acpp-blas-rng-dft-nvidia-gpu.log.gz](https://github.com/user-attachments/files/21184042/ut-acpp-blas-rng-dft-nvidia-gpu.log.gz)
 - AdaptiveCpp with rocBLAS, rocRAND, rocFFT backends on AMD GPU: [ut-acpp-blas-rng-dft-amd-gpu.log.gz](https://github.com/user-attachments/files/21184044/ut-acpp-blas-rng-dft-amd-gpu.log.gz)

PR #751, opened 2026-08-11
[CMake] Accept legacy 'hipsycl' as deprecated alias for 'adaptivecpp'

zjin-lcf:
## Summary

Addresses #711.

hipSYCL was renamed to AdaptiveCpp, and #699 updated `ONEMATH_SYCL_IMPLEMENTATION` (and all internal references / docs) to use `adaptivecpp`. As a result, users following older instructions or scripts that pass `-DONEMATH_SYCL_IMPLEMENTATION=hipsycl` now hit a fatal error:

```
SYCL implementation hipsycl is not known
```

This PR accepts the legacy `hipsycl` value as a deprecated alias for `adaptivecpp`, emitting a deprecation warning that points users to the new value, so existing build scripts keep working:

```cmake
if (ONEMATH_SYCL_IMPLEMENTATION STREQUAL "hipsycl")
  message(WARNING "ONEMATH_SYCL_IMPLEMENTATION=hipsycl is deprecated: ... Treating it as 'adaptivecpp'. Please use -DONEMATH_SYCL_IMPLEMENTATION=adaptivecpp instead.")
  set(ONEMATH_SYCL_IMPLEMENTATION "adaptivecpp")
endif()
```

Note: the original report (v0.9) predates #699; on current `develop` the docs and CMake already agree on `adaptivecpp`. This change only adds backward compatibility for the old spelling.

## Test plan

- [x] `-DONEMATH_SYCL_IMPLEMENTATION=adaptivecpp` unchanged.
- [x] `-DONEMATH_SYCL_IMPLEMENTATION=hipsycl` (and `hipSYCL`, via existing `TOLOWER`) now maps to AdaptiveCpp with a deprecation warning instead of failing.
- [x] Unknown values still produce the existing fatal error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,6 +296,12 @@ else()
   # Find necessary packages
   if(ONEMATH_SYCL_IMPLEMENTATION)
     string( TOLOWER "${ONEMATH_SYCL_IMPLEMENTATION}" ONEMATH_SYCL_IMPLEMENTATION)
+    # hipSYCL was renamed to AdaptiveCpp. Accept the legacy value as a
+    # deprecated alias so existing build scripts keep working.
+    if (ONEMATH_SYCL_IMPLEMENTATION STREQUAL "hipsycl")
+      message(WARNING "ONEMATH_SYCL_IMPLEMENTATION=hipsycl is deprecated: hipSYCL has been renamed to AdaptiveCpp. Treating it as 'adaptivecpp'. Please use -DONEMATH_SYCL_IMPLEMENTATION=adaptivecpp instead.")
+      set(ONEMATH_SYCL_IMPLEMENTATION "adaptivecpp")
+    endif()
     if (ONEMATH_SYCL_IMPLEMENTATION STREQUAL "adaptivecpp")
       message(STATUS "Looking for AdaptiveCpp")
       find_package(AdaptiveCpp CONFIG REQUIRED)


sknepper (approved):
Looks good - it makes sense to accept the hipsycl alias to preserve backward compatibility for existing build scripts.

melonakos (approved):
Thanks Zheming — this is a straightforward usability win and I'm happy to approve it.

I checked the surrounding logic: before this change, `-DONEMATH_SYCL_IMPLEMENTATION=hipsycl` fell through to the `else()` branch and died with `FATAL_ERROR "SYCL implementation hipsycl is not known"`. So anyone with a build script written before the AdaptiveCpp rename got a hard failure with no hint about what to change. Now they get a clear warning and a working build. Good.

Placement is right — after the `TOLOWER`, before the `adaptivecpp` comparison — and I confirmed there are no other `hipsycl` references left anywhere in `CMakeLists.txt`, so nothing else needs updating to match.

One optional suggestion, take it or leave it: consider `message(DEPRECATION ...)` rather than `message(WARNING ...)`. It's CMake's purpose-built mechanism for exactly this case, and it lets downstream users opt into strictness with `-DCMAKE_ERROR_DEPRECATED=ON` or quiet it with `-DCMAKE_WARN_DEPRECATED=OFF`. With plain `WARNING` they have no control. Not worth another CI cycle on its own if you'd rather leave it.

Heads up on CI: **those four red ArmPL failures are not yours.** Every one of them comes from a run on 2026-08-11, and ArmPL passes on runs before (Jul 28) and after (Aug 12 onward) on other PRs — it was a CI outage that day. This PR touches only `CMakeLists.txt` and cannot break ArmPL BLAS/LAPACK/RNG/DFT unit tests. A re-run should clear it, so please don't spend any time chasing those traces.

With this approval you should be at the two required, so a CI re-run is the only thing standing between this and a merge.

PR #754, opened 2026-08-11
[blas] Make USM rotm param argument const

### LINKED ISSUE #592: The blas rotm interface is inconsistent with the document
The 7th parameter has the type `const T *` in the document https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onemkl/source/domains/blas/rotm#rotm-usm-version. But in the header file, it is `T *` https://github.com/oneapi-src/oneMKL/blob/develop/include/oneapi/mkl/blas.hxx#L3167

zjin-lcf:
## Summary

The oneAPI specification declares the 7th argument (`param`) of the USM version of `rotm` as `const T *` (see the [rotm spec](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onemkl/source/domains/blas/rotm#rotm-usm-version)), but the oneMath headers and backends declared it as `T *`. `param` is a read-only input containing the `flag` and the components of the modified Givens rotation matrix `H`, so it should be `const`.

This PR makes the USM `rotm` interface const-correct and consistent with the specification (the in-tree spec `docs/spec/domains/blas/rotm.rst` already uses `const T *param`). The buffer version was already correct (it uses a read accessor) and is unchanged.

Fixes #592.

## Changes

Updated the `const`-ness of `param` in the USM `rotm` overloads across every layer so the signatures stay consistent:

- Public API: `include/oneapi/math/blas.hxx`
- Backend dispatch / loader / CT headers: `onemath_blas_backends.hxx`, `blas_loader.hxx`, `blas_ct_backends.hxx`, and every backend `blas_ct.hxx`
- Backend declaration headers: `onemath_blas_cublas.hxx`, `onemath_blas_rocblas.hxx`
- Loader + function table: `src/blas/blas_loader.cpp`, `src/blas/function_table.hpp`
- Backends: `mkl_common`, `cublas`, `rocblas`, `netlib`, `openblas`, `generic`, `armpl`

The cuBLAS/rocBLAS wrappers now `reinterpret_cast` `param` to a `const` device-type pointer (the underlying `cublas*rotm`/`rocblas_*rotm` already take `const` param).

## Test plan

- [x] `clang-format` (pre-commit v19.1.0) passes.
- [x] Built the **MKL GPU** backend (BLAS) and ran the functional tests on an Intel Arc A770; `RotmTests` and `RotmUsmTests` (CT + RT, column- and row-major, single precision) **pass**. Double precision skipped (device lacks fp64).
- [x] **cuBLAS** backend library compiles cleanly against CUDA 12.1 with the const change.
- [x] **rocBLAS** backend object compiles cleanly against ROCm 6.4.4 with the const change.

### DIFF
--- a/include/oneapi/math/blas.hxx
+++ b/include/oneapi/math/blas.hxx
@@ -3102,14 +3102,14 @@ static inline sycl::event rotg(sycl::queue& queue, std::complex<double>* a, std:
 }
 
 static inline sycl::event rotm(sycl::queue& queue, std::int64_t n, float* x, std::int64_t incx,
-                               float* y, std::int64_t incy, float* param,
+                               float* y, std::int64_t incy, const float* param,
                                const std::vector<sycl::event>& dependencies = {}) {
     auto done = detail::rotm(get_device_id(queue), queue, n, x, incx, y, incy, param, dependencies);
     return done;
 }
 
 static inline sycl::event rotm(sycl::queue& queue, std::int64_t n, double* x, std::int64_t incx,
-                               double* y, std::int64_t incy, double* param,
+                               double* y, std::int64_t incy, const double* param,
                                const std::vector<sycl::event>& dependencies = {}) {
     auto done = detail::rotm(get_device_id(queue), queue, n, x, incx, y, incy, param, dependencies);
     return done;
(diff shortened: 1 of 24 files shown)


ndingle-arm (commented):
Could we add a regression test that passes param as const fp* in rotm_usm.cpp? The current test passes param.data() as a mutable pointer, so it would also compile against the old signature and doesn’t verify the issue being fixed. Using a const pointer would ensure both runtime and compile-time dispatch accept the specified API.

ndingle-arm (approved):
Thank you for making the additional change.

sknepper (approved):
Thank you for fixing this long-standing issue, @zjin-lcf ! Looks good, including the test changes.

PR #756, opened 2026-08-11
[cublas] Implement dgmm_batch backend

### LINKED ISSUE #562: oneMKL only has `dgmm_batch` variant: doesn't map to cublas
cublas does not have a specific batch variant of `dgmm`, it only has `cublas<t>dgmm()`, see https://docs.nvidia.com/cuda/cublas/#id10
However oneMKL only supports the "batch-style" `dgmm_batch` interface

This is probably one reason why the cublas backend to oneMKL has no implementation for any `dgmm` functions. I'm not sure how widely used `dgmm` is, but I'm guessing since it is a core blas algorithm it should probably be supported.

Is there a reason why there is only the batch style variant of `dgmm` in oneMKL?

Thanks

zjin-lcf:
## Summary

`dgmm_batch` was unimplemented (threw `unimplemented`) in the cuBLAS backend. As noted in #562, cuBLAS provides only the non-batched `cublas<t>dgmm` — there is **no** strided/batched `dgmm` in any CUDA library (verified across cuBLAS, cuBLASLt, cuBLASXt, cuSPARSE headers; only `cublasSdgmm/Ddgmm/Cdgmm/Zdgmm` and their `_64` twins exist). This PR implements `dgmm_batch` for the cuBLAS backend by looping over `cublas<t>dgmm`.

- Implements column-major **buffer strided**, **USM strided**, and **USM group** `dgmm_batch`
- Types: `float`, `double`, `complex<float>`, `complex<double>`
- Row-major maps to column-major by flipping `side` and swapping `m`/`n` (same approach as the rocBLAS backend)
- Handles negative `incx` via the standard BLAS start-at-end convention

## Tests

Extends the shared `dgmm_batch` tests for stronger coverage:
- strided (buffer + USM): adds `incx = 3` and `incx = -1` cases
- group (USM): adds `group_count = 1` and `group_count = 10` cases

## Test plan

Verified on an NVIDIA RTX PRO 6000 Blackwell (sm_120), CUDA 13.3 / cuBLAS 13.6, DPC++:

- [x] `test_main_blas_ct --gtest_filter=*Dgmm*` → all pass (col/row major, all 4 types)
- [x] `test_main_blas_rt --gtest_filter=*Dgmm*` → all pass
- [x] Covers `DgmmBatchStride`, `DgmmBatchStrideUsm`, `DgmmBatchUsm` (group API)

Closes #562.

### DIFF
--- a/src/blas/backends/cublas/cublas_batch.cpp
+++ b/src/blas/backends/cublas/cublas_batch.cpp
@@ -110,35 +110,50 @@ void gemv_batch(sycl::queue& queue, transpose transa, int64_t m, int64_t n,
     throw unimplemented("blas", "gemv_batch", "for column_major layout");
 }
 
-void dgmm_batch(sycl::queue& queue, side left_right, int64_t m, int64_t n,
-                sycl::buffer<float, 1>& a, int64_t lda, int64_t stride_a, sycl::buffer<float, 1>& x,
-                int64_t incx, int64_t stride_x, sycl::buffer<float, 1>& c, int64_t ldc,
-                int64_t stride_c, int64_t batch_size) {
-    throw unimplemented("blas", "dgmm_batch", "for column_major layout");
+// cuBLAS has no batched/strided variant of <t>dgmm (only cublas<t>dgmm), so
+// dgmm_batch is implemented as a loop of cublas<t>dgmm calls. See issue #562.
+template <typename Func, typename T>
+inline void dgmm_batch(const char* func_name, Func func, sycl::queue& queue, side left_right,
+                       int64_t m, int64_t n, sycl::buffer<T, 1>& a, int64_t lda, int64_t stride_a,
+                       sycl::buffer<T, 1>& x, int64_t incx, int64_t stride_x, sycl::buffer<T, 1>& c,
+                       int64_t ldc, int64_t stride_c, int64_t batch_size) {
+    using cuDataType = typename CudaEquivalentType<T>::Type;
+    overflow_check(m, n, lda, ldc, stride_a, stride_x, stride_c, batch_size);
+    queue.submit([&](sycl::handler& cgh) {
+        auto a_acc = a.template get_access<sycl::access::mode::read>(cgh);
+        auto x_acc = x.template get_access<sycl::access::mode::read>(cgh);
+        auto c_acc = c.template get_access<sycl::access::mode::read_write>(cgh);
+        onemath_cublas_host_task(cgh, [=](CublasScopedContextHandler& sc) {
+            auto handle = sc.get_handle();
+            auto a_ = sc.get_mem<cuDataType*>(a_acc);
+            auto x_ = sc.get_mem<cuDataType*>(x_acc);
+            auto c_ = sc.get_mem<cuDataType*>(c_acc);
+            cublasStatus_t err;
+            auto mode = get_cublas_side_mode(left_right);
+            for (int64_t i = 0; i < batch_size; i++) {
+                cublas_native_named_func(func_name, func, err, handle, mode, (int)m, (int)n,
+                                         a_ + i * stride_a, (int)lda, x_ + i * stride_x, (int)incx,
+                                         c_ + i * stride_c, (int)ldc);
+            }
+        });
+    });
 }
 
-void dgmm_batch(sycl::queue& queue, side left_right, int64_t m, int64_t n,
-                sycl::buffer<double, 1>& a, int64_t lda, int64_t stride_a,
-                sycl::buffer<double, 1>& x, int64_t incx, int64_t stride_x,
-                sycl::buffer<double, 1>& c, int64_t ldc, int64_t stride_c, int64_t batch_size) {
-    throw unimplemented("blas", "dgmm_batch", "for column_major layout");
-}
+#define DGMM_STRIDED_BATCH_LAUNCHER(TYPE, CUBLAS_ROUTINE)                                          \
+    void dgmm_batch(sycl::queue& queue, side left_right, int64_t m, int64_t n,                     \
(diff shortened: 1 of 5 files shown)


src/blas/backends/cublas/cublas_batch.cpp
```diff
@@ -1521,63 +1563,40 @@ sycl::event gemv_batch(sycl::queue& queue, transpose* transa, int64_t* m, int64_
...
+
+#undef DGMM_STRIDED_BATCH_LAUNCHER_USM
+
+#define DGMM_GROUP_BATCH_LAUNCHER_USM(TYPE)                                                      \
+    sycl::event dgmm_batch(sycl::queue& queue, side* left_right, int64_t* m, int64_t* n,          \
+                           const TYPE** a, int64_t* lda, const TYPE** x, int64_t* incx, TYPE** c, \
+                           int64_t* ldc, int64_t group_count, int64_t* groupsize,                 \
+                           const std::vector<sycl::event>& dependencies) {                        \
+        for (int64_t i = 0; i < group_count; i++)                                                 \
+            left_right[i] = dgmm_flip_side(left_right[i]);                                         \
```
sknepper: I'm not sure if this is allowed by the spec. Per https://github.com/uxlfoundation/oneMath/blob/develop/docs/spec/domains/blas/dgmm_batch.rst , left_right is an input parameter, and so shouldn't be modified.

sknepper (commented):
Thanks for the cublas implementation, as well as a nice fix for rocBLAS! One concern I had about writing over an input parameter. Please also check the clang-format check results and adjust the formatting as needed.

melonakos (changes requested):
The cuBLAS implementation is a welcome addition — #562 has been open a while and a loop of `cublas<t>dgmm` calls is the right answer given cuBLAS has no batched variant. But there's a correctness gap in the new path, and the rocBLAS half of this PR is a separate change that I think breaks builds you're actively supporting elsewhere. Requesting changes.

## Must fix: the new cuBLAS path has no dimension validation

`cublas<t>dgmm` takes `int` for `m`, `n`, `lda`, `incx`, `ldc`. oneMath hands it `int64_t`. The new implementation goes straight from the interface to the call with nothing in between:

```cpp
inline void dgmm_batch(const char* func_name, Func func, sycl::queue& queue, side left_right,
                       int64_t m, int64_t n, sycl::buffer<T, 1>& a, int64_t lda, ...) {
    using cuDataType = typename CudaEquivalentType<T>::Type;
        ...
        cublas_native_named_func(func_name, func, err, handle, mode, m, n,
                                 a_ + i * stride_a, lda, x_ + i * stride_x, incx,
                                 c_ + i * stride_c, ldc);
```

I grepped the whole diff: `overflow_check` appears three times and all three are **deletions**. Nothing is added. So a caller passing a dimension above `INT_MAX` gets silent truncation and either wrong results or an out-of-bounds access, where previously they got a clean `unimplemented` throw. That's a regression in behavior for those inputs even though the feature is new.

This needs the same guard every other cuBLAS entry point has:

```cpp
overflow_check(m, n, lda, ldc, incx, stride_a, stride_x, stride_c, batch_size);
```

and the equivalent per-`i` check in the group variant.

## The rocBLAS change raises the minimum ROCm to 6.3, unguarded

Separately, this PR switches rocBLAS from `rocblas_?dgmm_strided_batched` to `rocblas_?dgmm_strided_batched_64` with no version gating, and removes the `overflow_check` calls on the grounds that ILP64 makes them unnecessary.

From the rocBLAS CHANGELOG, ILP64 arrived in stages:

```
## rocBLAS 4.1.0 for ROCm 6.1  — Level 1 gets _64
## rocBLAS 4.2.0 for ROCm 6.2  — Level 2 and level 3 trsm get _64
## rocBLAS 4.3.0 for ROCm 6.3  — Level 3 and EX functions get _64
```

`dgmm` falls in that last group, so these entry points need **ROCm 6.3 or newer**. Worth confirming the exact minimum for `dgmm` specifically, but the release it landed in is not 6.2.

Two problems with that:

1. **It contradicts your own #757.** That PR exists specifically to make oneMath build against ROCm < 6.3 by handling the code-object ABI mismatch. This PR would make those same builds fail to compile at all, for want of a symbol. The two can't both be right.
2. **You already know how to do this properly.** #762 gates the rocSOLVER `_64` entry points behind `ONEMATH_ROCSOLVER_HAS_64BIT_PIVOTS` and keeps a legacy path, explicitly so the minimum ROCm doesn't move. This PR needs the same treatment — a `ROCBLAS_VERSION`-based gate with the legacy calls retained.

And note the two changes are coupled: removing `overflow_check` is only justified while the ILP64 path is in use. If you gate it, the legacy branch needs those three checks put back, or you've moved the truncation bug to the older-ROCm path.

## Please split this PR

The title is `[cublas] Implement dgmm_batch backend`, and a reviewer going by the title would not expect to find an unrelated rocBLAS ILP64 migration that removes three validation calls and moves the minimum supported ROCm. That's the kind of change that gets waved through precisely because the title says something else.

The cuBLAS implementation is ready to go once it has its `overflow_check`. I'd merge that as its own PR. The rocBLAS ILP64 migration deserves its own PR, its own title, and a version gate — and it should reference #757 so whoever reviews it thinks about the interaction.

## Smaller notes

- **`dgmm_flip_side` is duplicated verbatim** as a `static inline` in both `cublas_batch.cpp` and `rocblas_batch.cpp`. It's correct in both, and the row-major mapping (swap the side, swap m/n) is right — but it's a candidate for a shared header rather than two copies that can drift.
- **The batch loop is serial**, one `cublas<t>dgmm` launch per matrix. That's the only option available and you documented why, so no objection — but it's worth a line in the comment noting the performance characteristic for large `batch_size`, so nobody is surprised when it doesn't scale like a real batched kernel.
- Good call referencing #562 in the comment. That's exactly the context the next reader needs.

PR #753, opened 2026-08-11
[cmake] Combine -fsycl-targets across all enabled GPU backends (fixes #708)

zjin-lcf:
## Summary

Fixes #708.

`example_lapack_getrs_usm` (and other USM examples) throw a SYCL `No kernel named ... was found` exception on an NVIDIA GPU when oneMath is built with multiple GPU vendor backends enabled at once (e.g. CUDA + Intel + AMD).

### Root cause

The SYCL offload targets were set on the shared `ONEMATH::SYCL::SYCL` interface target in two places:

- `cmake/FindCompiler.cmake` used a mutually-exclusive `if/elseif`, so only one vendor's `-fsycl-targets` triple was ever selected.
- Each vendor backend's `CMakeLists.txt` (`cublas`, `rocblas`, `cusparse`) additionally appended its own `-fsycl-targets` flag.

DPC++ only honors the **last** `-fsycl-targets` flag on the command line, so in a multi-vendor build all but the last triple were silently dropped. The device images for the other vendor(s) were never generated, so their SYCL kernels (e.g. the `getrf` 32→64-bit pivot-casting `parallel_for` in the cuSOLVER backend) were missing at runtime.

### Fix

- Assemble the offload target triples for **all** enabled GPU backends once, in `cmake/FindCompiler.cmake`, and pass them together as a single comma-separated `-fsycl-targets=...` list (with per-target `-Xsycl-target-backend=<triple>` arch options).
- Preserve the Intel GPU `spir64` image when `ENABLE_MKLGPU_BACKEND` is combined with a CUDA/HIP backend.
- Remove the now-redundant per-backend `-fsycl-targets` appends from the `cublas`, `rocblas`, and `cusparse` `CMakeLists.txt`.

For a tri-vendor build this now produces e.g. `-fsycl-targets=nvptx64-nvidia-cuda,amdgcn-amd-amdhsa,spir64`, so every enabled vendor gets a device image. Single-vendor builds are unchanged.

## Test plan

- [ ] Build with `-DENABLE_MKLGPU_BACKEND` + all `CU*` + all `ROC*` backends (`-DHIP_TARGETS=gfx1100`) as in #708.
- [ ] Run `example_lapack_getrs_usm` on the NVIDIA GPU and confirm it no longer throws `No kernel named ... was found`.
- [ ] Confirm single-vendor (CUDA-only, HIP-only) builds still configure and run correctly.

PR #757, opened 2026-08-11
[cmake][docs] Handle AMD code object ABI mismatch for ROCm < 6.3

### LINKED ISSUE #687: [roc*][DPC++] Build fails for rocm version < 6.3
# Summary
While building any domain using DPC++ compiler and rocm v6.2.4 leads to the following error:
```bash
clang++: error: cannot find ROCm device library for ABI version 6, which requires ROCm 6.3 or higher; provide its path via '--rocm-path' or '--rocm-device-lib-path', or pass '-nogpulib' to build without ROCm device library
```

# Version
oneMath: 0c57e49c0edc0bb66e6cc46c88f3fc11a7a6f86d

zjin-lcf:
## Summary
Recent DPC++/LLVM defaults to AMD code object ABI version 6, which only ROCm 6.3+ supports, so building any domain against an older ROCm fails with `cannot find ROCm device library for ABI version 6` (#687).

- `cmake/FindCompiler.cmake`: auto-append `-mcode-object-version=5` to the amdgcn compile/link options when ROCm < 6.3 is detected via `.info/version` (`ROCM_PATH` / `HIPROOT` / `/opt/rocm`), skipped if the user already set the flag.
- `docs/building_the_project_with_dpcpp.rst`: document the toolchain issue and the manual `-DCMAKE_CXX_FLAGS="-mcode-object-version=5"` workaround.

## Test plan
- Reproduced the exact error targeting `amdgcn` against ROCm 6.2.4; confirmed `-mcode-object-version=5` compiles cleanly (exit 0).
- Exercised the real `FindCompiler.cmake` end-to-end: flag added for ROCm 6.2.4/6.0.3, absent for 6.3.4/7.2.4, opt-out honored when user sets the flag, safe fall-through when no version file is found.

Fixes #687

### DIFF
--- a/cmake/FindCompiler.cmake
+++ b/cmake/FindCompiler.cmake
@@ -57,6 +57,32 @@ if(is_dpcpp)
       list(APPEND UNIX_INTERFACE_LINK_OPTIONS
         -fsycl-targets=amdgcn-amd-amdhsa -Xsycl-target-backend 
 	--offload-arch=${HIP_TARGETS})
+      # Recent DPC++/LLVM defaults to AMD code object ABI version 6, which is
+      # only supported by ROCm 6.3+. When building against an older ROCm, fall
+      # back to code object version 5 so the ROCm device library can be found.
+      # See https://github.com/uxlfoundation/oneMath/issues/687
+      if(NOT CMAKE_CXX_FLAGS MATCHES "-mcode-object-version")
+        set(_onemath_rocm_version_file "")
+        foreach(_onemath_rocm_root "$ENV{ROCM_PATH}" "$ENV{HIPROOT}" "/opt/rocm")
+          if(_onemath_rocm_root AND EXISTS "${_onemath_rocm_root}/.info/version")
+            set(_onemath_rocm_version_file "${_onemath_rocm_root}/.info/version")
+            break()
+          endif()
+        endforeach()
+        if(_onemath_rocm_version_file)
+          file(READ "${_onemath_rocm_version_file}" _onemath_rocm_version)
+          string(STRIP "${_onemath_rocm_version}" _onemath_rocm_version)
+          if(_onemath_rocm_version MATCHES "^([0-9]+)\\.([0-9]+)")
+            set(_onemath_rocm_ver "${CMAKE_MATCH_1}.${CMAKE_MATCH_2}")
+            if(_onemath_rocm_ver VERSION_LESS "6.3")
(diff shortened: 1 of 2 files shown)


melonakos (commented):
Useful fix, and the documentation half is genuinely good — spelling out the actual error text (`cannot find ROCm device library for ABI version 6`) is what makes a note like this findable when someone hits it.

The auto-detection logic is sound. Probing `ROCM_PATH`, then `HIPROOT`, then `/opt/rocm` for `.info/version` covers the normal installs, and every failure path is a no-op that leaves current behavior untouched — if the version file is missing or unparseable you add no flag, which is the right way for a heuristic like this to fail. The `if(NOT CMAKE_CXX_FLAGS MATCHES "-mcode-object-version")` escape hatch is a nice touch and lines up with the workaround you documented, so a user who sets the flag by hand doesn't get it twice.

I also checked the scoping concern I had on first read: because the AMD block is an `elseif` after the CUDA block, this only fires on AMD-only builds and can't leak `-mcode-object-version=5` into an NVIDIA build. So that's fine as written.

**I'm holding off on approving only to sort out merge order with your #753.** That PR rewrites this same `if/elseif` region of `FindCompiler.cmake` (+43/-23) to combine `-fsycl-targets` across all enabled backends, and this PR adds a block *inside* the `elseif` that #753 restructures. Whichever lands second will need a rebase, and the interaction is more than textual: once #753 merges, CUDA and ROCm backends can be enabled together, at which point this block is no longer guaranteed to be AMD-only — and `-mcode-object-version=5` would end up on the shared interface options for a mixed build.

My suggestion: **land #753 first**, then rebase this one on top and re-scope the flag so it's attached only when an AMD target is actually being built. Reviewing them in the other order means reviewing the same region twice. Happy either way — I'd just rather we pick deliberately than have it decided by whoever approves first. Tell me which order you prefer and I'll review to match.

Minor, non-blocking: some distro-packaged ROCm installs don't ship `.info/version`. Your code handles that correctly by doing nothing, so this is just a note that auto-detection will silently not engage for those users — the documented manual flag covers them.

On CI: **those four red ArmPL failures are not yours.** They all come from a run on 2026-08-11, and ArmPL passes on runs before and after that date on other PRs — it was a CI outage that day. This PR adds 42 lines of cmake and docs and removes nothing, so it can't have broken ArmPL BLAS/LAPACK/RNG/DFT. A re-run should clear it, so please don't spend any time chasing those traces.

PR #759, opened 2026-08-14
[lapack][cusolver] Use the 64-bit pivot API for getrf and getrs

### LINKED ISSUE #230: [LAPACK][CUSOLVER] Inefficient conversion from 32-bit ints to 64-bit ints.
# Summary
The legacy cuSolver API only accepts 32-bit ints as it is a limit of the library. oneMKL uses 64-bit ints, so in cases with `ipiv` vectors we use sycl kernels to convert from 32-bits to 64-bits. This is inefficient as it dispatches work to the GPU and requires an additional memory allocation. This could be improved by using an inplace expansion using the already 64-bit allocated memory.

The problem was discussed on #209. See https://github.com/oneapi-src/oneMKL/pull/209#discussion_r975077263

This issue is to track the problem and its solution.

zjin-lcf:
## Summary

Fixes #230.

oneMath types `ipiv` as `int64_t`, while the legacy cuSOLVER and cuBLAS entry points take `int`. Every affected routine therefore allocated a temporary 32-bit array and ran a cast kernel around the call, and in the USM paths it also forced a `queue.wait()` to release that temporary.

cuSOLVER has accepted `int64_t` pivots since CUDA 11.0, so `getrf`, `getrs` and their batch variants now pass the user's `ipiv` straight through. That removes 10 of the 14 temporary allocations and their cast kernels.

The entry points are named `cusolverDnGetrf`/`cusolverDnGetrs` in CUDA 11.0 and were renamed to `cusolverDnXgetrf`/`cusolverDnXgetrs` in 11.1. Inline wrappers hide that difference, and CUDA 10.x keeps the previous conversion path, so no minimum-toolkit requirement changes.

`sytrf` and `getri_batch` have no 64-bit counterpart. Their pivot array is now carved out of the tail of the scratchpad rather than allocated separately, which removes the blocking wait that releasing it required. The matching `*_scratchpad_size` queries account for the tail, so callers that query the size (as the API requires) are unaffected.

Because the 64-bit API takes the dimensions as `int64_t`, `getrf` and `getrs` no longer need the `overflow_check` that rejected sizes above the 32-bit limit, so these routines now accept dimensions beyond 2^31.

### Scope note

This reduces synchronization but does not make the USM paths fully asynchronous: routines that report `info` still block in `lapack_info_check`, which is pre-existing behaviour shared by the rest of the backend and left for separate work.

### Separate commit

The first commit is an independent bug fix in `get_cusolver_devinfo`, kept separate for review. It issued an asynchronous `memcpy` and returned without waiting, while callers read the host vector immediately and then freed the device buffer, so the copy raced with both a destroyed destination and a released source. The copy was also hard-coded to a single `int` even though batched routines pass `batch_size` as the info count, leaving the remaining entries uninitialised and allowing a spurious `computation_error`. This is latent on `develop`; removing the pivot waits widened the window enough to make it reproducible.

## Test plan

Built against CUDA 13.3.1 with DPC++ and tested on an NVIDIA A100.

- [x] Functional LAPACK tests, both compile-time and run-time dispatch: 328 passed, 0 failed, 70 skipped, matching the `develop` baseline. Covers `getrf`, `getrs`, `getri`, `sytrf` and the `getrf_batch`, `getrs_batch`, `getri_batch` variants for all four types in both buffer and USM forms.
- [x] Repeated full-suite runs (6x compile-time, plus run-time dispatch) to rule out intermittency.
- [x] The devinfo fix was validated against a flaky `Unmtr`/`Ormtr` dependency failure: 2 failures in 60 focused runs before the fix, 0 in 60 after, versus 0 in 60 on the unmodified baseline.
- [x] Version-guarded fallbacks compile-checked against real headers for CUDA 10.2 (cuSOLVER 10.3.0, legacy conversion path), 11.0.3 (10.6.0, non-X names via wrappers), 11.1.1 (11.0.1, X names) and 12.0.1.
- [x] `clang-format` clean against the repository's `_clang-format`.

### DIFF
--- a/src/lapack/backends/cusolver/cusolver_batch.cpp
+++ b/src/lapack/backends/cusolver/cusolver_batch.cpp
@@ -184,6 +184,66 @@ GETRI_STRIDED_BATCH_LAUNCHER(std::complex<double>, cublasZgetriBatched)
 
 #undef GETRI_STRIDED_BATCH_LAUNCHER
 
+#if ONEMATH_CUSOLVER_HAS_64BIT_PIVOTS
+
+template <typename T>
+inline void getrs_batch(const char* func_name, sycl::queue& queue, oneapi::math::transpose trans,
+                        std::int64_t n, std::int64_t nrhs, sycl::buffer<T>& a, std::int64_t lda,
+                        std::int64_t stride_a, sycl::buffer<std::int64_t>& ipiv,
+                        std::int64_t stride_ipiv, sycl::buffer<T>& b, std::int64_t ldb,
+                        std::int64_t stride_b, std::int64_t batch_size, sycl::buffer<T>& scratchpad,
+                        std::int64_t scratchpad_size) {
+    using cuDataType = typename CudaEquivalentType<T>::Type;
+
+    queue.submit([&](sycl::handler& cgh) {
+        auto a_acc = a.template get_access<sycl::access::mode::read>(cgh);
+        auto ipiv_acc = ipiv.template get_access<sycl::access::mode::read>(cgh);
+        auto b_acc = b.template get_access<sycl::access::mode::write>(cgh);
+
+        onemath_cusolver_host_task(cgh, queue, [=](CusolverScopedContextHandler& sc) {
+            auto handle = sc.get_handle(queue);
+            auto a_ = sc.get_mem<cuDataType*>(a_acc);
+            auto ipiv_ = sc.get_mem<std::int64_t*>(ipiv_acc);
+            auto b_ = sc.get_mem<cuDataType*>(b_acc);
+            cusolverStatus_t err;
+            CusolverDnParams params;
+            constexpr cudaDataType data_type = CudaDataType<T>::Type;
+
+            // Does not use scratch so call cuSolver asynchronously and sync at end
+            for (int64_t i = 0; i < batch_size; ++i) {
+                CUSOLVER_ERROR_FUNC_T(func_name, cusolverDnXgetrs, err, handle, params.get(),
+                                      get_cublas_operation(trans), n, nrhs, data_type,
+                                      a_ + stride_a * i, lda, ipiv_ + stride_ipiv * i, data_type,
+                                      b_ + stride_b * i, ldb, nullptr);
+            }
+#ifndef SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND
+            CUSOLVER_SYNC(err, handle)
+#endif
+        });
+    });
+}
+
+#define GETRS_STRIDED_BATCH_LAUNCHER(TYPE)                                                    \
+    void getrs_batch(sycl::queue& queue, oneapi::math::transpose trans, std::int64_t n,       \
+                     std::int64_t nrhs, sycl::buffer<TYPE>& a, std::int64_t lda,              \
+                     std::int64_t stride_a, sycl::buffer<std::int64_t>& ipiv,                 \
+                     std::int64_t stride_ipiv, sycl::buffer<TYPE>& b, std::int64_t ldb,       \
+                     std::int64_t stride_b, std::int64_t batch_size,                          \
+                     sycl::buffer<TYPE>& scratchpad, std::int64_t scratchpad_size) {          \
+        return getrs_batch<TYPE>("cusolverDnXgetrs", queue, trans, n, nrhs, a, lda, stride_a, \
+                                 ipiv, stride_ipiv, b, ldb, stride_b, batch_size, scratchpad, \
+                                 scratchpad_size);                                            \
+    }
+
+GETRS_STRIDED_BATCH_LAUNCHER(float)
+GETRS_STRIDED_BATCH_LAUNCHER(double)
+GETRS_STRIDED_BATCH_LAUNCHER(std::complex<float>)
+GETRS_STRIDED_BATCH_LAUNCHER(std::complex<double>)
+
+#undef GETRS_STRIDED_BATCH_LAUNCHER
+
+#else // no 64-bit pivot API: narrow the pivots into a temporary 32-bit array
+
 template <typename Func, typename T>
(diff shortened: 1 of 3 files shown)


melonakos (commented):
This answers the question I asked on #762 about why this PR is four times the size, and the answer is good — it isn't scope creep. Withdrawing that implication.

## Why this is legitimately bigger than #762

Two reasons, both inherent to cuSOLVER rather than to your approach:

1. **cuSOLVER's 64-bit pivot API has a three-way version history.** The entry points appeared in CUDA 11.0 (`CUSOLVER_VERSION` 10600) as `cusolverDnGetrf`/`cusolverDnGetrs`, were renamed with the `X` prefix in CUDA 11.1 (11000) with the old names deprecated, removed outright in CUDA 12, and don't exist at all in CUDA 10.x. rocSOLVER just added `*_64` names in 3.26 and left it at that. Hence the compatibility shims and the second `HAS_X_NAMES` macro, which have no rocSOLVER counterpart.
2. **cuSOLVER's 64-bit API is type-erased.** It takes `cudaDataType` tags rather than having per-type function names, so you need the `CudaDataType<T>` trait. rocSOLVER's typed `rocsolver_sgetrf_64`/`dgetrf_64`/… names mean #762 needs nothing equivalent.

And a third reason that isn't inherent — see below.

## You fixed the helper I complained about in #748

This is the part I most wanted to flag:

```cpp
-    queue.memcpy(dev_info_.data(), devInfo, sizeof(int));
+    // The copy must complete before dev_info_ is read: callers inspect it as soon as this
+    // returns and then release devInfo, both of which race with an outstanding copy.
+    queue.memcpy(dev_info_.data(), devInfo, sizeof(int) * dev_info_.size()).wait();
```

That closes both defects I raised on #748 — the discarded copy event, and the hardcoded `sizeof(int)` that silently dropped every batch entry past the first. Your comment identifies exactly the same two hazards. Good.

It also matters more here than in #748, because this PR is the first code to call `lapack_info_check` with `dev_info_size = batch_size`. On the old helper, `getrf_batch` would have checked matrix 0 and silently ignored failures in matrices 1..N−1. So the fix isn't incidental tidying, it's a prerequisite for the batched paths you're adding.

**Which means the dependency runs opposite to what I told you on #748.** I said the plumbing there sits on a broken helper and the helper should be fixed first in its own commit. You've fixed it — just in this PR instead. Cleanest resolution: pull the `get_cusolver_devinfo` fix out into its own small PR that both #748 and this one build on. It's a two-line change with a clear rationale, it would go through review in minutes, and it stops the fix being hostage to whichever of these two larger PRs lands first.

## Still outstanding: devInfo leaks on the error path

Same shape as #748 and #762, at three sites here:

```cpp
lapack_info_check(queue, devInfo, __func__, func_name, batch_size);
sycl::free(devInfo, queue);
```

`lapack_info_check` throws `computation_error` on info > 0 — a singular matrix, entirely ordinary — and the `free` never runs. In the batched paths the allocation is `sizeof(int) * batch_size`, so the leak scales with batch size.

I noticed you deferred one of the frees into a host task:

```cpp
cgh.host_task([=](sycl::interop_handle ih) { sycl::free(devInfo, queue); });
```

That's better for the asynchrony, but it doesn't help the throw path either. One RAII holder for the devInfo allocation would fix all three of these PRs at once, the same way the pointer-mode guard in #764 fixed thirteen call sites. I'd rather see that than the same fix written three times.

## Coverage asymmetry with #762 — needs a decision

The third reason this PR is larger: it migrates `getrf_batch` and `getrs_batch` as well, and #762 doesn't touch `rocsolver_batch.cpp` at all.

So if both land as-is, cuSOLVER's batched factorisations use native 64-bit pivots while rocSOLVER's still round-trip through a 32-bit temporary — different code paths, different scratchpad requirements, and different behaviour at the int32 boundary depending on which backend is loaded. That's the kind of divergence that produces a confusing bug report a year from now.

Either extend #762 to cover the batched entry points, or file a follow-up issue so it's tracked rather than forgotten. Your call which; I'd just like it to be deliberate.

## Question for the maintainers: what is the actual CUDA floor?

This determines how much of the PR is necessary, so it's worth settling before merge.

- `cmake/FindcuSOLVER.cmake`: `find_package(CUDA 10.0 REQUIRED)`
- `src/sparse_blas/backends/cusparse/CMakeLists.txt`: `find_package(CUDA 12.2 REQUIRED)`

Taken at face value the 10.0 floor justifies everything here: CUDA 10.x needs the legacy conversion path, and CUDA 11.0 exactly needs the pre-`X` shims. But the project clearly isn't uniformly targeting CUDA 10 if another backend demands 12.2.

If the real supported floor is CUDA 12, then `ONEMATH_CUSOLVER_HAS_X_NAMES` is always true, `ONEMATH_CUSOLVER_HAS_64BIT_PIVOTS` is always true, and both the shims and the entire `#else` conversion path are unreachable code — this PR would shrink substantially and get much easier to review.

@sknepper — this is really a project-policy question rather than something Zheming should have to guess at. Worth answering once and writing down, since the same question will come up on every backend PR.

## Minor

**The shims shadow vendor symbol names.** When `!HAS_X_NAMES`, you define `cusolverDnXgetrf_bufferSize` etc. in the global namespace. It's guarded by version so it can't collide in practice, and released CUDA versions don't change — but a reader can't tell at a glance whether they're looking at NVIDIA's function or yours. Something like `onemath_cusolver_xgetrf_bufferSize` would be unambiguous and costs nothing.

**`pivot32_scratchpad_size` looks right** — `(count * sizeof(int) + sizeof(T) - 1) / sizeof(T)` is the correct ceiling conversion from bytes to `T` elements, and the alignment note above it is the right thing to have thought about.

**Good that you flagged the blocking behaviour** in the "this routine still blocks in `lapack_info_check` below" comment. That's the same async-contract concern I raised on #748, and having it visible in the code is better than leaving it implicit.

PR #762, opened 2026-08-14
[lapack][rocsolver] Use the 64-bit pivot API for getrf and getrs

zjin-lcf:
## Summary

The rocSOLVER counterpart of #759, which does the same for cuSOLVER (#230).

oneMath types `ipiv` as `int64_t`, while the legacy rocSOLVER entry points take `rocblas_int`. `getrf` and `getrs` therefore each allocated a temporary 32-bit array and ran a cast kernel around the call, and in the USM `getrs` that temporary also forced a `queue.wait()` before it could be released.

rocSOLVER has accepted `int64_t` pivots since 3.26 (ROCm 6.2), so both routines now pass the user's `ipiv` straight through. That removes four temporary allocations, their cast kernels and the blocking wait. The 64-bit `getrf` reports `info` as `int64_t` rather than `int`, so `get_rocsolver_devinfo` and `lapack_info_check` are templated on the info type. `ONEMATH_ROCSOLVER_HAS_64BIT_PIVOTS` keeps the previous conversion path for rocSOLVER older than 3.26, so the minimum supported ROCm is unchanged.

Because the 64-bit entry points take the dimensions as `int64_t`, `getrf` and `getrs` no longer need the `overflow_check` that rejected sizes above the 32-bit limit.

### Scope, and how this differs from #759

rocSOLVER has no `sytrf_64`, and `getri` and the `getrf_batch`, `getrs_batch` and `getri_batch` variants are `unimplemented` in this backend, so `getrf` and `getrs` are the only routines affected. There is nothing here matching #759's scratchpad-tail change, and `sytrf` keeps its existing conversion path.

As in #759, this reduces synchronization without making the USM paths fully asynchronous: `getrf` still blocks in `lapack_info_check`, which is pre-existing behaviour shared by the rest of the backend and left for separate work.

### Upstream rocBLAS dependency above 2^28 rows

The API now accepts dimensions past 2^31, but actually factorizing a matrix with more than 2^28 rows also needs the rocBLAS fix from ROCm/rocm-libraries#10831, which I filed while testing this. `rocblas_internal_scal_launcher_64` advances the alpha scalar pointer once per n-chunk, so `getrf_64` either faults with `hipErrorIllegalAddress` or returns a silently wrong factorization for any column longer than 2^28.

This is not a regression introduced here: it affects every rocSOLVER release that has the 64-bit API (verified on 6.2.4, 6.3.4, 6.4.4, 7.0.1, 7.1.1, 7.2.0 and 7.2.4), the fix is one line in rocBLAS, and nothing at or below 2^28 rows is affected. Reviewers should just know that the "dimensions beyond 2^31" part of this change is not usable until that fix lands. I am happy to gate the 64-bit path on a fixed rocBLAS version, or to cap rows at 2^28 with the legacy path above it, if you would prefer this not to advertise a size range that upstream cannot yet deliver.

## Test plan

Built against ROCm 7.2.4 with DPC++ and tested on an AMD Instinct MI300A (gfx942).

- [x] Functional LAPACK tests, both compile-time and run-time dispatch: 204 passed, 0 failed, 364 skipped, with identical results in both modes. Baseline `develop` on the same machine gives 180 passed, 0 failed, 364 skipped. The skips are the batch variants this backend does not implement.
- [x] The 24 newly passing tests are the `Getrf` and `Getrs` accuracy and dependency suites for all four types in buffer and USM form. To be precise about why they were not passing before: this DPC++ build trips a pre-existing assertion (`It != m_DeviceKernelInfoMap.end()`) on kernels compiled into a backend shared library, which the legacy path's cast kernels hit and the 64-bit path avoids by launching no kernels at all. On a toolchain without that problem both paths pass, so this is not a correctness change — the routines are simply now exercised locally.
- [x] Three `Sytrf` suites still abort from that same pre-existing assertion, unchanged from baseline; `sytrf` is untouched by this PR.
- [x] `int64_t` pivots verified end to end at `m = 2^31 + 1024` (8 GiB, `n = 1`, largest entry at row `m - 512`): `ipiv[0] = 2147484161`, which is 514 past `INT32_MAX` and would have truncated to `-2147483135` in the old 32-bit array. `a[0]` and the scaled column are correct. This run needs ROCm/rocm-libraries#10831; I validated it with that one-line fix applied at runtime through an `LD_PRELOAD` shim, cross-checked by confirming the shim also makes a standalone rocBLAS reproducer of the bug pass.
- [x] Version-guarded fallback compile-checked against real headers for rocSOLVER 3.24 (ROCm 6.0.3, legacy conversion path) and 3.26 through 3.32 (ROCm 6.2.4, 6.3.4, 6.4.4, 7.0.1, 7.1.1, 7.2.0, 7.2.4).
- [x] `clang-format` clean against the repository's `_clang-format`.

### DIFF
--- a/src/lapack/backends/rocsolver/rocsolver_helper.hpp
+++ b/src/lapack/backends/rocsolver/rocsolver_helper.hpp
@@ -255,16 +255,33 @@ struct RocmEquivalentType<std::complex<double>> {
     using Type = rocblas_double_complex;
 };
 
+/* 64-bit pivot API */
+
+#if !defined(ROCSOLVER_VERSION_MAJOR)
+#define ONEMATH_ROCSOLVER_VERSION 0
+#else
+#define ONEMATH_ROCSOLVER_VERSION (ROCSOLVER_VERSION_MAJOR * 100 + ROCSOLVER_VERSION_MINOR)
+#endif
+
+// rocSOLVER exposes getrf and getrs entry points taking int64_t pivots, which lets oneMath hand
+// the user's ipiv array straight through instead of converting it. They were added in rocSOLVER
+// 3.26 (ROCm 6.2); earlier versions keep the conversion path, so the minimum supported ROCm is
+// unchanged.
+#define ONEMATH_ROCSOLVER_HAS_64BIT_PIVOTS (ONEMATH_ROCSOLVER_VERSION >= 326)
+
 /* devinfo */
 
-inline int get_rocsolver_devinfo(sycl::queue& queue, sycl::buffer<int>& devInfo) {
-    sycl::host_accessor<int, 1, sycl::access::mode::read> dev_info_{ devInfo };
+// The 64-bit entry points report info as int64_t, the legacy ones as int.
+template <typename INFO_T>
+inline INFO_T get_rocsolver_devinfo(sycl::queue& queue, sycl::buffer<INFO_T>& devInfo) {
+    sycl::host_accessor<INFO_T, 1, sycl::access::mode::read> dev_info_{ devInfo };
     return dev_info_[0];
 }
 
-inline int get_rocsolver_devinfo(sycl::queue& queue, const int* devInfo) {
-    int dev_info_;
-    queue.memcpy(&dev_info_, devInfo, sizeof(int));
+template <typename INFO_T>
+inline INFO_T get_rocsolver_devinfo(sycl::queue& queue, const INFO_T* devInfo) {
+    INFO_T dev_info_;
+    queue.memcpy(&dev_info_, devInfo, sizeof(INFO_T));
     queue.wait();
     return dev_info_;
 }
@@ -273,7 +290,7 @@ template <typename DEVINFO_T>
 inline void lapack_info_check(sycl::queue& queue, DEVINFO_T devinfo, const char* func_name,
                               const char* cufunc_name) {
     queue.wait();
-    const int devinfo_ = get_rocsolver_devinfo(queue, devinfo);
+    const auto devinfo_ = get_rocsolver_devinfo(queue, devinfo);
     if (devinfo_ > 0)
         throw oneapi::math::lapack::computation_error(
             func_name, std::string(cufunc_name) + " failed with info = " + std::to_string(devinfo_),
--- a/src/lapack/backends/rocsolver/rocsolver_lapack.cpp
+++ b/src/lapack/backends/rocsolver/rocsolver_lapack.cpp
@@ -132,6 +132,49 @@ GEQRF_LAUNCHER(std::complex<double>, rocsolver_zgeqrf)
 
 #undef GEQRF_LAUNCHER
 
+#if ONEMATH_ROCSOLVER_HAS_64BIT_PIVOTS
+
+template <typename Func, typename T>
+void getrf(const char* func_name, Func func, sycl::queue& queue, std::int64_t m, std::int64_t n,
+           sycl::buffer<T>& a, std::int64_t lda, sycl::buffer<std::int64_t>& ipiv,
+           sycl::buffer<T>& scratchpad, std::int64_t scratchpad_size) {
+    using rocmDataType = typename RocmEquivalentType<T>::Type;
+    sycl::buffer<std::int64_t> devInfo{ 1 };
+
+    queue.submit([&](sycl::handler& cgh) {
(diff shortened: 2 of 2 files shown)


melonakos (commented):
This is well-researched work and the structure is right. I verified the parts that would be easy to get wrong and they're all correct. One real defect to fix and I'll approve.

## Verified correct

**The version gate actually works.** This pattern usually fails silently, so I checked it:

- `rocsolver_helper.hpp` includes `<rocsolver/rocsolver.h>` at line 31, and the macro block is at line 260 — right order.
- Upstream, `rocsolver.h` includes `rocsolver-version.h`, so `ROCSOLVER_VERSION_MAJOR` is genuinely defined at the point of use. I checked this because the rocFFT code in this repo includes `<rocfft-version.h>` explicitly, which made me suspect ROCm didn't pull version headers in automatically. For rocSOLVER it does.

**The 3.26 threshold is exactly right.** From the rocSOLVER CHANGELOG, under `## rocSOLVER 3.26.0 for ROCm 6.2.0`:

```
* 64-bit APIs for existing functions:
    * GETF2_64 (with batched and strided_batched versions)
    * GETRF_64 (with batched and strided_batched versions)
    * GETRS_64 (with batched and strided_batched versions)
```

Both `getrf` and `getrs` landed in that release, so `>= 326` is correct for both, and your claim that the minimum supported ROCm is unchanged holds because the legacy path is retained. Nice attention to detail — this is the kind of constant that silently breaks builds if it's off by one release.

## Must fix: `devInfo` leaks on the error path

In the USM `getrf`:

```cpp
std::int64_t* devInfo = (std::int64_t*)malloc_device(sizeof(std::int64_t), queue);
auto done = queue.submit(...);
// lapack_info_check calls queue.wait()
lapack_info_check(queue, devInfo, __func__, func_name);
free(devInfo, queue);
return done;
```

`lapack_info_check` throws `computation_error` whenever info > 0 — a singular matrix, which is an ordinary outcome for `getrf`, not an exotic one. When it throws, `free(devInfo, queue)` never runs. A caller factorising a batch of matrices in a loop leaks one allocation per singular matrix.

Two things worth noting:

- This is the same shape you fixed correctly in #764 with the pointer-mode guard: cleanup on the happy path only. A small RAII holder here fixes it, and would fix the identical leak in #748 too.
- `malloc_device`'s return isn't checked for null, whereas in #748 you added `create_devinfo` precisely to throw `device_bad_alloc` on that. Worth using the same approach in both PRs rather than having them diverge.

Fix that and I'll approve.

## Suggestions, not blockers

**Make silent fallback impossible.** The gate is correct today, but if the include is ever reordered or someone builds against a rocSOLVER packaging without the version header, `ONEMATH_ROCSOLVER_VERSION` quietly becomes 0, everyone gets the slow conversion path, and nothing says so. Since `rocsolver.h` is included unconditionally, a missing `ROCSOLVER_VERSION_MAJOR` means something is genuinely wrong — so I'd make it loud:

```cpp
#if !defined(ROCSOLVER_VERSION_MAJOR)
#error "ROCSOLVER_VERSION_MAJOR is not defined; rocsolver-version.h was not pulled in"
#endif
```

Cheap insurance against the failure mode where this PR becomes a no-op without anyone noticing.

**Plan the legacy path's retirement.** The `#if/#else` duplicates all of `getrf` and `getrs` across buffer and USM. Once builds standardise on ROCm ≥ 6.2, the `#else` branch stops being compiled anywhere and will rot — and I doubt either branch is exercised in CI today, since there's no AMD runner. Worth a tracking issue to delete it when the minimum ROCm rises, so it doesn't sit there indefinitely as untested code.

**Scratchpad may now be over-sized.** The 64-bit path no longer needs the temporary int32 pivot array, but `getrf` still takes `scratchpad`/`scratchpad_size` and ignores them. If `getrf_scratchpad_size` includes room for that temporary, it now over-reports under the 64-bit path — harmless for correctness, but it's a user-visible allocation size, and you may also pick up unused-parameter warnings.

## On the pair

This is the ROCm half of #759. The two are strikingly asymmetric — #762 is +203/-6 across 2 files, #759 is +890/-27 across 3, with +479 in `cusolver_batch.cpp` that has no counterpart here. I'd expected the pair-review shortcut to apply, and it doesn't. Could you say what #759 is doing that this isn't? If it's extending the batched entry points too, that's a genuine gap on the ROCm side; if it's incidental scope, #759 may want trimming to match this one. Either way I'd rather know before reviewing #759 in detail.

Issue #486, opened 2024-05-14
[BLAS] rocBLAS backend test can fail with HSA_STATUS_ERROR_INVALID_PACKET_FORMAT


Rbiessy:
# Summary
From the CI, the test `BLAS/CT/RotTestSuite/RotTests.RealSinglePrecision/Column_Major_AMD_Instinct_MI210` can fail with the rocBLAS backend. The error does not seem deterministic.

# Version
Tested with the HEAD of develop branch: https://github.com/oneapi-src/oneMKL/commit/99e85be9bb31fe80ff01ab267f3b0c679d6406b2

# Environment
CI build with AdaptiveCpp enabled and the rocBLAS backend on AMD.

# Observed behavior
Runtime error:
```
:0:rocdevice.cpp            :2788: 73916926068 us: 38926: [tid:0x7f1f5346f640] Callback: Queue 0x7f1e43e00000 aborting with error : HSA_STATUS_ERROR_INVALID_PACKET_FORMAT: The AQL packet is malformed. code: 0x1009
```

PR #764, opened 2026-08-15
[blas][rocblas] Restore the pointer mode when a rocBLAS call throws

zjin-lcf:
## Summary

Fixes the pointer-mode leak that can produce the intermittent `RotTests.RealSinglePrecision` failure reported in #486.

The reduction routines (`asum`, `nrm2`, `dot`, `dotc`, `dotu`, `rotg`, `rotm`, `rotmg`, `iamax`, `iamin`) switch the rocBLAS handle to device pointer mode and switch it back after the call:

```cpp
rocblas_set_pointer_mode(handle, rocblas_pointer_mode_device);
...
rocblas_native_func(func, err, handle, n, x_, std::abs(incx), res_);
rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host);
```

`rocblas_native_func` throws a `rocblas_error` on any status other than `rocblas_status_success`, so the reset is skipped. Handles are cached in thread-local storage and reused, so the handle stays in device pointer mode for every subsequent call on that thread.

Routines that pass a **host** address for their scalar arguments then have that address dereferenced on the device. `rot` is the most exposed one — it forwards `(rocDataType2*)&c` and `(rocDataType3*)&s` and never sets the pointer mode itself, relying entirely on the handle already being in host mode. On a discrete GPU, where host memory is not device-addressable, that is an illegal address.

This replaces the manual set/reset pairs at all 13 sites with an RAII guard, so the mode is restored during stack unwinding as well as on the normal path. The guard restores the *previous* mode rather than unconditionally forcing host mode, so it is safe to nest.

## Validation

On an AMD Instinct MI210 (gfx90a, ROCm 7.1.1, DPC++), provoking a failure inside a guarded region and then calling `rot`:

| | crashes |
|---|---|
| without the guard | 10 / 10 |
| with the guard | 0 / 10 |

Without the guard the `rot` call fails with `hipErrorIllegalAddress` (700) and leaves the queue unrecoverable. With it, `rot` returns correct results and a following `asum` is correct. Both arms take the identical first exception on the same queue, so the guard is the only difference.

Reproducer:

```cpp
// step 1: provoke a failure inside a pointer-mode-guarded region
try {
    oneapi::math::blas::column_major::asum(q, n, x, 1, (float*)nullptr).wait();
} catch (const std::exception& e) { /* expected */ }

// step 2: rot passes &c and &s, which are host addresses
const float c = 1.0f, s = 0.0f;          // identity rotation
oneapi::math::blas::column_major::rot(q, n, x, 1, y, 1, c, s).wait();
```

Also ran the BLAS test suite on the same MI210 with the fix applied: 740 tests pass with no failures.

## Caveats

- The reproducer forces the initial rocBLAS failure with a null result pointer. This change explains the cascade *given* that some earlier reduction call failed; it does not identify what failed first in the CI run behind #486.
- #486 was reported with AdaptiveCpp, which surfaces the fault as `HSA_STATUS_ERROR_INVALID_PACKET_FORMAT`. I could only verify with DPC++, which reports the same underlying invalid device address as `hipErrorIllegalAddress`. AdaptiveCpp was not available on the machine used.

## Test plan

- [x] Builds against `develop` with the rocBLAS backend (gfx90a, ROCm 7.1.1)
- [x] Deterministic MI210 reproducer passes with the fix and fails without it
- [x] BLAS functional tests pass on MI210 (740 passed, 0 failed)
- [ ] Confirmation on MI210 with AdaptiveCpp, matching the exact error in #486

### DIFF
--- a/src/blas/backends/rocblas/rocblas_helper.hpp
+++ b/src/blas/backends/rocblas/rocblas_helper.hpp
@@ -173,6 +173,30 @@ public:
     hipError_t hip_err;                                                    \
     HIP_ERROR_FUNC(hipStreamSynchronize, hip_err, currentStreamId);
 
+// Sets the pointer mode of a rocBLAS handle and restores the previous mode on scope
+// exit. rocBLAS handles are cached and reused across calls, so an unwound scope must
+// not leave the handle in device mode: later routines pass host addresses for their
+// scalar arguments and rocBLAS would dereference them on the device.
+class rocblas_pointer_mode_guard {
+    rocblas_handle handle_;
+    rocblas_pointer_mode previous_;
+
+public:
+    rocblas_pointer_mode_guard(rocblas_handle handle, rocblas_pointer_mode mode)
+            : handle_(handle),
+              previous_(rocblas_pointer_mode_host) {
+        rocblas_get_pointer_mode(handle_, &previous_);
+        rocblas_set_pointer_mode(handle_, mode);
+    }
+
+    ~rocblas_pointer_mode_guard() {
+        rocblas_set_pointer_mode(handle_, previous_);
+    }
+
+    rocblas_pointer_mode_guard(const rocblas_pointer_mode_guard&) = delete;
+    rocblas_pointer_mode_guard& operator=(const rocblas_pointer_mode_guard&) = delete;
+};
+
 template <class Func, class... Types>
 inline void rocblas_native_func(Func func, rocblas_status err, rocblas_handle handle,
                                 Types... args) {
--- a/src/blas/backends/rocblas/rocblas_helper.hpp
+++ b/src/blas/backends/rocblas/rocblas_helper.hpp
@@ -177,6 +177,13 @@ public:
 // exit. rocBLAS handles are cached and reused across calls, so an unwound scope must
 // not leave the handle in device mode: later routines pass host addresses for their
 // scalar arguments and rocBLAS would dereference them on the device.
+//
+// Note this restores whatever mode was in effect on entry, rather than unconditionally
+// forcing host mode as the manual set/reset pairs it replaces used to. That is correct
+// as long as every call site goes through this guard, but it does drop the old code's
+// incidental self-healing: if some other, unguarded path ever left a handle in device
+// mode, the previous code would have normalized it back to host on the next guarded
+// call, whereas this guard simply propagates the leaked mode.
 class rocblas_pointer_mode_guard {
     rocblas_handle handle_;
     rocblas_pointer_mode previous_;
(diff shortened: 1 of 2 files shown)


melonakos (approved):
This is a real bug fix and a genuine net simplification — +37/-53 while making the code *more* correct is the right direction. Approving.

The problem is well identified. rocBLAS handles are cached and reused across calls, so if `rocblas_native_func` threw between the `set_pointer_mode(device)` and the manual `set_pointer_mode(host)` at the bottom, the handle stayed in device mode for every subsequent caller — and the next routine passing a host address for a scalar argument would have rocBLAS dereference it on the device. That's a nasty, action-at-a-distance failure mode, and the old comment block ("we need to reset this to the default value in order to avoid invalid memory accesses") shows the hazard was known but only handled on the happy path. RAII is exactly the right fix.

I verified the conversion is **complete** — I checked every remaining `rocblas_set_pointer_mode` occurrence in `rocblas_level1.cpp` at your head commit, and all 11 hits are inside comments (plus one long-dead commented-out call in `rot`). No live call site was left on the old manual pattern, and `level2`/`level3`/`extensions`/`batch` never used it. Nice job not leaving the cleanup half-done — that's usually where these go wrong.

Details I liked: deleting copy/assign, and seeding `previous_` to `rocblas_pointer_mode_host` before the `get` so a failed query still degrades to the sane default rather than to garbage.

Two small things, neither blocking:

1. **Unchecked rocBLAS status.** The constructor ignores the return of both `rocblas_get_pointer_mode` and `rocblas_set_pointer_mode`, while the rest of this backend routes rocBLAS calls through the error-checking helpers. The destructor obviously can't throw, so leave that one alone — but it'd be more consistent to check in the constructor. The old code ignored these too, so this isn't a regression either way.

2. **Behavior change worth naming explicitly.** The old code unconditionally forced the mode back to `host`; the guard restores whatever was there on entry. Since every call site now goes through the guard, the invariant holds and I think your version is the more principled one. But it does remove an accidental self-healing property: if some other path ever leaked device mode, the old code would have normalized it back and the guard won't. I'm fine with this — just flagging it so the change is deliberate rather than incidental.

One thing I noticed that is **out of scope for this PR** — no action wanted here: the USM overloads of `dot`, `sdsdot`, `rotg`, `rotm`, and `rotmg` never set device pointer mode at all, while the USM `asum`/`iamax`/`iamin`/`nrm2` do. That asymmetry predates your change. If it's a real gap it deserves its own issue; please don't grow this PR to cover it.

ndingle-arm (approved):
Approved: sound, focused fix with strong rationale and validation. The CI testing gap is worth noting but non-blocking.

- The RAII guard correctly restores pointer mode during exceptions, covers all 13 former manual reset sites, and checks constructor operations. This matches rocBLAS pointer-mode semantics. AMD documentation (https://rocm.docs.amd.com/projects/rocBLAS/en/develop/reference/helper-functions.html)
- The primary commit message is excellent. 
- No automated rocBLAS regression test was added but the reported MI210 results are persuasive.

PR #768, opened 2026-08-16
[lapack][cuSOLVER] Implement native LAPACK gaps

zjin-lcf:
## Summary
- implement `trtrs` (buffer and USM) with `cublas<t>trsm`, and return 0 from `trtrs_scratchpad_size` since `trsm` needs no scratchpad. `trsm` does not detect the exactly singular matrix that LAPACK reports through `info`, so the diagonal is searched for a zero before the solve is enqueued, which keeps the LAPACK behaviour of leaving `b` untouched and reporting the one based index of the first zero diagonal element (as a `computation_error`)
- implement the grouped (`group_count`/`group_sizes`) USM `getri_batch` with `cublas<t>getriBatched`, one native call per group, and grow the grouped `getri_batch_scratchpad_size` query so the out-of-place inverses of the whole batch fit
- support `nrhs > 1` in `potrs_batch` (buffer strided, USM strided and USM grouped) by solving the columns of B one at a time with `cusolverDn<t>potrsBatched`, which only handles a single right-hand side
- report the info of every matrix of a batch through a new `lapack_info_check_batch` helper that throws `lapack::batch_error`, fix the USM `get_cusolver_devinfo` overload that copied a single `int` regardless of the batch size and did not wait on the asynchronous copy, and give the batched `potrs` solves a real (zero-initialised) info array instead of `nullptr`
- release the temporary device allocations of the strided and grouped `potrs_batch`, including on failure paths, instead of leaking them
- leave `gerqf`/`ormrq`/`unmrq`, `hetrf` and the `m < n` cases of `gebrd`/`gesvd` unimplemented because cuSOLVER and cuBLAS have no native equivalents

### Implementation notes
- `cublas<t>getriBatched` requires a uniform `n` and `lda` within a call, so one call is issued per group. The oneMath 64-bit pivots are converted into a single 32-bit array holding `n` contiguous pivots per matrix, as expected by cuBLAS, and only the `n x n` part of each result is copied back from the scratchpad so that the `lda` padding of `a` is preserved.
- For `potrs_batch`, all right-hand side column pointer arrays are built and uploaded to the device *before* the first native call. When `SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND` is available the native calls are not synchronised, so rewriting a device pointer array between two calls would race with a call still reading it.
- `cusolverDn<t>potrsBatched` only writes its info array when a parameter is invalid, so the array is zero-initialised before the call.
- A unit diagonal is not referenced by `trsm`, so `trtrs` does not search it for zeros and stays fully asynchronous in that case.

## Test plan
- [x] Build the cuSOLVER LAPACK backend with DPC++ and CUDA 13.3.1 (cuSOLVER 12.2.6) on NVIDIA A100
- [x] Run runtime dispatch LAPACK tests on NVIDIA GPU: 398 tests, 352 passed, 46 skipped, 0 failed (base commit: 320 passed, 78 skipped)
- [x] Run compile-time dispatch LAPACK tests on NVIDIA GPU: 398 tests, 352 passed, 46 skipped, 0 failed (base commit: 320 passed, 78 skipped)
- [x] Verify the 32 tests per dispatch mode that stopped being skipped: `Trtrs` accuracy and dependency, `GetriBatchGroup` accuracy and dependency, `PotrsBatchGroup` accuracy (`nrhs = {9, 6}`), `PotrsBatchStride` accuracy for both the buffer and the USM API
- [x] Check the `trtrs` singularity path, which the test suite does not cover, with a standalone program: a zero third diagonal element gives `computation_error` with `info() == 3` and leaves `b` unchanged, for both APIs and for real and complex types, while the same matrix with a unit diagonal solves normally
- [x] Run clang-format 19.1.7 and `git diff --check`

## Portability
The implementation only uses documented cuSOLVER and cuBLAS entry points together with basic SYCL kernels, and keeps the existing `SYCL_EXT_ONEAPI_ENQUEUE_NATIVE_COMMAND` code paths intact.

## Left unimplemented
| Routine | Reason |
| --- | --- |
| `gerqf`, `ormrq`, `unmrq` | neither cuSOLVER nor cuBLAS provides an RQ factorization; left unimplemented for rocSOLVER as well |
| `hetrf` | oneMath's `hetrf` is Hermitian, while cuSOLVER only offers the complex *symmetric* `cusolverDn<C,Z>sytrf` |
| `gebrd`, `gesvd` with `m < n` | pre-existing documented cuSOLVER limitations |


melonakos (commented):
I've reviewed the `gesvd` m < n path here in detail and it's good work. But I want to be straight with you about the rest, and about what I think needs to happen for this PR to actually land.

## Please split this into four PRs

This bundles four independent features, each removing a different `unimplemented` throw:

1. `getri_batch`
2. `trtrs` and `trtrs_scratchpad_size`
3. `gesvd` for m < n
4. `potrs_batch` for `nrhs > 1`

They share no code beyond the file they live in. Each one is a self-contained numerical-library feature that deserves its own review, and each carries its own risk of being subtly wrong in a way that only shows up on real hardware.

I reviewed the `gesvd` work properly — see below. **I have not reviewed `trtrs`, `getri_batch`, or `potrs_batch`, and I'm not going to pretend otherwise by approving the whole thing.** Reviewing four independent LAPACK implementations in one pass is how mistakes get waved through, and the next reviewer will face the same problem: this needs two approvals, and asking someone to sign off on 935 lines covering four features is asking them either to spend a full day or to rubber-stamp it.

Split it and each piece is a 200-line PR that a reviewer can actually hold in their head. I'd expect the `gesvd` one to be approvable more or less as-is on my read, which means you'd get a quarter of this merged quickly instead of all of it sitting.

## What I verified: the gesvd m < n path

**Using `gesvdj` instead of the transpose trick is the better choice.** The obvious way to handle m < n is to compute the SVD of Aᴴ and swap U and V — and that route has a nasty trap, because for complex matrices the relation is Aᴴ = V S Uᴴ, so getting the conjugation wrong gives you plausible-looking but wrong vectors. Switching to the Jacobi solver, which natively supports wide matrices, sidesteps that entirely.

**The `vt` copy-out transposes *and* conjugates correctly.** This was the thing I most expected to be wrong, since `gesvdj` returns V while oneMath's contract is vt = Vᴴ:

```cpp
const std::int64_t row = i % vt_rows;
const std::int64_t column = i / vt_rows;
const auto value = scratch_acc[work_size + u_size + column + row * n];
vt_acc[row + column * ldvt] = conjugate_if_needed(value);
```

Reading `v_[column + row * n]` picks out V[column][row] from a column-major V with leading dimension n — that's the transpose — and `conjugate_if_needed` completes it to Vᴴ. Correct for both real and complex.

**The scratchpad query matches the use site.** The query computes `native_scratch_size + gesvdj_temporary_size(jobu, jobvt, m, n)`, and the implementation recovers `work_size = scratchpad_size - temporary_size` with an `invalid_argument` if the caller under-allocated. Those agree, which matters — if the query had returned only the native size, every user following the documented sizing call would have hit that exception.

**The `gesvdjInfo_t` lifetime is handled properly**, and this is the part I want to single out. The params object has to outlive the asynchronous native work, and you got both halves right: destroyed via `CUSOLVER_ERROR_FUNC` on the success path, and in the `catch (...)` before rethrowing — with the `queue.wait()` in the handler wrapped in its own try/catch so a second exception can't escape during unwinding.

That's careful, and it makes the `devInfo` leaks I flagged on #748, #759 and #762 read as oversights rather than a blind spot — the pattern here is exactly what those need. Worth reusing your own approach there.

## One design question on gesvd

**The algorithm now silently depends on matrix shape.** A 101×100 matrix goes through QR-based `gesvd`; a 100×101 matrix goes through Jacobi `gesvdj`. Those have genuinely different numerical characteristics — Jacobi is often more accurate on graded matrices, but it's iterative, and its `devInfo` means "number of unconverged singular values" rather than the QR routine's "number of unconverged superdiagonals". Both get funnelled into the same `computation_error`, which is defensible, but the failure mode a user sees now varies with the shape of their input.

Two things I'd ask for:

1. **Document it** — in the code at minimum, and ideally wherever backend limitations are listed for users. Someone benchmarking or chasing an accuracy difference between a tall and a wide matrix will otherwise have no way to know the solver changed underneath them.
2. **Set the tolerance and max-sweeps explicitly** rather than relying on `cusolverDnCreateGesvdjInfo` defaults. Right now the convergence behaviour of oneMath's `gesvd` for wide matrices is whatever NVIDIA happens to default to in a given CUDA release, and it can change under you between versions. Pinning them with `cusolverDnXgesvdjSetTolerance`/`SetMaxSweeps` makes the behaviour yours and reproducible.

## Also worth confirming when you split

For the `potrs_batch` piece: cuSOLVER's batched `potrs` only supports `nrhs = 1`, so I assume you're looping over the columns of B. If so, please note the performance characteristic in a comment the way you did for the `dgmm_batch` loop in #756 — it's the same trade-off and the same courtesy to the next reader.

For `trtrs`: LAPACK's contract is that `info = i` when A(i,i) is exactly zero, and I see a `trtrs_throw_if_singular` helper, so you've clearly thought about it. That's precisely the kind of detail I'd want to look at properly in its own PR rather than in passing here.

zjin-lcf:
Split as requested into focused PRs:

- #773 — shared batched `devInfo` correctness fix
- #774 — `trtrs` / `trtrs_scratchpad_size`
- #775 — grouped `getri_batch`
- #776 — wide `gesvd`
- #777 — multiple-RHS `potrs_batch`

I kept the `devInfo` fix separate because both #775 and #777 depend on it; folding it into either feature would create an unrelated dependency. The `gesvd` PR also documents the shape-dependent QR/Jacobi selection and explicitly pins machine-precision tolerance and 100 sweeps. The `potrs_batch` PR documents the one-native-call-per-RHS performance characteristic.

PR #770, opened 2026-08-24
[blas] Add bfloat16-output GEMM support

zjin-lcf:
## Summary
- add the [specified](https://github.com/uxlfoundation/oneMath/blob/develop/docs/spec/domains/blas/gemm.rst#data-types) `bfloat16` A/B/C GEMM overload with float scalars for buffer and USM APIs
- implement BF16 storage with FP32 compute in cuBLAS and expose the existing rocBLAS implementation for both layouts and dispatch modes
- add deterministic transpose/layout coverage and version the expanded backend dispatch-table ABI

This is independent of #769. Both PRs expand the dispatch table, so whichever merges second will need its shared ABI-version changes rebased.

## Test plan
- [x] H100 (CUDA 13.2): cuBLAS CT and RT bfloat16-output buffer + USM tests, 4/4 each
- [x] MI210 (HIP 7.1): rocBLAS CT and RT bfloat16-output buffer + USM tests, 4/4 each (CT repeated 3x without contention)
- [x] H100 and MI210: broader ordinary `GemmTestSuite` + `GemmUsmTestSuite`, CT and RT
- [x] HeCBench PR 323 `blas-gemmEx-sycl` oneMath path on H100 and MI210: BF16 output PASS at 128x128x128 (out-of-scope int8 calls locally guarded for validation)
- [x] `git diff --check`

Toolchain: DPC++ clang 23.0.0git (`fbf7d1fd2cbf`), CUDA toolkit 12.5 / CUDA runtime 13.2, ROCm/HIP 7.1.


src/blas/backends/armpl/armpl_level3.cxx
```diff
@@ -142,6 +142,12 @@ void gemm(sycl::queue& queue, transpose transa, transpose transb, int64_t m, int
     throw unimplemented("blas", __func__, MAJOR_MINOR_TEXT);
 }
 
+void gemm(sycl::queue& queue, transpose transa, transpose transb, int64_t m, int64_t n, int64_t k,
+          float alpha, sycl::buffer<bfloat16, 1>& a, int64_t lda, sycl::buffer<bfloat16, 1>& b,
+          int64_t ldb, float beta, sycl::buffer<bfloat16, 1>& c, int64_t ldc) {
+    throw unimplemented("blas", __func__, MAJOR_MINOR_TEXT);
+}
+
```
ndingle-arm, line 197: I'm not sure if this is in scope for this change, but as of version 26.07 (which is installed in the CI tests here) ArmPL has bf16->fp32 GEMM ("SBGEMM") and bf16->bf16 GEMM ("BGEMM"). See:

https://support.arm.com/documentation/101004/2607/BLAS-Basic-Linear-Algebra-Subprograms/BLAS-Extensions/sbgemm

and

https://support.arm.com/documentation/101004/2607/BLAS-Basic-Linear-Algebra-Subprograms/BLAS-Extensions/bgemm

Could this change include support for calling these ArmPL routines here?

zjin-lcf: Thanks — both are wired up for ArmPL 26.07 now.

`cblas_sbgemm` backs both bfloat16 GEMM overloads:

* bf16 A/B -> **fp32** C (the pre-existing overload just above this one, previously
  `unimplemented` here): a direct 1:1 call, no staging.
* bf16 A/B -> **bf16** C (the overload this PR adds): C is staged in an fp32 buffer around
  `cblas_sbgemm`, so the fp32 accumulation is rounded to bfloat16 exactly once. That matches the
  reference the new test compares against, and the same pattern the `hgemm` path in this file
  already uses.

I did not call `cblas_bgemm`, for two reasons.

**1. Scalar types.** `cblas_bgemm` takes `alpha` and `beta` as `armpl_bf16_t` by value, but this
oneMath overload takes them as `float` (per the spec's [data types
table](https://github.com/uxlfoundation/oneMath/blob/develop/docs/spec/domains/blas/gemm.rst#data-types),
and matching the cuBLAS/rocBLAS paths in this PR), so calling `bgemm` would silently round them.
This is measurable: on `-cpu max`, with `alpha`/`beta` exactly representable in bf16, `bgemm` and
`sbgemm`+rounding are bit-identical on every element I compared. Change only the scalars to
`alpha=1.1, beta=-0.3` and 1951-2047 of 6460 elements diverge, per transpose combination.

**2. Accumulation is not fp32 on every core.** Using C(1x1) = A*B with A = [1, 1/256, 1/256, ...]
and B all ones -- 1/256 is exactly half the bf16 spacing at 1.0, so a bfloat16 accumulator stays
pinned at 1.0 while fp32 accumulation reaches 1 + (k-1)/256:

| k | fp32 result | `sbgemm` | `bgemm` on `-cpu max` | `bgemm` on `-cpu neoverse-v1` |
|---|---|---|---|---|
| 65 | 1.25 | 1.25 | 1.25 | 1.25 |
| 129 | 1.5 | 1.5 | 1.5 | 1.0 |
| 257 | 2.0 | 2.0 | 2.0 | 1.0 |
| 513 | 3.0 | 3.0 | 3.0 | 1.0 |

So on `neoverse-v1` the small terms are dropped entirely beyond k=65. `sbgemm` gives the fp32
answer on every core I tried.

To be clear, this is not a case of `bgemm` failing the new test: with the uniform random data the
test generates, `bgemm` passes the 2-ULP bfloat16 tolerance with 0 failures on all four transpose
combinations and both layouts. The objection is to the contract, not the test result.

Validation of what I pushed: I transcribed both host_task bodies (including oneMath's `copy_mat`
calls, with the same argument order) and ran them against the unit test's reference and tolerance
-- 0 failures out of 661,264 compared elements, across both layouts, all four transpose
combinations, `beta=0`, scalars not representable in bf16, and a larger 64x64x512 case, on
`-cpu max`, `neoverse-v1` and `neoverse-n2`. Elements outside the m-by-n window are left
untouched.

Caveat, same as on #769: I have no Arm hardware, so this was measured by running the real ArmPL
26.07 aarch64 `libarmpl_ilp64.so` under qemu-aarch64 user-mode emulation. Worth confirming on real
silicon, and if `bgemm` is meant to accumulate in fp32 on all cores, the `neoverse-v1` column above
may be an ArmPL bug worth filing.

One note on the guard: the calls are behind `#ifdef _ARMPL_BF16_INTERFACE`, which `armpl_float.h`
introduces in 26.07 (26.01.1 has no bf16 BLAS at all, and there is no version macro in `armpl.h`).
I used `#ifdef` rather than `#if _ARMPL_BF16_INTERFACE == 1` deliberately, since `cblas_sbgemm` is
declared even in the `armpl_bf16_t == void` configuration -- it only takes bf16 by pointer -- so
this keeps working there. `cblas_bgemm` is not, being behind `#if _ARMPL_BF16_INTERFACE == 1` in
`cblas.h`. Happy to change the guard if you'd prefer a different signal.

melonakos (commented):
Reviewed alongside #769, since the two are structurally the same change to different routines. The ABI handling here is done correctly, same as there — but the interaction between these two PRs is the thing that matters most, and it can fail silently.

## Critical: this and #769 both claim `SPEC_VERSION 2`

Both PRs insert new function pointers into the middle of `function_table_t`, and both bump `SPEC_VERSION` from 1 to 2 plus the same 28 `WRAPPER_VERSION` files.

Mid-struct insertion is safe on its own, because the loader rejects a mismatched backend:

```cpp
if (t->version != SPEC_VERSION)
    throw math::specification_mismatch();
```

But if both land at version 2, there are two *different* table layouts both asserting the same version. The check then passes while the offsets disagree, and a backend built from one tree loaded against the other calls the wrong function pointer — the exact corruption the guard exists to catch, now with no diagnostic whatsoever.

The textual merge conflict will show up in `function_table.hpp` and on the version line, so it should be visible — but it would be very easy to resolve that conflict and leave both PRs at 2, which is the failure case.

**Whichever of these merges second needs `SPEC_VERSION 3`** and another pass over all 28 wrapper files. My suggestion: land #769 first (the `dot` surface is simpler), then rebase this one on top and take it to 3. I've flagged the same thing there. Please decide the order deliberately rather than letting merge timing decide it.

## What's right here

The mechanics match #769 and I checked them the same way: `SPEC_VERSION` bumped, all 28 real backend wrapper files bumped, and the only two files mentioning `WRAPPER_VERSION` that you left alone are the backend-authoring templates carrying the `<Wrapper version number>` placeholder. Nothing missed.

As with #769, most of the 56-file diff is one-line wrapper regeneration rather than 56 files of logic — the actual review surface is `function_table.hpp`, `blas.hxx`, `blas_loader.cpp`, and the per-backend `level3` implementations. Worth saying for whoever reviews this second: don't be put off by the file count.

## The bfloat16 output question

For `dot` in #769 the decisive detail was that `cublasDotEx` runs with `CUDA_R_32F` as the execution type, so accumulation happens in fp32 and only the final store narrows. The same question applies here and matters more, because GEMM accumulates over `k` rather than over a single reduction: **the accumulate type must stay fp32 with only the C output written as bfloat16.** Please confirm that's what the cuBLAS and generic paths do, and add a short comment stating it — for a bfloat16-*output* GEMM that's the entire numerical contract, and it shouldn't require a reader to decode a positional argument to find it.

Also worth confirming consistency across the CPU backends you touched (`armpl_level3.cxx`, `netlib_level3.cxx`, `openblas_level3.cxx`, `mkl_level3.cxx`, `generic_level3_bfloat16.cpp`). If any of them accumulates in bfloat16 rather than float, the same call will give materially different answers depending on which backend is loaded, and that's the kind of discrepancy that takes a long time to track down from a user's bug report.

## Same maintainer question as #769

This extends the public API in `include/oneapi/math/blas.hxx` with bfloat16-output GEMM overloads. Whether that's oneMath-spec'd or a vendor extension is a governance decision, not a review one.

@sknepper — same flag as on #769. If these overloads aren't in the spec, deciding now whether they live on the main BLAS surface or behind an extensions header is far cheaper than revisiting it after release.

## Test coverage

You've extended `gemm.cpp` and `gemm_usm.cpp` plus `reference_blas_templates.hpp` (+30) and `test_common.hpp` (+14), which is the right set of places. Given the tolerance work you did in #761 for int8→float, I'd expect bfloat16 output to need similar thought — bfloat16 has 8 mantissa bits, so the output rounding dominates and a plain relative check against a float reference will be loose or flaky depending on the values. Did you reuse the `abs_bound` mechanism from #761 here, or derive a separate tolerance? If the latter, it's worth explaining why in a comment, since the two PRs are solving closely related problems and a future reader will wonder why they differ.

src/blas/backends/openblas/openblas_level3.cxx
```diff
@@ -123,6 +123,17 @@ void gemm(sycl::queue& queue, transpose transa, transpose transb, int64_t m, int
...
+void gemm(sycl::queue& queue, transpose transa, transpose transb, int64_t m, int64_t n, int64_t k,
+          float alpha, sycl::buffer<bfloat16, 1>& a, int64_t lda, sycl::buffer<bfloat16, 1>& b,
+          int64_t ldb, float beta, sycl::buffer<bfloat16, 1>& c, int64_t ldc) {
+#ifdef COLUMN_MAJOR
+    throw unimplemented("blas", "gemm", "for column_major layout");
+#endif
+#ifdef ROW_MAJOR
+    throw unimplemented("blas", "gemm", "for row_major layout");
+#endif
+}
```
ndingle-arm, line 135: Sorry, this occurred to me after I left my previous comments: OpenBLAS also has BGEMM and SBGEMM, so you could apply the same change here as you did for ArmPL and support both this and lines 679-690 in OpenBLAS as well. I don't know if it has a similar ifdef to ArmPL to allow you to check if BF16 is enabled in a certain build though.

src/blas/backends/armpl/armpl_level3.cxx
ndingle-arm: That's great, thank you. That's a good point on scalar types and internal accumulation.

zjin-lcf:
**Merge order with #769:** land #769 first.

Both PRs insert into the middle of `function_table_t` and currently claim `SPEC_VERSION` 2. #769 is the simpler `dot` surface, so it should take ABI v2. This PR stays at 2 until that merge, then rebases onto `develop` and bumps to `SPEC_VERSION` / `WRAPPER_VERSION` 3 (all 28 wrapper files) so the two table layouts never share a version.

Please do not resolve a merge conflict by keeping both insertions at version 2.

